PR #14, opened 2022-03-22
chore: add dockerfile and a makefile to run local server

alcipir:
To make it easier to develop in a consistent way across different machines/OS, this patch adds a dockerfile and a makefile to build a local image, caching installed dependencies and running the container.


Gattaca1 (approved):
Dockerfile and Makefile work as expected; dev env changes look good.

PR #27, opened 2022-03-25
hotfix: get environment variable from netlify directly

alcipir:


alcipir:
New version of this PR deployed on: https://fix-internal-server-error.paybutton.io

PR #35, opened 2022-03-27
chore: adding Error Boundary component

alcipir:
Adds ErrorBoundary component so we can better understand "500 Internal Server Errors" and the rest.
Also have more control of where to log those errors, so in the future we can keep track with a service.


alcipir:
New version of this PR deployed on: https://chore-error-boundary.paybutton.io

Klakurka:
Any errors we can display to showcase this?

alcipir:
Wasn't to able to automatically throw an error due to nextjs/netlify build process. Leaving as-is for now.

PR #40, opened 2022-03-29
ci: fix deployment check to replace all occurences of slashes

alcipir:
Description: adds a second round of find and replace since the action replaces only occurences

Test Plan: check for PR comments on branches with more than one slash, see if URL is correct


alcipir:
merging as it is a small change

PR #46, opened 2022-04-02
ci: remove commit check

alcipir:


alcipir:
will switch to a PR check instead, as is better for github flow

PR #50, opened 2022-04-09
ci: change to pull request depends on check

alcipir:


alcipir:
Merging this, it just replaces the "Depends on" check to something more easier to use with Github, checking Pull Request body messages instead of commits

PR #55, opened 2022-04-29
fix: docker-compose, makefile and update README

alcipir:
Description: 
- Removes MySQL image and volumes from docker-compose **for now**: it doesn't spin up containers in some contexts , **a separate PR will be made with new and tested configuration using mariadb image.**
-  Update README with information regarding credentials (otherwise, currently, the app wont start)
-  Update Makefile with proper docker-compose commands

Test Plan: 
- add credentials to .env or .env.local file
- run `make dev`
- app should be visible on http://localhost:3000


Gattaca1:
looks good to me

Gattaca1 (approved):

PR #56, opened 2022-04-29
fix: run app with available social login credentials

alcipir:
Description:
- Fix issue where, without all social login credentials, app would crash
- Now app will add only available social logins, if none, only email/password will show up on sign in/sign up page.

Test Plan: 
`make dev` and then run without social login credentials, add some social login credentials and run `make dev` again. 
app should be available at http://localhost:3000 with only available credentials, or only email/password.

### DIFF
--- a/config/backendConfig.ts
+++ b/config/backendConfig.ts
@@ -45,7 +45,7 @@ export let backendConfig = () : TypeInput => {
     framework: 'express',
     supertokens: {
       apiKey: process.env.SUPERTOKENS_API_KEY,
-      connectionURI: process.env.SUPERTOKENS_CONNECTION_URI,
+      connectionURI: process.env.SUPERTOKENS_CONNECTION_URI || 'https://try.supertokens.com',
     },
     appInfo,
     recipeList: [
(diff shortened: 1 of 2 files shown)


Gattaca1:
doesn't work at all if `SUPERTOKENS_CONNECTION_URI`is left out.
uncaught promise when only `SUPERTOKENS_CONNECTION_URI` is added to the .env file

works as intended after adding `SUPERTOKENS_CONNECTION_URI` & `SUPERTOKENS_API_KEY`

could you add error catchers to let the user know those things need to be added to the .env file?

alcipir:
Nice catch! yes, definitely, will update asap

alcipir:
I put on a fallback for now, as it was before we added a custom connectionURI: try.supertokens.io (it is the recommended one for testing on supertokens.io docs) 

In future PR related to the frontend specifically, we could add a tooltip/warning modal to inform the user that in production one need to generate a supertokens API key or run it locally.

Gattaca1 (approved):
Looks good to me

PR #57, opened 2022-05-01
feat: add mariadb (mysql drop-in replacement) to docker-compose

alcipir:
Description: Adds mariadb (mysql drop-in replacement) to docker-compose file

Test Plan: run `make dev` and then `docker ps` to see if `db` container is up and running

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,9 +1,23 @@
 version: "3.9"
 services:
   paybutton:
+    restart: always
+    depends_on: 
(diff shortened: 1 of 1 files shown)


Gattaca1:
the containers run as expected, however http://localhost:3000/ is an error page, lets get that fixed

alcipir:
Depends on #56

Gattaca1 (approved):
Looks good to me

PR #58, opened 2022-05-01
chore: add codeowners file to add reviewers automatically

alcipir:
Adding @Klakurka and @Gattaca1 (and me) as reviewers automatically on GitHub

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+@alcipir @Klakurka @Gattaca1


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+@alcipir @Klakurka @Gattaca1
```
Gattaca1, line 1: @Klakurka looks like I need write access to the repo

PR #59, opened 2022-05-05
chore: add check logs commands to makefile

alcipir:
Description: adds `make check-logs-dev` and `make check-logs-db` to run commands to follow logs from paybutton-server container and db container.

Test Plan: run `make dev`, then each of the commands described above and see if logs are shown properly.


Gattaca1 (approved):
looks good

PR #60, opened 2022-05-05
chore: remove deployment check action

alcipir:
Description: Remove deployment check action, as now we are not deploying anywhere (disabled netlify).

Test Plan: check if this PR runs a deployment check action.


Gattaca1 (approved):
looks good

PR #61, opened 2022-05-05
feat: add db dependencies, config and migrations

alcipir:
Description: Adds four initial necessary components for the database integration: **dependencies for sequelize** (ORM for nodejs)/mysql, **package.json with commands to run db migration**, **docker-compose command to make the paybutton-server initialize with migrations** and **a .sequelizerc config file to change the sequelize root folder to `db/`**

Test Plan: run `make dev`, then `check-logs-dev` , see if no errors occurred and this log appeared:

image

Depends on #59


db/models/index.js
```diff
@@ -0,0 +1,37 @@
...
+  });
+
+Object.keys(db).forEach(modelName => {
+  if (db[modelName].associate) {
+    db[modelName].associate(db);
+  }
+});
+
+db.sequelize = sequelize;
+db.Sequelize = Sequelize;
```
Gattaca1, line 35: Disambiguate these somehow if they are both needed

db/migrations/20220408204736-create-paybuttons.js
```diff
@@ -0,0 +1,27 @@
...
+  async up(queryInterface, Sequelize) {
+    await queryInterface.createTable('paybuttons', {
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.INTEGER
+      },
+      userId: {
+        type: Sequelize.STRING,
```
Gattaca1, line 12: If this is a userID it should probably be an integer. If its a username though, lets rename it

alcipir: In this case it should be a string as it refers to a user_id on the supertokens server: https://github.com/supertokens/core-driver-interface/wiki

db/models/index.js
alcipir: the **Sequelize** one doesn't make sense to me at first glance, so I'm removing it for now.

db/migrations/20220408204736-create-paybuttons.js
alcipir: we could rename it to supertokens_user_id or user_from_provider_id or something more explicit, do you have any suggestion?

alcipir:
I sketched out a ERD for the db entities being implemented:

image
 
 We probably will add more in the near future, so I'm thinking about a starting point. We can always add migrations to add/remove more tables/entities

db/migrations/20220408204736-create-paybuttons.js
Gattaca1: Yeah lets rename that to either of `SupertokensUserId`, or `ProviderUserId` is fine

Gattaca1:
Could you include some more in the test plan? Other than just no errors, what should I really be testing? Am I just looking that the db tables were created, or should there be some mock data in them? Is `SequelizeMeta` an expected table in the db?

db/migrations/20220408204736-create-paybuttons.js
alcipir: done

alcipir:
The objective of this PR is to add required db dependencies, config and migrations, so the Test Plan only requires those steps. The logs itself will tell if every initial table was properly created. 

Regarding SequelizeMeta, it's to keep track of migrations:
image

(From Sequelize Docs: https://sequelize.org/docs/v6/other-topics/migrations/)

alcipir:
Regarding errors before finally connecting to db: since the paybutton-server container was already set up, it will start before the db container, that waits for image downloading and some startup process. Paybutton will keep restarting until it is able to connect to the database. It will not happen often, however, we could use a tool to wait for other containers to start before starting paybutton: https://github.com/ufoscout/docker-compose-wait

I'm just not sure I want to add another thing into this PR as to keep it small. Let me know what you guys think @Klakurka @Gattaca1

Gattaca1:
ok, looks good to me

PR #62, opened 2022-05-07
Chore/add users service (supertokens local instance)

alcipir:
Description: adds supertokens (users service) local instance
Test Plan: 
- Remove existing SUPERTOKENS_API_KEY and SUPERTOKENS_CONNECTION_URI from your .env.local or .env file
- Run `make dev` **(`make stop-dev` first if containers already running)**  and then: 
  - Open up http://localhost:3567/hello on you browser, there should be a 'Hello' text on the page.
  - Sign up with email and password, and run `make check-logs-users`, there should be something similar to this:
      image

### DIFF
--- a/.env
+++ b/.env
@@ -1,1 +1,1 @@
-GRPC_NODE_URL=bchd.greyh.at:8335
+SUPERTOKENS_CONNECTION_URI=http://users_service:3567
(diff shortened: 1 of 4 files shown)


Gattaca1:
Looks good

PR #64, opened 2022-05-11
Save paybutton & addresses on the  database

chedieck:
Filling form and clicking the **Submit** button now commits the newly created paybutton and its addresses to the database.

saving-pb-to-db-2022-05-11_13 11 46

There is no validation being performed on these addresses (if the string before the colon is a valid chain, if the text field is not empty, etc); I'll do that in the next PR.
I am also mocking the user because I couldn't get their ID right now, so I'll also solve that in a different PR.

### DIFF
--- a/components/AddPayButtonForm.tsx
+++ b/components/AddPayButtonForm.tsx
@@ -13,7 +13,7 @@ import axios from "axios";
 const schema = {
     fields: [
         {
-            component: "text-field",
+            component: "textarea",
             name: "addresses",
             label: "Addresses",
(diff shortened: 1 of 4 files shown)


Klakurka:
One other thing to be thinking about is that buttons can be generated on the fly via JS (like we do on https://coin.dance/support).

Also in the future, xpub keys (or something like that) will also be valid.

Gattaca1 (approved):
Looks good

alcipir:
LGTM!

PR #65, opened 2022-05-12
fix: remove mocked user id from api

alcipir:
Description: Removes 'mocked-user-id' from API, using userId provided by the app

Test Plan: `make dev`, go to http://localhost:3000 and then submit a form, a list with payButtonId and addresses associated with it should appear after submiting.


Gattaca1 (approved):
Unmentioned in the task description is the removal of the 'Add Paybutton' button. Makes sense with the task though, so everything looks good

PR #66, opened 2022-05-12
feature: create chains table, added constraints to database.

chedieck:
This branch introduces some changes in the database that are paramount for validating the input field when creating paybuttons. The branch doesn't include the validation yet, this will come in a next PR. What is includes is:

- Creates a new table `chains` in the database with two entries: one for `bitcoincash` and another one for `ecash`
-  Adds constraints to the database:
  + a paybutton address now MUST have the columns: address, chain & paybuttonId.
  + when deleting a paybutton, all its addresses are also deleted,
  + not possible to delete a chain from the database if there are any addresses of that chain.
- Created service files to make changes / queries to the database.
- Creating a paybutton now does all in one transaction, if any step fails, all is rollbacked.

 Depends on #65


db/models/chains.js
```diff
@@ -3,21 +3,22 @@ const {
...
+    slug: DataTypes.STRING,
+    title: DataTypes.STRING
   }, {
     sequelize,
-    modelName: 'paybuttons_addresses',
+    modelName: 'chains',
   });
-  return paybuttons_addresses;
+  return chains;
 };
```
Gattaca1, line 24: Don't rename files only to re-add the old file. It makes finding the history of changes annoying to find. Just add the new file

chedieck: I didn't rename them, I actually deleted them and created new ones, and in different commits, to try to avoid that. I don't understand why the diff is like that

chedieck: I deleted the migrations file on [this](https://github.com/PayButton/paybutton-server/pull/66/commits/bd1f6ab44f42f492dc39b9b940dcd84cb0dccb23) commit and created the new migrations on [this one](https://github.com/PayButton/paybutton-server/pull/66/commits/70a2c99bb13558e0d4b4853f148612aadb2dbb11).

This is a model file, so what I did doesn't apply to it, nonetheless, this problem of interpreting the deletion + creation as a rename is also happening on the aforementioned migration files, so I think it wouldn't have helped to do the same in the model files.

Gattaca1 (changes requested):
- Change file names back to what they were previously
- New files (such as `create-chains.js` and `paybutton_addresses.js`) should actually be a new file (instead of `create-paybutton-addresses.js` now being a new file)

db/models/chains.js
Gattaca1: Ok, we can ignore this for now. Next time instead of deleting, try just using `git mv oldFileName newFileName`

Gattaca1:
This branch is pretty trusting in how the db accepts data from the api without any checks, but as you said that will be patched in the next branch. This looks good to me

Gattaca1 (approved):

PR #68, opened 2022-05-13
Added validation function for add paybutton form

chedieck:
Input of paybutton addresses is now validated on the front end:
validating-input-2022-05-13_11 48 21

Legacy addresses are not being allowed.


chedieck:
The width changes accordingly to the error message, it would be interest to fixate it the future

alcipir:
IMO we should leave frontend for upcoming PRs since it will be restructured a bit. 
We could add validation on the backend as well though, just in case.

types/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const supportedChains = [
```
Gattaca1, line 1: Supported chains should be requested from the server

Gattaca1:
> IMO we should leave frontend for upcoming PRs since it will be restructured a bit. We could add validation on the backend as well though, just in case.

I'll second this. Lets add back end validation instead + error codes. Also lets do some sanitizing before the data gets put into the database (probably should have been part of yesterdays branch anyways)

types/constants.ts
chedieck: You mean this should be a query that gets all chain slugs from the database? I thought about implementing that way, but that would be an extra query thatt would have to be done every time some address is input. Having a constant that corresponds to the supported chains in the database (this correspondence will be enforced by unittests) looked like a more efficient solution.

chedieck:
I moved the validation logic to the backend.

chedieck:
I've added unittests (and fixed the problems these unittests brought to the surface), now this PR depends on #72

alcipir:
@Gattaca1

constants/index.ts
```diff
@@ -0,0 +1,5 @@
+export const supportedChains = [
+  'ecash',
+  'bitcoincash'
+] // WIP integration tests to check if this corresponds to the database
+export const supportedAddressesPattern = /(q|p)[a-z0-9]{41}/
```
Gattaca1, line 5: Lets get more specific with supported addresses. Heres the address format spec
[https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/src/cashaddr.cpp#L16](https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/src/cashaddr.cpp#L16)

relevant part: `const char *CHARSET = "qpzry9x8gf2tvdw0s3jn54khce6mua7l";`

constants/index.ts
```diff
@@ -0,0 +1,5 @@
+export const supportedChains = [
```
Gattaca1, line 1: Lets also add the testnets to the supported chains `bchtest`, `bchreg`

constants/index.ts
Gattaca1: An equally valid charset is all uppercase. An address cannot mix upper and lower case though

`ecash:QPZRY9X8GF2TVDW0S3JN54KHCE6MUA7L` is valid
`ecash:QPZRY9X8GF2TVDW0S3jn54khce6mua7l` is not

chedieck:
@Gattaca1 All remarks addressed.

utils/validators.test.ts
```diff
@@ -0,0 +1,104 @@
...
+  )).toBe(true)
+});
+
+test('Accept example address one uppercase other lowercase', () => {
+  expect(v.validateAddresses(
+    ['ecash:QPZ274AAJ98XXNNKUS8HZV367ZA28J900C7TV5V8PC',
+     'bitcoincash:qz0dqjf6w6dp0lcs8cc68s720q9dv5zv8cs8fc0lt4',]
+  )).toBe(true)
+});
+test('Reject chain uppercase', () => {
```
Gattaca1, line 33: Sorry man, I forgot to mention this from the spec. We should squash the chain to lowercase when we validate it, example here:
[https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/src/cashaddr.cpp#L264](https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/src/cashaddr.cpp#L264)

so `ECASH` becomes `ecash` making it valid

Gattaca1 (approved):
Looks good

PR #73, opened 2022-05-18
Fetching paybutton from db

chedieck:
Implemented the `GET` endpoints for `api/paybutton/` and `api/paybutton/[id]`, that were being mocked.
Depends on #68 
# Test plan
- Create some paybuttons using the front-end,
- Then test the `api/paybutton/[id]` endpoint by issuing something like:

 ```
curl -X GET localhost:3000/api/paybutton/7
```
-  Finally  test the `api/paybutton/` with something like

```
curl -X GET localhost:3000/api/paybutton -H "Content-Type: application/json"  -d '{"userId" : "u-id"}'
```

You can find the user ID to substitute `"u-id"` for by querying it in the database, or just create a user `u@u.com` with password `u` with the statements
```
INSERT INTO emailpassword_users VALUES ('u-id', 'u@u.com', $2a$11$QOyNFFmciE6Hda1avqxxa.qLMYIJSBWn7eV8i6Ude73qu5l2/wA5e', 1652220489244);
INSERT INTO emailverification_verified_emails VALUES ('u-id', 'u@u.com');
INSERT INTO all_auth_recipe_users VALUES ('u-id', 'emailpassword', 1652220489244);
```
# Remarks:
I stumbled into many errors when trying to mock the database to unittest the end points, so I decided to implement a test DB for integration tests already (that I will commit in a next PR).

### CODE REVIEW

**Comment by Gattaca1** [comment]
This all looks fine but I'll wait till #68 is merged till I do a thorough review

**Comment by alcipir** [comment]
Can we settle for async/await instead of regular promise chaining/callback? Do you guys think is a good idea to set a standard usage in this regard, just to keep the code consistent?

**Comment by chedieck** [comment]
> Can we settle for async/await instead of regular promise chaining/callback? Do you guys think is a good idea to set a standard usage in this regard, just to keep the code consistent?

Addressed, check out last commit. The only other place using `then(...)` is `server/index.ts:14`, which AFAIK is deprecated, so I didn't change it.

File: types/index.ts [hunk 1]
```diff
@@ -1,5 +1,14 @@
+// By default, Sequelize automatically adds the fields createdAt and updatedAt to every model,
+// so they are also present in these types
+
```
alcipir, line 3: I don't understand this comment, since these types are not present here. I'm not sure if they are needed anyway. Can we remove this comment?

constants/index.ts
```diff
@@ -4,4 +4,11 @@ export const SUPPORTED_CHAINS = [
   'bchtest',
   'bchreg'
 ] // WIP integration tests to check if this corresponds to the database
```
alcipir, line 6: I believe we should leave WIP comments for _draft_ pull requests, not ones that are ready for review.

pages/api/paybutton/index.ts
```diff
@@ -1,48 +1,38 @@
...
-}
-
-
-export default (req: NextApiRequest, res: NextApiResponse) => {
+export default async (req: NextApiRequest, res: NextApiResponse) => {
+  const values = req.body
+  const userId = values.userId
+  if (!userId) {
+    res.status(400).json(RESPONSE_MESSAGES.USER_ID_NOT_PROVIDED_400)
+    return
```
alcipir, line 12: Do we need this return statement here?

alcipir (changes requested):
Left some comments there. After those changes, I will merge this PR as it seems almost stale.

pages/api/paybutton/index.ts
chedieck: Yes, or the paybutton will be created without a providerUserId.

types/index.ts
chedieck: I meant that, since we use these types to typehint objects that come from the database, they will also have these fields. I wasn't sure if I should add the fields to every type, because it would be a lot of repeated code. In any case, I removed the comment

Gattaca1:
> I stumbled into many errors when trying to mock the database to unittest the end points, so I decided to implement a test DB for integration tests already (that I will commit in a next PR).

I don't think we should skip unit tests in favour of integration tests - it should be both. If you have any specific errors, maybe we can help out. Here is a link quickly describing mocking a db with ts-jest [https://stackoverflow.com/questions/68825658/mock-database-jest-using-ts-jest-utils](https://stackoverflow.com/questions/68825658/mock-database-jest-using-ts-jest-utils)

There is a broken link in the stack overflow reply there, [heres the correct link](https://kulshekhar.github.io/ts-jest/docs/27.0/guides/test-helpers/)

pages/api/paybutton/index.ts
alcipir: can we move all errors to the exception handling block? 

in this case would be something like `throw new Error(RESPONSE_MESSAGES.USER_ID_NOT_PROVIDED)`

and then on catch block:
`switch (e.message): 
    case RESPONSE_MESSAGES.USER_ID_NOT_PROVIDED:
           res.status(400).json......
           
 
 we could throw an error on validateAddresses as well and add the response on a switch case

chedieck:
> > I stumbled into many errors when trying to mock the database to unittest the end points, so I decided to implement a test DB for integration tests already (that I will commit in a next PR).
> 
> I don't think we should skip unit tests in favour of integration tests - it should be both. If you have any specific errors, maybe we can help out. Here is a link quickly describing mocking a db with ts-jest https://stackoverflow.com/questions/68825658/mock-database-jest-using-ts-jest-utils
> 
> There is a broken link in the stack overflow reply there, [heres the correct link](https://kulshekhar.github.io/ts-jest/docs/27.0/guides/test-helpers/)

You're right, I've fortunately managed to add the unittests (after doing the integration tests & the unittests for the BCHD services, I think I started getting the hang of it), check the last commit.

I haven't addressed @alcipir  last requests though, so the PR as a whole is not finished.

pages/api/paybutton/index.ts
alcipir: I think for now we can just return res.status... instead of just a void return call

chedieck: @alcipir I've change the logic on the endpoint to throw errors on invalid requests, and a `switch` case to  adapt the response accordingly.

alcipir (approved):

Gattaca1 (changes requested):
> # Test plan
> 
>     * Create some paybuttons using the front-end,
> 
>     * Then test the `api/paybutton/[id]` endpoint by issuing something like:
> 
> 
> ```
> curl -X GET localhost:3000/api/paybutton/7
> ```

- The front end no longer has a way to create paybuttons
- `pages/api/paybutton/[id].ts` doesn't compile due to missing curly brace

Could you re-add the button for creating paybuttons to the front end?

Klakurka:
Also I'd say this one and #77 are examples of "feature branches" which we want to try to avoid (by making more frequent, smaller branches) as they're easier to work with: https://ourmachinery.com/post/step-by-step-programming-incrementally/

chedieck:
> > # Test plan
> > ```
> > * Create some paybuttons using the front-end,
> > 
> > * Then test the `api/paybutton/[id]` endpoint by issuing something like:
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > curl -X GET localhost:3000/api/paybutton/7
> > ```
> 
>     * The front end no longer has a way to create paybuttons
> 
>     * `pages/api/paybutton/[id].ts` doesn't compile due to missing curly brace
> 
> 
> Could you re-add the button for creating paybuttons to the front end?

Indeed, that was because I merged master into it to already solve possible conflicts.
You actually don't need the front end though, you can create paybuttons with, for example:
```
curl -X POST localhost:3000/api/paybutton -H "Content-Type: application/json"  -d '{"userId": "u-id", "addresses":"ecash:qpz274aaj98xxnnkus8hzv367za28j900c7tv5v8pc\nbitcoincash:qpz274aaj98xxnnkus8hzv367za28j900
c7tv5v8sc"}'
```
Just change `u-id` for your user id, or run the SQL in the PR description to create an user with such ID

chedieck:
> Also I'd say this one and #77 are examples of "feature branches" which we want to try to avoid (by making more frequent, smaller branches) as they're easier to work with: https://ourmachinery.com/post/step-by-step-programming-incrementally/

Alright, I'll separate big changes into multiple PRs in the future

Gattaca1 (approved):

Gattaca1 (approved):
Looks good

PR #77, opened 2022-05-25
Feat/Add integration tests

chedieck:
Created a test database and some sample integrations tests. Currently, I am only adding tests for `POST` requests on the `api/paybutton/` endpoint, they serve as a sample so the reviewer can test if the integration tests are working. I'll add more integration tests for other endpoints in  future PRs.

Test plan:
- Recreate the `db` container,  so it also includes the test database & test user;
- Inside the node container:
  -  Run yarn;
   - Run `yarn test`, see if all tests are run and they succeed. Of course, running the tests should not affect the `paybutton` table on the database.

Depends on #73


tests/api.test.ts
```diff
@@ -0,0 +1,76 @@
+// import * as paybuttonsService from 'services/paybuttonsService'
```
alcipir, line 1: commented code

alcipir (changes requested):
Just some commented code to remove. 

Otherwise looks good to me.

alcipir (approved):

Gattaca1:
Perhaps I'm running this incorrectly. I ran `yarn`, then `make dev` then `yarn test` which resulted in the following error
```
jeff@MainUbuntu:~/Documents/paybutton-server$ yarn test
yarn run v1.22.18
$ cross-env NODE_ENV=test yarn db:reset
$ sequelize-cli db:drop && sequelize-cli db:create && sequelize-cli db:migrate && sequelize-cli db:seed:all

Sequelize CLI [Node: 16.14.2, CLI: 6.4.1, ORM: 6.19.2]

Loaded configuration file "db/config/config.js".
Using environment "test".

ERROR: getaddrinfo EAI_AGAIN db

```

alcipir:
@chedieck Forgot to ask: can we have different folders for unit and integration tests? It's not quite clear which is which.

chedieck:
> @chedieck Forgot to ask: can we have different folders for unit and integration tests? It's not quite clear which is which.

Addressed. PR should be working now, ready to be reviewed again.

chedieck:
> Perhaps I'm running this incorrectly. I ran `yarn`, then `make dev` then `yarn test` which resulted in the following error


Did you run `make stop-dev` before?

alcipir:
Is the test running inside the container? 

On this draft, I changed the "test" npm script to run inside the container (possible now that we have fixed container names).

https://github.com/PayButton/paybutton-server/pull/84/files

chedieck:
> Is the test running inside the container?
> 
> On this draft, I changed the "test" npm script to run inside the container (possible now that we have fixed container names).
> 
> https://github.com/PayButton/paybutton-server/pull/84/files

I have only run the tests inside the container, by running `yarn test` inside it (not sure if that is what you're asking).

alcipir:
The Test Plan should be changed then, as it says to just run `yarn test`, without details about running it only inside the container. 

Let's make sure we keep the test plan always updated,  saving time for our reviewers. 
(I'm guilty of this as well, just a general observation)

In this case, it would be helpful to implement the npm script I talked about earlier: it's an alias to run yarn test inside the container to make it easier than finding the proper docker command, container name and such.

chedieck:
> The Test Plan should be changed then, as it says to just run `yarn test`, without details about running it only inside the container.
Indeed, I have now edited it, and will make it more clear on the future


> In this case, it would be helpful to implement the npm script I talked about earlier: it's an alias to run `yarn test` inside the container to make it easier than finding the proper docker command, container name and such.

About this, I use an [alias script](https://github.com/chedieck/bin/blob/master/dca) to perform all the operations I usually want to do inside the docker container, do you think I should move it to this repository or port it to the Makefile?

alcipir:
I think we should make it simple: https://github.com/PayButton/paybutton-server/pull/84/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R26 would be enough if you agree. 


Although this alias is very handy indeed.

alcipir:
I think we should make it simple: https://github.com/PayButton/paybutton-server/pull/84/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R26 would be enough if you agree. 


Although this alias is very handy indeed.

services/paybuttonsService.ts
```diff
@@ -3,7 +3,7 @@ import { PayButton } from 'types'
 import * as chainService from 'services/chainsService'
 
 export async function createPaybutton (userId: string, prefixedAddressList: string[]): Promise<PayButton>  {
-    const result = await models.sequelize.transaction(async (t) => {
+    const result = await models.sequelize.transaction(async (t: any) => {
```
Gattaca1, line 6: Lets try to keep linting of existing files to their own PR's, as linting only PR's need very little review.

alcipir: done, please git pull @chedieck

alcipir (approved):

Gattaca1 (approved):
looks good!

PR #80, opened 2022-05-28
refactor: update appInfo.ts following ts-standard

alcipir:
Description: update appInfo.ts, making it follow our new ts-standard format.

Test Plan:  `npx ts-standard config/appInfo.ts` should not fail.


chedieck (commented):
Tested, works well

chedieck (approved):

PR #81, opened 2022-05-28
ci: add named containers and update Makefile

alcipir:
Description: Add named containers, updating Makefile accordingly. This fixes an issue where some operating systems would create a container with a different name (e.g. paybutton_server instead of paybutton-server) and would fail to run `make check-logs[...]` due to this mismatch.

Test Plan: `make dev`, then `make check-logs-dev`, `make check-logs-db`, `make check-logs-users`, see if logs are being printed accordingly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,11 +8,10 @@ stop-dev:
 	docker-compose down
 
 check-logs-dev:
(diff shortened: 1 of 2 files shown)


chedieck (approved):
Works well.

PR #82, opened 2022-05-28
refactor: fix ts-standard on config files at config

alcipir:
Description: Refactor backend and frontend config files inside config/, following **ts-standard** conventions.

Test Plan: `make dev`, check if everything is running. `npx ts-standard config/backendConfig`, `npx ts-standard config/frontendConfig` should both pass.


chedieck (approved):
Looks good

PR #83, opened 2022-05-30
docs: update README and fix formatting

alcipir:
Description: Fix README.md formatting, remove netlify badge and remove detailed instructions on social providers as (1) it's redundant since the provided link has a more detailed explanation already and (2) social login might not work properly on localhost depending on the social provider, so it could be misleading to newcomers.

Test Plan: [Check README preview here](https://github.com/PayButton/paybutton-server/tree/docs/update-readme#readme)


Gattaca1 (approved):
Looks good

PR #84, opened 2022-05-31
ci: add testing to PR workflow

alcipir:
Description: adds testing to Github Actions PR workflow

Test Plan: under `checks` tab, should show the workflow running lint, format and **unit tests successfully**

Depends on #77


chedieck (dismissed):

Makefile
```diff
@@ -21,3 +21,6 @@ lint-master:
 	$(git_diff_to_master)
 	npx --yes ts-standard --stdin --stdin-filename DIFF
 
+test:
+	make dev
+	docker exec -it paybutton-dev sh -c "CI=true yarn test"
```
chedieck, line 26: I think this command does not work:
image

chedieck (changes requested):

Makefile
alcipir: Removed the `it` argument as it was running interactively, checking if action now goes through.

alcipir:
A docker container inside github actions was having issues of memory due to sequelize. I think it is a good practice to keep unit tests and integration tests separated, so I was wondering if we could make this workflow for now only for lint + format + unit tests, so no starting docker here. 

Integration tests will change/behave differently due to prisma (and probably better than sequelize). 

I had to make tests and sequelize models relative to their path, tested here in a container of mine and it didn't change anything. Can you check in your machine @chedieck  @Gattaca1? 

Should I put these changes in a separate PR?

chedieck (approved):

chedieck:
> A docker container inside github actions was having issues of memory due to sequelize. I think it is a good practice to keep unit tests and integration tests separated, so I was wondering if we could make this workflow for now only for lint + format + unit tests, so no starting docker here.
> 
> Integration tests will change/behave differently due to prisma (and probably better than sequelize).
> 
> I had to make tests and sequelize models relative to their path, tested here in a container of mine and it didn't change anything. Can you check in your machine @chedieck @Gattaca1?
> 
> Should I put these changes in a separate PR?

I tested on my machine and looks like everything is still working. I agree that when we finish merging the sequelize -> prisma PRs we can  try to run the integration tests again.

Gattaca1 (approved):
Looks good

PR #85, opened 2022-05-31
fix: users-service container name in make check-logs-users

alcipir:
Description: fix missing proper name for users-service (supertokens) container on `make check-logs-users`.

Test Plan: `make dev`  && `make check-logs-users`, should show logs for the users-service container.


Gattaca1 (approved):
Quick fix, looks good

Issue #87, opened 2022-05-31
Move from Sequelize to Prisma


chedieck:
[Sequelize](sequelize.org/), mainly because it lacks a decent documentation and TypeScript compatibility, has already proved to be a huge source of bugs, and has already demanded a lot of work time dedicated to trying to find workarounds around things that should have been simple, some examples:
- TypeScript complains accessing any of the model objects, since it cannot infer this from typing. This means that we either cannot check for typing in tests, or we have to pollute a whole lot of our code with workarounds such as `// @ts-ignore`
- Updating the database is probably easier done with raw queries than with manually creating migration files, since these are not auto-generated from the model files, are not a single source of truth (SSOT), and the documentation for it... let's say that the only way to find how to properly use `Sequelize.queryInterface` methods are either lurking on obscure StackExchange threads or by hacking Sequelize source code itself. Many times these threads will provide raw queries solutions, unsurprisingly.

[Prisma](https://www.prisma.io/), unlike Sequelize, is well documented, built for typescript, supports auto-generated migrations and works through a SSOT scheme. It's also almost as popular as Sequelize, but Sequelize is still the most used ORM.

Gattaca1:
Makes sense to have an ORM that plays better with TS. I'm in favour of this switch while the project is still young.

chedieck:
I've separated this task into smaller ones, each of which I shall address in a different PR.

- [ ] Add & configure Prisma
- [ ] Remove `types/index.ts` and use Prisma generated types in its places
- [ ] Change `cross-env` for `dotenv`
- [ ] Change all sequelize queries for prisma queries;
- [ ] Change the database instantiation to use prisma instead of sequelize
- [ ] Change the prisma scheme so that table names / field names adhere with prisma conventions
- [ ] Remove sequelize from the project

PR #91, opened 2022-06-01
Add a configured Prisma to the repository

chedieck:
**Description:**
Adds Prisma plus its configuration to the project.
As [recommended from Prisma docs](https://www.prisma.io/docs/guides/development-environment/environment-variables/using-multiple-env-files), I've addedd the environment files `.env.development` and `.env.test`. In a future PR, I'll refactor `cross-env` calls to use `dotenv` instead, to reduce unnecessary complexity.

**Test plan**:
This PR does not changes the current functionality, which depends on sequelize, so there is nothing in particular to be tested. For the full Prisma implementation plan, refer to #87


alcipir:
`cross-env` is actually a leftover from the custom express server. Have you tried running it without it?

NextJS should be able to handle it, it has a builtin `next/env` library that you can use inside jest/other scripts as well:
https://nextjs.org/docs/basic-features/environment-variables

prisma/schema.prisma
```diff
@@ -0,0 +1,188 @@
...
+  @@index([role], map: "user_roles_role_index")
+}
+
+// End region: supertokens models
+
+// To be removed with sequelize
+model SequelizeMeta {
+  name String @id @unique(map: "name") @db.VarChar(255)
+}
+
```
alcipir, line 188: Do we really need all these models for supertokens? We are currently pointing a local instance to our database, but that's not the most common way of doing it. 

Some people might run it in separate databases (even our production stack), and most people would just use the free tier from https://supertokens.com/auth. 

We might be coupling it to a third-party service without benefits.

alcipir:
Aside from my comment about the supertokens schema, thanks for keeping **diffs small**. 

I will just ask to follow the "Description: <description> *new line* Test Plan: <test plan, if no test plan, then explanation why>. I will add a github action to check for that in the future (the old one stopped working properly).

prisma/schema.prisma
chedieck: Indeed, there is no need for modeling all of these super token tables.

I think that in the future we may want to access users' models on the backend, for example, if we decide to have something like granulate permissions for the server users. Since this is not currently on scope, I just removed all of it, together with the `SequelizeMeta` model.

Gattaca1 (dismissed):
looks good

alcipir (approved):

alcipir (approved):

Gattaca1 (approved):
Re-approving

PR #94, opened 2022-06-03
Added script to perform docker exec operations

chedieck:
**Description**:
Add a `docker-exec-shortcuts.sh` script with shortcuts for inside-container commands to be run. Also, a yarn shortcut to run it:
To enter the `paybutton` database, for example, inside the docker `paybutton-db` container, simply run `yarn des db`.

There were also commands related to migrations & prisma,  but I took them off and will add then in the future PRs that have to do with migrating from sequelize to prisma, when I am sure that they are working.

**Test Plan:**
To see a fulll list of available commands, run `yarn des`. Then try the commands by running each one of them, such as `yarn des ns`


alcipir:
`assets` is usually reserved for static files, can we move it to `scripts` folder?

alcipir:
Can we call it `yarn docker` instead of `yarn des`? Might help reviewers understand it as a docker alias.

chedieck:
@alcipir both addressed.

alcipir (approved):

Gattaca1 (approved):

PR #100, opened 2022-06-06
Sequelize to prisma: Instantiate database with prisma

chedieck:
**Description:**
The database is now initialized with prisma instead of sequelize. This is the case in:
- `docker-compose.yml`,
- The test database,
- The `package.json` script commands.

Also, since I've already had to recreate the migrations for prisma,  I used table names that they are in accord with [prisma naming conventions](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#naming-conventions) (PascalCase & singular).

**Test plan:**
- Run `make reset-dev`
- Run the tests with `yarn docker t`
- Run `yarn docker mr` 

Depends on:
- [x]  #77 
- [x]  #99 
- [x]  #98 
- [x]  #101

PR #102, opened 2022-06-07
Sequelize to prisma: Delete sequelize

chedieck:
**Description:**
Removes the modules `sequelize`, `sequelize-cli` and all its unused dependencies. Also removes all other sequelize-related files.

**Test plan:**
None, it should work the same as the master branch.

Depends on #100


Gattaca1 (approved):
good riddance to sequelize; looks good

PR #103, opened 2022-06-07
Feature: API integration tests improvement

chedieck:
**Description:**
- Added tests to the GET endpoints
- Fixed bug (found by the tests) on the [id] endpoint (the switch case wouldn't break after setting the error as 404 and thus set it to 500, or raise another error)

**Test plan:**
`make reset-dev`
`yarn docker t`

- [ ] Depends on #102


Gattaca1 (approved):
Looks good

Issue #67, opened 2022-05-13
Extremely slow call to address/transactions/xxx


Klakurka:
Example: https://api.paybutton.org/address/transactions/bitcoincash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7q9ayphulp

Klakurka:
https://api.paybutton.org/address/balance/bitcoincash:qzqh7ej2vz26a9xaxq7capzfwgxt5gem9g8rvfxc5t

Klakurka:
All good with the new server as we expected.

PR #104, opened 2022-06-09
feat: add addresses transactions api routes

alcipir:
Description: adds address/<cashaddress/transactions endpoint (from legacy custom server), related to #67.

Test Plan: add a working bchd node in `.env` file, like `GRPC_NODE_URL=bchd.fountainhead.cash:443`. 
**Make sure you remove lal your node_modules to not conflict with container ones as it could result in grpc error**.
`rm -rf node_modules`
`make dev`

Go to http://localhost:3000/api/address/transactions/bitcoincash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7q9ayphulp, transactions should be there.


chedieck (commented):
The [address/transactions/[address].ts](https://github.com/PayButton/paybutton-server/pull/104/files#diff-4fcb5506aae8029587cbaed551815651694fb582b3295164b551963cf82af3b1) endpoint is only for BCH for now, right? Wouldn't it be good if we included this aspect on the endpoint path or in the request?

chedieck:
> We don't need the currency in the url if we can do detection based on the address format.
> 
> The button will require that the correct format is used.

I see, you mean based on the prefix, right? In that case, I suggest changing the name of the file from `[address].ts` to `[prefixedAddress].ts`, just to make clear what is being expected here. Since we plan on supporting more than one network, I think that clearly stating when an address is expected to come with a prefix is important.

Klakurka:
> > We don't need the currency in the url if we can do detection based on the address format.
> > The button will require that the correct format is used.
> 
> I see, you mean based on the prefix, right? In that case, I suggest changing the name of the file from `[address].ts` to `[prefixedAddress].ts`, just to make clear what is being expected here. Since we plan on supporting more than one network, I think that clearly stating when an address is expected to come with a prefix is important.

Sure it's the prefix but technically its the entire address. For example bch and xec have different checksums because the prefix is different, and other currencies that may be supported in the future have completely different address formats.

alcipir:
Since it's a priority to get the server up and running again, I would leave it for further issues/updates. Instead of renaming the file, we could just add a middleware that checks the address and then instantiates the proper service for it, but we can discuss it further in the future.

chedieck:
> Sure it's the prefix but technically its the entire address. For example bch and xec have different checksums because the prefix is different, and other currencies that may be supported in the future have completely different address formats.

Okay, I made the distinction because some places we save the address unprefixed, e.g. the `PaybuttonAddress` table already points to a `chainId`, so the address is saved without the prefix.

> Since it's a priority to get the server up and running again, I would leave it for further issues/updates. Instead of renaming the file, we could just add a middleware that checks the address and then instantiates the proper service for it, but we can discuss it an ideal implementation later when implementing XEC/Chronik for example.

Okay, I am approving it then.

chedieck (approved):

Klakurka (approved):
Looks good and matches what we need on the client.

PR #111, opened 2022-06-15
feat: add transaction details route

alcipir:
Description: Related to #67. Adding the transactions details route from old custom express server to NextJS API routes.

Test Plan: `make dev`, then navigate to `http://localhost:3000/api/transaction/acb72bfd2d0e8f2b3cba34dada596c811eb2fb360c9292a9223a6aa4714fbfbb`, should display transaction details for this transaction or any other valid BCH transaction. **Remember to set a valid node on .env**: `GRPC_NODE_URL=bchd.fountainhead.cash` as an example.


chedieck:
> Should we add unit tests or integration tests for this kind of routes? Since we are already testing the service behind it, I'm not sure we need it. Would like some input from you guys here @chedieck @Gattaca1.
> 
> I want to avoid at all costs unit tests that comes down to `mock === mock` as I find them to be useless and causing unnecessary noise on our testing suite, but I might be totally wrong here so want to double check. Testing for invalid inputs/validation is fine, like in this PR: https://github.com/PayButton/paybutton-server/pull/103/files

I agree with avoid such tests that literally do `mock === mock`. As for the integration tests, I think is a good idea to add them, because running into the service it's not the only option: it's possible to test the "missing ID" message too. It's also good practice to assume we may change the code in the future, and the tests will help clarifying what is/was the expected behavior.

chedieck:
I followed the test plan but got this:
image


Also:
- Is there any reason why the route uses `transactions` in the plural and not the singular? The `address` and `paybutton` endpoints are in the singular, so if not, I think it's best to maintain the consistency

Klakurka:
If the API is grabbing 1 transaction, then the url should be `transaction` (singular). Right?

chedieck:
> @chedieck @Klakurka
> 
> This is basically re-implementing the same code from custom server, as to no break compatibility, but the correct RESTful standard is the name of the resource in the URL to be always in plural:
> 
> GET /transactions/details should give details about all your transactions (can have some sort of query param ?page= I think) GET /transactions/details/[txid] should give details about a specific transaction.
> 
> So address is the one not following the pattern here.
> 
> PS: I missed the paybutton in singular thing, maybe it was my mistake earlier on.
> 
> More information here: https://nextjs.org/docs/api-routes/dynamic-api-routes#index-routes-and-dynamic-api-routes

Alright, I created another PR #117 to rename the endpoint from `paybutton` -> `paybuttons`, beside fixing the other problem that I've had introduced of the GET request expecting a body instead of a query.

About what is currently the `api/address/transactions` endpoint, should I simply change it to `api/addresses/`? I didn't understand why the `transactions` at the end of the URI

Klakurka:
OK so just went around and asked a bunch of people and consensus is:

1. Pick a set of rules and be consistent about it.
2. Use singulars for unique stuff, plural for collections.

https://stackoverflow.com/a/30948612

So stuff like this:

api/transaction/[transaction id] - works (lists 1, singualar)
api/transactions/ - works (lists all, in an array)
api/transaction/ - throws error

Klakurka (changes requested):
OK so just went around and asked a bunch of people and consensus is:

Pick a set of rules and be consistent about it.
Use singulars for unique stuff, plural for collections.
https://stackoverflow.com/a/30948612

So stuff like this:

api/transaction/[transaction id] - works (lists 1, singualar)
api/transactions/ - works (lists all, in an array)
api/transaction/ - throws error

Gattaca1:
> OK so just went around and asked a bunch of people and consensus is:
> 
> Pick a set of rules and be consistent about it. Use singulars for unique stuff, plural for collections. https://stackoverflow.com/a/30948612
> 
> So stuff like this:
> 
> api/transaction/[transaction id] - works (lists 1, singualar) api/transactions/ - works (lists all, in an array) api/transaction/ - throws error

We can make `api/transaction` redirection to `api/transactions`

Klakurka (approved):

Issue #33, opened 2022-03-25
"Add button" feature


Klakurka:
User flow:

1. Click "Add a PayButton"
2. Enter a name (unique to this user+server) and 1 or more (multiple) addresses. Only accept BCH or XEC formats as we will use these to determine which currency was intended on each payment.
3. Generate a unique ID hash for the button that can be added to it to track payments regardless of the address used (although whichever address IS used will allow us to know the intended currency). https://github.com/PayButton/paybutton/issues/67

chedieck:
This can be closed when the four PRs above are merged.

PR #113, opened 2022-06-15
Feature: Add paybutton `name`, `uuid` and `buttonData` fields — Create parsers

chedieck:
Description:
Part of #33: Creates functions to parse the new fields, plus their unittests. In this PR, the fields still don't exist. Thus, the validators created are are still not being used.

Test plan:
All tests should pass.

### DIFF
--- a/constants/index.ts
+++ b/constants/index.ts
@@ -11,6 +11,9 @@ export const RESPONSE_MESSAGES = {
   NOT_FOUND_404: { statusCode: 404, message: 'Not found.' },
   INVALID_INPUT_400: { statusCode: 400, message: 'Invalid input.' },
   USER_ID_NOT_PROVIDED_400: { statusCode: 400, message: "'userId' not provided." },
+  NAME_NOT_PROVIDED_400: { statusCode: 400, message: "'name' not provided." },
+  NAME_ALREADY_EXISTS_400: { statusCode: 400, message: 'Button name already exists for this user.' },
   ADDRESSES_NOT_PROVIDED_400: { statusCode: 400, message: "'addresses' not provided." },
-  INVALID_CHAIN_SLUG_400: { statusCode: 400, message: 'Invalid chain slug.' }
+  INVALID_CHAIN_SLUG_400: { statusCode: 400, message: 'Invalid chain slug.' },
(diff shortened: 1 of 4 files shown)


Klakurka:
I was just thinking about this one today.

We are going to be supporting BIP 70 which already supports arbitrary data like what might go into `buttonData`. So is this fine? We don't want to be maintaining two separate fields for arbitrary data after BIP 70 is out so I assume we'll reuse this one?

Klakurka (commented):
The task was to add a client ID (to identify the app/website) and another for arbitrary data but there's 3 new ones here.

What's `userId` and `name` for?

Klakurka:
Actually maybe we can have there be a url/app field that is automatically populated with the url the button is on.

Then, we could also allow them to enter a button name + ButtonData.

Thoughts?

chedieck:
> I was just thinking about this one today.
> 
> We are going to be supporting BIP 70 which already supports arbitrary data like what might go into `buttonData`. So is this fine? We don't want to be maintaining two separate fields for arbitrary data after BIP 70 is out so I assume we'll reuse this one?

When we go on supporting BIP 70, won't this arbitrary data used by the paybutton need to be saved on the database?

> The task was to add a client ID (to identify the app/website) and another for arbitrary data but there's 3 new ones here.
> 
> What's `userId` and `name` for?

I had understand, from the issue description, that the user would input a unique name (for the user + server) for the paybutton, that's why I implemented the `name` field.
`userId` is not a field of the paybutton, there is already a field called `providerUserId`, which is the `userId` of the user who created the paybutton, and now there is the `uuid` field, which is just a universal unique identifier that is generated automatically when creating the paybutton:
image


> Actually maybe we can have there be a url/app field that is automatically populated with the url the button is on.
> 
> Then, we could also allow them to enter a button name + ButtonData.
> 
> Thoughts?

Can we know this URL without the user inputting it, in order to automatically populate this field? Maybe the paybutton would have to be used at least once in order to do that

Klakurka (approved):
Ok nevermind I thought this was about the API to the buttons which it's not.

That looks good to me.

Gattaca1 (approved):
looks good

PR #114, opened 2022-06-15
Feature: Add paybutton name, uuid and buttonData fields — Create scheme & migrations

chedieck:
**Description:**
Part of #33: Creates the migration & updates the prisma scheme, adding the new fields.

- The `uuid` field is `VARCHAR` and not `UUID` because [prisma does not support UUID](https://www.prisma.io/docs/concepts/database-connectors/mysql#native-type-mappings) for MySQL yet.
- The `buttonData` field is set to NOT NULL so that null values are saved as `'{}', and thus even empty values can be parsed as JSON.

**Test plan:**
This, on its own, breaks all previous code related to the PayButton API & DB services, as well as their tests. The changes to that code will be added in subsequent PRs that point to this one as their dependency.

However, it's possible to test the migrations and the scheme by running:
```
yarn docker migratereset
yarn docker prismagenerate
```

### DIFF
--- a/prisma/migrations/20220609133700_add_name_hash_data_fields/migration.sql
+++ b/prisma/migrations/20220609133700_add_name_hash_data_fields/migration.sql
@@ -1,0 +1,10 @@
+ALTER TABLE `Paybutton`
+ADD COLUMN `name` VARCHAR(255) NOT NULL AFTER `id`;
+
+ALTER TABLE `Paybutton`
+ADD COLUMN `uuid` VARCHAR(191) NOT NULL DEFAULT UUID() AFTER `name`;
+
+ALTER TABLE `Paybutton`
+ADD COLUMN `buttonData` JSON NOT NULL AFTER `uuid`;
+
+ALTER TABLE `Paybutton` ADD CONSTRAINT `Paybutton_name_providerUserId_unique_constraint` UNIQUE (`name`, `providerUserId`);
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -1,39 +1,46 @@
 generator client {
-  provider = "prisma-client-js"
+  provider      = "prisma-client-js"
   binaryTargets = ["native", "linux-musl"]
 }
 
 datasource db {
-  provider = "mysql"
-  url      = env("DATABASE_URL")
+  provider          = "mysql"
+  url               = env("DATABASE_URL")
   shadowDatabaseUrl = env("SHADOW_DATABASE_URL")
 }
 
 model PaybuttonAddress {
-  id          Int        @id @default(autoincrement())
-  address     String     @db.VarChar(255)
-  createdAt   DateTime   @default(now())
-  updatedAt   DateTime   @updatedAt
+  id          Int       @id @default(autoincrement())
+  address     String    @db.VarChar(255)
+  createdAt   DateTime  @default(now())
+  updatedAt   DateTime  @updatedAt
   chainId     Int
   paybuttonId Int
   chain       Chain     @relation(fields: [chainId], references: [id], onUpdate: Restrict)
-  paybutton  Paybutton @relation(fields: [paybuttonId], references: [id], onDelete: Cascade, onUpdate: Restrict)
+  paybutton   Paybutton @relation(fields: [paybuttonId], references: [id], onDelete: Cascade, onUpdate: Restrict)
+
+  @@index([chainId], map: "PaybuttonAddress_chainId_fkey")
+  @@index([paybuttonId], map: "PaybuttonAddress_paybuttonId_fkey")
 }
 
 model Paybutton {
-  id                  Int                   @id @default(autoincrement())
-  providerUserId      String?               @db.VarChar(255)
-  createdAt           DateTime              @default(now())
-  updatedAt           DateTime              @updatedAt
-  addresses           PaybuttonAddress[]
+  id             Int                  @id @default(autoincrement())
+  name           String               @db.VarChar(255)
+  uuid           String               @default(dbgenerated("uuid()"))
+  buttonData     String               @db.LongText
+  providerUserId String?              @db.VarChar(255)
+  createdAt      DateTime             @default(now())
+  updatedAt      DateTime             @updatedAt
+  addresses      PaybuttonAddress[]
+
+  @@unique([name, providerUserId], map: "Paybutton_name_providerUserId_unique_constraint")
 }
 
 model Chain {
-  id                  Int                   @id @default(autoincrement())
-  slug                String                @unique @db.VarChar(255)
-  title               String                @db.VarChar(255)
-  createdAt           DateTime              @default(now())
-  updatedAt           DateTime              @updatedAt
-  paybuttonAddresses  PaybuttonAddress[]
-
+  id                 Int                @id @default(autoincrement())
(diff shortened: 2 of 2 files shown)


Gattaca1:
This branch requires #113 before its tests will pass, correct?

prisma/schema.prisma
```diff
@@ -1,39 +1,46 @@
...
 model Paybutton {
-  id                  Int                   @id @default(autoincrement())
-  providerUserId      String?               @db.VarChar(255)
-  createdAt           DateTime              @default(now())
-  updatedAt           DateTime              @updatedAt
-  addresses           PaybuttonAddress[]
+  id             Int                  @id @default(autoincrement())
+  name           String               @db.VarChar(255)
+  uuid           String               @default(dbgenerated("uuid()"))
+  buttonData     String               @db.LongText
```
Klakurka, line 30: What were you thinking we put in here?

My original idea for `buttonData` was tied to specific payments but I'm thinking `paymentData` makes more sense for that anyway. `paymentData` would also be used as the data container for BIP 70 transactions.

prisma/schema.prisma
```diff
@@ -1,39 +1,46 @@
...
-  address     String     @db.VarChar(255)
-  createdAt   DateTime   @default(now())
-  updatedAt   DateTime   @updatedAt
+  id          Int       @id @default(autoincrement())
+  address     String    @db.VarChar(255)
+  createdAt   DateTime  @default(now())
+  updatedAt   DateTime  @updatedAt
   chainId     Int
   paybuttonId Int
   chain       Chain     @relation(fields: [chainId], references: [id], onUpdate: Restrict)
```
Klakurka, line 19: `Chain` is a valid field type?

prisma/schema.prisma
```diff
@@ -1,39 +1,46 @@
...
-  createdAt           DateTime              @default(now())
-  updatedAt           DateTime              @updatedAt
-  paybuttonAddresses  PaybuttonAddress[]
-
+  id                 Int                @id @default(autoincrement())
+  slug               String             @unique @db.VarChar(255)
+  title              String             @db.VarChar(255)
+  createdAt          DateTime           @default(now())
+  updatedAt          DateTime           @updatedAt
+  paybuttonAddresses PaybuttonAddress[]
```
Klakurka, line 45: How's this to be used? It doesn't seem like it belongs here as a property of a chain.

prisma/schema.prisma
```diff
@@ -1,39 +1,46 @@
...
 model Chain {
-  id                  Int                   @id @default(autoincrement())
-  slug                String                @unique @db.VarChar(255)
-  title               String                @db.VarChar(255)
-  createdAt           DateTime              @default(now())
-  updatedAt           DateTime              @updatedAt
-  paybuttonAddresses  PaybuttonAddress[]
-
+  id                 Int                @id @default(autoincrement())
+  slug               String             @unique @db.VarChar(255)
```
Klakurka, line 41: Is this supposed to be the currency ticker symbol / currency code?

prisma/schema.prisma
chedieck: I didn't think about anything specific. If it should be only for payments, then `paymentData` seems like a good idea, but it's less generic, so if we want to allow users to store another kind of data in the future, storing it as a key of `paymentData` would be weird, while it would be easy to store something like `buttonData = {'paymentData': ..., 'otherData': ...}`

prisma/schema.prisma
chedieck: Yes. Almost all of this was automatically generated by prisma by inspecting the database, btw

prisma/schema.prisma
chedieck: a `PaybuttonAddress` has a foreign key to a `chainId`, so this related attribute is automatically generated, it's all the `PaybuttonAddress`es that point to that chain.

prisma/schema.prisma
chedieck: Yes, e.g.:
image

prisma/schema.prisma
Klakurka: The thing is that it still makes sense IMO to put any sort of payment or button info in there if it's done at the payment level (like we do on Coin Dance). But what this is in the code is at the button level -- something that's set once -- which is different.

prisma/schema.prisma
Klakurka: Wouldn't this get exceptionally large?... it's all paybutton addresses ever? So if we only supported 1 currency, it would effectively just be every address connected to the system that people have entered?

prisma/schema.prisma
Klakurka: Ok this is a bit different than that. Can we add a column for ticker?

`BCH`
`XEC`

(testnet would have the same tickers as mainnet).

prisma/schema.prisma
chedieck: I see. I implemented this thinking about something that is done at the button level (not necessarily only once, since we will probably have an "Edit" page). If it is done in the payment level, I think we would best save it in another table, something related to payments (e.g. if we start tracking payments as rows in some "PaybuttonPayment" table)

prisma/schema.prisma
chedieck: Yes. But all this line does is make so that the PrismaClient is able to access this related field, until we explicitly run `chain.paybuttonAddresses` it's not gonna make a difference.

Also, it's not something we can take off because spelling out this relation is prisma syntax, this is what happens if I take it off:
image
There is further info [here](https://www.prisma.io/docs/concepts/components/prisma-schema/relations)

prisma/schema.prisma
chedieck: Yes, I'll address that in a separate PR.

prisma/schema.prisma
Klakurka: Yeah ok. I'm good with having both.

prisma/schema.prisma
Klakurka: Ok I'd like @ScottMcDermid or @Gattaca1 to review this part as well as I don't fully understand the implications.

ScottMcDermid: Ya looks like a harmless one-to-many association.

> Wouldn't this get exceptionally large?
Large in which way? Storage or computation? There's no added storage, since all addresses need to belong to some chain. Querying over all addresses in a chain is going to computationally difficult whether or not we're joining on the associated chain.

prisma/migrations/20220609133700_add_name_hash_data_fields/migration.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE `Paybutton`
+ADD COLUMN `name` VARCHAR(255) NOT NULL AFTER `id`;
+
+ALTER TABLE `Paybutton`
+ADD COLUMN `uuid` VARCHAR(191) NOT NULL DEFAULT UUID() AFTER `name`;
```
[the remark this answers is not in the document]
chedieck: It's some sort of default for MySQL, this is what Prisma automatically chose after inspecting the scheme changes.

Klakurka (approved):

ScottMcDermid (approved):
Looks good!

PR #115, opened 2022-06-15
Feature: Add paybutton name, uuid and buttonData fields — Change paybuttonsService

chedieck:
Description:
Part of #33: Makes the necessary changes related to the new added fields to  `paybuttonsService.ts` and its tests.

Depends on:

- [x] #113 
- [x] #114 

Test plan:
The API & tests will be addressed in another PR, so those tests will fail. `yarn docker t tests/unittests/paybuttonsService.test.ts` should succeed.

### DIFF
--- a/services/paybuttonsService.ts
+++ b/services/paybuttonsService.ts
@@ -10,9 +10,9 @@ export interface CreatePaybuttonInput {
   prefixedAddressList: string[]
 }
 
-export async function createPaybutton (userId: string, prefixedAddressList: string[]): Promise<Paybutton> {
+export async function createPaybutton (values: CreatePaybuttonInput): Promise<Paybutton> {
   const paybuttonAddressesToCreate: Prisma.PaybuttonAddressUncheckedCreateWithoutPaybuttonInput[] = await Promise.all(
-    prefixedAddressList.map(
+    values.prefixedAddressList.map(
       async (addressWithPrefix) => {
         const [prefix, address] = addressWithPrefix.split(':').map(
           (substring) => substring.toLowerCase()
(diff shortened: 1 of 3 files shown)


prisma/schema.prisma
```diff
@@ -1,39 +1,46 @@
...
-  createdAt           DateTime              @default(now())
-  updatedAt           DateTime              @updatedAt
-  paybuttonAddresses  PaybuttonAddress[]
-
+  id                 Int                @id @default(autoincrement())
+  slug               String             @unique @db.VarChar(255)
+  title              String             @db.VarChar(255)
+  createdAt          DateTime           @default(now())
+  updatedAt          DateTime           @updatedAt
+  paybuttonAddresses PaybuttonAddress[]
```
Klakurka, line 45: I feel like I already reviews this stuff in #114

chedieck: Yes, this PR depends on #114, it includes all code from there.

Klakurka: Ok I'll look at this again after it's merged into master.

Klakurka (changes requested):
Rebase with master.

chedieck:
> Rebase with master.

Done.

Klakurka (dismissed):

services/paybuttonsService.ts
```diff
@@ -10,9 +10,9 @@ export interface CreatePaybuttonInput {
   prefixedAddressList: string[]
 }
 
-export async function createPaybutton (userId: string, prefixedAddressList: string[]): Promise<Paybutton> {
+export async function createPaybutton (values: CreatePaybuttonInput): Promise<Paybutton> {
```
Gattaca1, line 13: ~~The API wasn't updated to use this new format, but no test catches that.

- Add a test that will fail (does the paybutton actually get created?)
- Update the API to use this new `createPayButton` argument format

EDIT: There is another test that fails in the integration test. After I fixed that, the integration test now fails to create the paybutton, so ignore the previous test feedback. Just update the API~~

I see that all the API stuff is added in a separate commit. This looks good to go then

Klakurka (approved):

Gattaca1 (approved):

PR #117, opened 2022-06-16
Chore/fix paybutton endpoint

chedieck:
**Description:**
Fixes the GET route for paybuttons:
- GET `api/paybutton` was expecting the userId to come in the body instead of in the query
- `api/paybutton` renamed to `api/paybuttons`

I also removed `"Should"` in the API tests descriptions, since it was not adding any relevant information

EDIT:
As decided in the telegram group, this PR brings the following behavior:
`GET api/paybutton/[id]` in the **singular**, to get **one** paybutton
`POST api/paybutton` in the **singular**, to create **one** paybutton
`GET api/paybuttons?userId=<user-id>`  in the **plural**, to get **many** paybuttons

**Test plan:**
Integration tests should run with no problems.


Klakurka (approved):
Accepted with the understanding that we'll be switching from providing userId as a parameter to pulling it from the auth'd user.

ScottMcDermid (approved):
Integration tests work for me.
Code looks good.
`it('Should...` is common for BDD-style tests, but not required.

PR #118, opened 2022-06-18
feat: change transaction/address api to /api/transactions

alcipir:
Description: change transaction/address api to api/transactions

Test Plan: `make dev` then open up `http://localhost:3000/api/transactions/bitcoincash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7q9ayphulp` for instance, remember to add a valid bchd instance on your .env or .env.local file: `GRPC_NODE_URL=bchd.fountainhead.cash` 

trying to open up `http://localhost:3000/api/transactions/` should throw an error.


Klakurka:
Can't get it working no matter what I try.

constants/index.ts
```diff
@@ -16,5 +16,6 @@ export const RESPONSE_MESSAGES = {
   NAME_ALREADY_EXISTS_400: { statusCode: 400, message: 'Button name already exists for this user.' },
   ADDRESSES_NOT_PROVIDED_400: { statusCode: 400, message: "'addresses' not provided." },
   INVALID_CHAIN_SLUG_400: { statusCode: 400, message: 'Invalid chain slug.' },
-  INVALID_BUTTON_DATA_400: { statusCode: 400, message: "'buttonData' is not valid JSON." }
+  INVALID_BUTTON_DATA_400: { statusCode: 400, message: "'buttonData' is not valid JSON." },
+  ADDRESS_NOT_PROVIDED_400: { statusCode: 400, message: 'Missing Address parameter' }
```
chedieck, line 20: To keep consistency with the rest of the `..._NOT_PROVIDED_400` errors, I think this error message would be better as `"'address' not provided"`.

pages/api/transactions/[address].ts
```diff
@@ -0,0 +1,19 @@
...
+import { getAddress } from 'services/bchdService'
+import { RESPONSE_MESSAGES } from 'constants/index'
+
+const { ADDRESS_NOT_PROVIDED_400 } = RESPONSE_MESSAGES
+
+export default async (req: NextApiRequest, res: NextApiResponse): Promise<void> => {
+  try {
+    const { address } = req.query
+    if (address === '' || address === undefined) {
+      res.status(ADDRESS_NOT_PROVIDED_400.statusCode).send({ message: ADDRESS_NOT_PROVIDED_400 })
```
Comment (chedieck) [inline pages/api/transactions/[address].ts:11]: Here it should be `...message: ADDRESS_NOT_PROVIDED_400.message})`

pages/api/transactions/[address].ts
```diff
@@ -0,0 +1,19 @@
+import { NextApiResponse, NextApiRequest } from 'next/types'
+import { getAddress } from 'services/bchdService'
+import { RESPONSE_MESSAGES } from 'constants/index'
+
+const { ADDRESS_NOT_PROVIDED_400 } = RESPONSE_MESSAGES
+
+export default async (req: NextApiRequest, res: NextApiResponse): Promise<void> => {
+  try {
+    const { address } = req.query
```
Comment (chedieck) [inline pages/api/transactions/[address].ts:9]: I think is best to wrap all of this inside a `if (req.method == 'GET') {` so the endpoint ignores all other kinds of requests

pages/api/transactions/[address].ts
```diff
@@ -0,0 +1,19 @@
...
+    if (address === '' || address === undefined) {
+      res.status(ADDRESS_NOT_PROVIDED_400.statusCode).send({ message: ADDRESS_NOT_PROVIDED_400 })
+    }
+
+    const response = await getAddress(address)
```
Comment (chedieck) [inline pages/api/transactions/[address].ts:14]: Values that come from `req.query` can be either `string` or `string[]`, so you have to address this second case or typescript will complain about it.

pages/api/transactions/[address].ts
```diff
@@ -0,0 +1,19 @@
...
+    res.status(200).send(response)
+  } catch (e) {
```
Comment (chedieck) [inline pages/api/transactions/[address].ts:16]: Here is necessary to type the `e` variable, else typescript complains about it too.

chedieck (changes requested):

chedieck:
> > `GRPC_NODE_URL=bchd.fountainhead.cash` gives 14 DNS RESOLUTION FAILED
> 
> this should be a connection error, make sure it is `GRPC_NODE_URL=bchd.fountainhead.cash`
> 
> the node could have gone offline for a bit. (accidentally edited your comment, using github app on mobile)

Indeed, I don't know how I didn't see that, it works now

pages/api/transactions/[address].ts
Comment (chedieck) [reply pages/api/transactions/[address].ts]: I don't understand this, indeed here the tests are working, but the same thing is being done in #124 and the tests there are failing (I left a comment there)

tests/integration-tests/api.test.ts
```diff
@@ -283,6 +285,23 @@ describe('GET /api/transaction/[transactionId]', () => {
...
+describe('GET /api/transaction/[transactionId]', () => {
+  const baseRequestOptions: RequestOptions = {
+    method: 'GET' as RequestMethod,
+    headers: {
+      'Content-Type': 'application/json'
+    },
+    query: {}
+  }
+
+  it('Should return HTTP 400 (Bad Request) if no address specified', async () => {
```
Gattaca1, line 304: its testing for "no transaction id", rather than "no address"

Gattaca1 (changes requested):

Klakurka:
Not providing an address takes us to the admin page 404 which should be changed at some point. We can make a new task for that.

Gattaca1:
Screenshot from 2022-06-23 14-09-54

The API should respond with a 400 error to bad requests

Klakurka (approved):
Assuming it's more than a few minutes to change, let's make a new task for an API-specific 404 reponse (or whatever for handling a bad request).

Gattaca1:
API should never respond with 404

Gattaca1 (approved):

PR #116, opened 2022-06-15
Feature: Add paybutton name, uuid and buttonData fields — Change API

chedieck:
Description:
Last part of #33: Makes the necessary changes related to the new added fields to the API endpoints and its tests.

Depends on:

- [x] #115

Test plan:
All tests should work. Requests such as:
```
curl -X POST localhost:3000/api/paybutton -H "Content-Type: application/json"  -d '{"userId": "some-uid", "name":"paybutton-name", "buttonData": "{\"foo\": \"bar\"}", "addresses":"ecash:qpz274aaj98xxnnkus8hzv367za28j900c7tv5v8pc\nbitcoincash:qpz274aaj98xxnnkus8hzv367za28j900c7tv5v8lc"}' | jq
curl -X GET 'localhost:3000/api/paybutton/1' | jq
curl -X GET localhost:3000/api/paybutton -H "Content-Type: application/json"  -d '{"userId" : "some-uid"}' | jq

```
should show the new fields `name`, `uuid` and `buttonData`

PR #119, opened 2022-06-20
Added `ticker` field to scheme, seed & initial migration

chedieck:
**Description:**
Adds the `ticker` field to the `Chain` table. Since the project is still young and there is no need to mantain a production database, I added it as a required field in the initial migration, instead of creating another migration for it.

**Test Plan:**
Run `yarn docker mr` to reset the migrations and rebuild the database. Then run all the tests with `yarn docker t`.

- [ ] Depends on #116


Klakurka (changes requested):
Let's add 'ectest' as well.

chedieck:
Waiting for #116 to merge this one.

PR #120, opened 2022-06-21
Create frontend for viewing paybuttons.

chedieck:
**Description:**
Creates a simple http://localhost:3000/paybuttons page to view paybuttons.
image
There is no front end to create paybuttons for now, they will be created in a separate PR.

**Test Plan:**
Creates some paybuttons with, e.g:
```
curl -X POST localhost:3000/api/paybutton -H "Content-Type: application/json"  -d '{"userId": "u-id", "name":"pb-example-name2", "buttonData": "{\"foz\": \"baz\"}", "addresses":"ecash:qpz274aaj98xxnnkus8hzv367za28j900c7tv5v8pc\nbitcoincash:qpz274aaj98xxnnkus8hzv367za28j900c7tv5v8lc"}' | jq
```

Then check if/how they appear in the page.

- [x] Depends on #116


Gattaca1:
can you pull master into this?

Klakurka (changes requested):
Rebase

Klakurka (changes requested):
I need some better instructions on how to add a paybutton for a user as right now I can't seem to do it manually in the db.

chedieck:
> I need some better instructions on how to add a paybutton for a user as right now I can't seem to do it manually in the db.

If you can't run the curl command because you're running Windows, you can enter the alpine docker container as root with `yarn docker nrs`, add the `curl` and `jq` utilities with `apk add curl jq` and run the `curl` command in the description from there.


> There's 1 paybutton entry in the db for me but none show up.

Probably is this entry has a `providerUserId` different from the logged in user you're using in the front-end. 


> Is the in-UI button/address creator form coming back soon? I just noticed it's missing now.

Yes, it's basically ready, I just didn't create the PR yet because I wanted to put up a PR making it so that when creating the dev environment we also create a entry in the supertokens database for a standard 'dev' user, thus making it easier to test/ work with the front-end in the future.

Klakurka:
> > I need some better instructions on how to add a paybutton for a user as right now I can't seem to do it manually in the db.
> 
> If you can't run the curl command because you're running Windows, you can enter the alpine docker container as root with `yarn docker nrs`, add the `curl` and `jq` utilities with `apk add curl jq` and run the `curl` command in the description from there.
> 
> > There's 1 paybutton entry in the db for me but none show up.
> 
> Probably this entry has a `providerUserId` different from the `userId` of the logged in user you're using in the front-end.
> 
> > Is the in-UI button/address creator form coming back soon? I just noticed it's missing now.
> 
> Yes, it's basically ready, I just didn't create the PR yet because I wanted to put up a PR making it so that when creating the dev environment we also create a entry in the supertokens database for a standard 'dev' user, thus making it easier to test/ work with the front-end in the future.

`apk` isn't a valid commant in Windows.

chedieck:
> `apk` isn't a valid commant in Windows.

You should run inside the docker node container, which is running Alpine Linux.

Gattaca1:
The test plan provides a way to create buttons via curl, but skips over how to actually log in and view those buttons. what username/password?

chedieck:
> The test plan provides a way to create buttons via curl, but skips over how to actually log in and view those buttons. what username/password?

Indeed, you can run `yarn docker dbu` to enter the user database and then run the following query there to create user `u@u.com` with password  `u` and id `u-id`:
```
INSERT INTO emailpassword_users VALUES ('u-id', 'u@u.com', '$2a$11$QOyNFFmciE6Hda1avqxxa.qLMYIJSBWn7eV8i6Ude73qu5l2/wA5e', 1652220489244);
INSERT INTO emailverification_verified_emails VALUES ('u-id', 'u@u.com');                                                                  
INSERT INTO all_auth_recipe_users VALUES ('u-id', 'emailpassword', 1652220489244);
```

I am creating a PR that will seed this DB with a default dev user, so in the future this is easier.

Klakurka (changes requested):
I know this is slightly out of scope for this task but figured we should address it. Can we switch to storing the address prefixes in the db? It's more useful to just have it in there than having to append it every time we need to work with the address.

chedieck:
> I know this is slightly out of scope for this task but figured we should address it. Can we switch to storing the address prefixes in the db? It's more useful to just have it in there than having to append it every time we need to work with the address.

I don't understand, how would you like it to be stored? Right now we are storing it in the DB by pointing each `PaybuttonAddress` to a chain that has the prefix for which that address refers to.
If you mean that we should save the string, for example, `ecash:qp...3f2` instead of `qp...3f2` in the field `address`, it's something we can do for sure, but it is indeed out of the scope of this PR; I could  open another PR addressing it.

Klakurka (approved):
We'll make a new PR for the address prefixes.

Issue #121, opened 2022-06-21
Alternative to problematic chown build command


Klakurka:
Prevents building on Windows.

PR #122, opened 2022-06-21
fix: remove USER node and chown -R on prisma

alcipir:
Description: fix for #121 **and other permissions issues**. It removes the multi-stage build since it was giving us too many permission issues to deal with, and workarounds on it, wasting time.

Test Plan: `git clean -fdx`, `docker rm -f $(docker ps -aq)`, `make dev`, then `make check-logs-dev` should show the server booting up. Prisma seeding issues will be fixed on #129.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,6 @@
-FROM node:lts-alpine as builder
-
-## Install build toolchain
+FROM node:lts-alpine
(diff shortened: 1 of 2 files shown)


Klakurka (approved):
Fixed it for me.

Klakurka (approved):
Took a while to start but it does work.

PR #124, opened 2022-06-22
fix: integration test for transaction api

amphetamarina:
Description: fix for broken integration test for transaction API.

Test Plan:  Run integration tests, all should pass.


chedieck:
Also, I see that the only test for this endpoint is for the 400 error  when there is no transaction id specified. Could you also add a test for the 200 OK case?

PR #133, opened 2022-06-23
fix: refactor transaction api and fix tests

alcipir:
Description: fix/refactor transaction API and fix tests. 

Test Plan: lint and tests should pass, should work the same as #111.


chedieck (approved):
Tested, API endpoint & tests work.

Klakurka (approved):

Issue #139, opened 2022-06-24
Create standard dev user on `supertokens` database


chedieck:
Problem:
When working on the frontend, it is necessary to create a user every time after recreating the docker containers (with `make reset-dev` for example). I usually copy & paste a SQL statement that does it; but it would be much better if this 'dev' user was created automatically, so that there is always a standard user for devs to work with.

PR #143, opened 2022-06-27
Feat: Create default supertokens user for devs

chedieck:
**Description:**
Starting the docker containers now will also create a default user: `dev@mail.com` with password `dev` and user id `dev-uid`.
Solves #139 

**Test Plan:**
Reset the docker environment with `make reset-dev`, see if the containers will run normally and then try to login in `localhost:3000` with the user described above.


Klakurka (changes requested):
Looks good but can we switch to using a domain we own for the email?

`dev@paybutton.org` / `dev` is fine.

chedieck:
> Looks good but can we switch to using a domain we own for the email?
> 
> `dev@paybutton.org` / `dev` is fine.

Addressed,

Klakurka (approved):

PR #144, opened 2022-06-27
Feat: Front-end to create Paybutton

chedieck:
Description:
Adds a primitive form in the front-end to create paybuttons.
For now there is a field for `userId`, this will be removed in a subsequent PR that changes the endpoint to authenticate the request with `supertokens` and infers the `userId`. In another PR I'll add front-end validation for the form, similar to the back-end validation that already exists.

Test Plan:
Go to http://localhost:3000/paybuttons, log in as the default developer user (see #143) and try to create some paybuttons, sending the `userId` as `dev-uid`

Depends on:
- [x] #143


chedieck:
> Can we get some logic in there to detect which address format is being used and validate it (even when somebody tries to enter one and omits the prefix)? `bch-js` / `xec-js` should be able to do this.
> 
> Basically if somebody enters an address without the prefix, which is a very common thing to do, we should still support that -- but convert it to the standardized format with the prefix for storage.
> 
> Right now we're not doing address validation so I can enter a BCH address with an `ecash:` prefix and it will still add it even though it's an invalid address.

I mentioned in the description that the front-end validation would come in a subsequent PR.
As for the back-end validation, it already exists, although is not using `bch-js` for `xec-js`. We could make an issue to perform this migration; but then I would like to understand better this example where the user could enter a BCH address with an `ecash:` prefix, how would we go about telling if it is an XEC or BCH address? AFAIK they use  the same syntax, minus the prefix.

Klakurka:
> > Can we get some logic in there to detect which address format is being used and validate it (even when somebody tries to enter one and omits the prefix)? `bch-js` / `xec-js` should be able to do this.
> > Basically if somebody enters an address without the prefix, which is a very common thing to do, we should still support that -- but convert it to the standardized format with the prefix for storage.
> > Right now we're not doing address validation so I can enter a BCH address with an `ecash:` prefix and it will still add it even though it's an invalid address.
> 
> I mentioned in the description that the front-end validation would come in a subsequent PR. As for the back-end validation, it already exists, although is not using `bch-js` for `xec-js`. We could make an issue to perform this migration; but then I would like to understand better this example where the user could enter a BCH address with an `ecash:` prefix, how would we go about telling if it is an XEC or BCH address? AFAIK they use the same syntax, minus the prefix.

That's why a lib is nice -- because you can verify it with the checksum. If you just swap out the prefixes, it'll be invalid because the checksum is calculated using the prefix as well.

Klakurka (approved):
I made a new task for the address validation: #145

Issue #141, opened 2022-06-25
Add option to deploy server API on custom domain/url


alcipir:
Add an option when starting paybutton-server to change the /api/ route (subfolder) to another domain/URL

Klakurka:
The goal with v1 here is to just make sure that the existing are able to get a response.

PR #146, opened 2022-06-29
[Issue 141] feat: optional domain for api/app and base path configuration

alcipir:
Description: Related to #141 (and will help #79), also enables changing api base path if needed (as in api.paybutton.org without subfolders for example).

Proper local environment variables available on `.env.development`

Test Plan: `make dev` should work as usual, adding different environment values on `.env` should break the website/change accordingly (e.g. if you are running on a local server or just changing `hosts` file to map to another domain. Or different subfolders.)

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,2 +1,7 @@
 DATABASE_URL="mysql://paybutton:paybutton@db:3306/paybutton"
 SHADOW_DATABASE_URL="mysql://paybutton:paybutton@db:3306/prisma-shadow-db"
+APP_NAME="PayButton"
+API_BASE_PATH=/api
+API_DOMAIN=http://localhost:3000
+WEBSITE_DOMAIN=http://localhost:3000
+WEBSITE_BASE_PATH=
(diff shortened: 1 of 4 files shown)


ScottMcDermid (changes requested):
I'm getting infinite redirects, when I attempt to load `localhost:3000`. If I pause the redirect at the right time, I get an error message.

ScottMcDermid:
> I'm getting infinite redirects, when I attempt to load `localhost:3000`. If I pause the redirect at the right time, I get an error message.

Wanted to clarify that I was using the proper local environment variables available on `.env.development`

Klakurka (changes requested):
Let's put paybutton.org in here since we don't really want to be advertising the .io domains.

ScottMcDermid:
> Thanks @ScottMcDermid, do you mind adding you to review these deployment related PRs?

Ya, for sure.

> I was going to ask if you ran docker rm -f $(docker ps -aq), but cache issues should not be happening anymore with the recent Dockerfile modification, so I will investigate.

Tried it anyways, and had the same result. The `master` branch does not have the same redirecting loop for me.

Klakurka:
Tests failing.

ScottMcDermid (approved):

Issue #142, opened 2022-06-27
Save addresses with prefixes in DB always


Klakurka:
Mentioned here: https://github.com/PayButton/paybutton-server/pull/120

PR #148, opened 2022-06-30
Chore: Save address with prefix

chedieck:
Description:
Solves #142, saves addresses in the DB with the prefix.

Test plan:
Nothing special, tests already cover this.


Klakurka (approved):

Gattaca1 (approved):
Looks good to me

PR #150, opened 2022-07-05
[Issue 7] feat: main nav style updates

johnkuney:
- Updated logo
- General style updates, added icons and hover effects
- Add placeholder pages for other menu items

Would be cool to hide the side bar if a user is not logged in...Wasnt sure the best way to check against this with supertokens?


Gattaca1 (changes requested):
I quite like this design, but my one reservation is the very low contrast on selected menu items. The bright green with white text fails colour contrast checks so lets look for a way to fix that (could be toning down the green, or switching the white to a grey or even black. I'll leave how that actually happens up to you, but lets try for a 4.5:1 ratio using a colour contrast checker (I used this one: https://webaim.org/resources/contrastchecker/)

johnkuney:
Yep good call, changed to black

Klakurka (changes requested):
- Change `PayButtons` to `Buttons` (same for the url)
- Add Telegram and Twitter links like you have in the original mockups (https://t.me/paybutton and https://twitter.com/thepaybutton)
- Add `Settings` menu item under `Account`, above `Help`.
- Hide nav when not signed in yet.

Dark mode can be done in a separate branch.

johnkuney:
okay those changes are in

Klakurka (approved):

Gattaca1 (approved):

PR #159, opened 2022-07-09
[Issue 151] feat:dashboard ui 

johnkuney:
Frontend design for the dashboard page using mock data


Klakurka (changes requested):
- Let's use `#000000` for the button text when the bg is our blue colour to hit that >7:1 contrast ratio.
- Add `1 Year, `YTD`, and `6 Month` options (can skip YTD if it's hard to do).

Nit:
- Possible to improve the chart hover so that it doesn't need to be exactly on the individual points to get the tooltip to show but anywhere in the vertical?
- Able to switch to a clearer date format for the x-axis? I'm not concerned with every single day showing up on the bottom (due to the date format being too long) since people can hover over the specific days they want.

johnkuney:
Okay made those adjustments. I tweaked the design for the buttons to be a bit cleaner I think, and added the year option and tooltip hover

Klakurka (approved):
The button colour contrast is considered to be pretty low but I'm personally fine with it.

The date format for 1W and 1M are fine IMO, clear enough.

Klakurka (changes requested):
I was thinking about mobile and IMO we can do a better job here.

Can we stack the 2 charts on mobile at a certain breakpoint? We're going to be adding lots more filters / config buttons over time and so having them be prominent at the top (just above the charts instead of below) is fine.

image

johnkuney:
For sure, added. Still need to address the menu for mobile as well, but is probably best to do that in another pr

Klakurka (approved):
https://github.com/PayButton/paybutton-server/issues/166

PR #163, opened 2022-07-11
[Issue #4] feat: create transactions table

chedieck:
Description:
Create the `Transaction` model. Also, reorganize the file for mocked objects, since some upcoming PRs will need it.

Test plan:
Nothing regarding the `Transaction` model is being  implemented on this PR, so simply the tests should pass.


prisma/migrations/20220707204448_add_transaction_model/migration.sql
```diff
@@ -0,0 +1,14 @@
+-- CreateTable
+CREATE TABLE `Transaction` (
+    `id` INTEGER NOT NULL AUTO_INCREMENT,
+    `hash` VARCHAR(255) NOT NULL,
+    `amount` VARCHAR(255) NOT NULL,
```
Klakurka, line 5: Shouldn't this be a decimal(16,2) ?

tests/mockedObjects.ts
```diff
@@ -0,0 +1,255 @@
...
+  createdAt: new Date('2022-05-27T15:18:42.000Z'),
+  updatedAt: new Date('2022-05-27T15:18:42.000Z')
+}
+
+// Transaction
+export const mockedTransaction = {
+  id: 1,
+  hash: 'Yh5DRDjd3AarAvQA1nwpPI4daDihY6hQfnMV6UKFqZc=',
+  paybuttonAddressId: 1,
+  amount: '431247724',
```
Klakurka, line 120: Assuming we change to Decimal(16,2), this will need to change to `431247724` or `431247724.00`.

Klakurka (changes requested):

prisma/migrations/20220707204448_add_transaction_model/migration.sql
chedieck: Why `(16,2)`?  I think we could do `(16.0)` and then register the default base unit steps in the `Chain` table (this would be 2 for eCash and 8 for Bitcoin Cash, for example). The BCH lib we are using to get the data just return the amount in satoshis as a string, so it would also be easy to do the conversion and we only have to worry about the decimal places when displaying this info to the user.

chedieck: Actually, `Decimal(x, 0)` looks like an overkill, we could just user INTEGER in this case to save the satoshis, what do you think?

chedieck: Addressed.

Klakurka (approved):

PR #164, opened 2022-07-11
[Issue #4] feat: create transactions services

chedieck:
Description:
Creates services for the `Transaction` and `PaybuttonAddress` model & tests for these services.

Test plan:
See if the tests are covering well the use cases of the services, then simply the tests themselves should pass.

Depends on:
- [x] #163


Klakurka (approved):
Checks failing fix attempt in https://github.com/PayButton/paybutton-server/pull/169

Issue #172, opened 2022-07-18
Refactor `Paybutton`  <-> `PaybuttonAddress` relations.


chedieck:
From one-to-many into many-to-many.

PR #173, opened 2022-07-18
[Issue #172] refactor: `Paybutton` <-> `PaybuttonAddress` relations

chedieck:
Description:
Refactor so that the relation `Paybutton` <-> `PaybuttonAddress`  is many-to-many instead of one-to-many. This together with the change that now the `address` field is unique among `PaybuttonAddress`es makes so that the same `PaybuttonAddress` might belong to multiple `Paybutton`s, and that address info (such as transactions, for example) will be centralized on that unique `PaybuttonAddress`.

Test Plan:
Besides passing the tests, the front-end for the Paybutton creation/view should work as before, at http://localhost:3000/buttons


prisma/migrations/20220715134600_init/migration.sql
```diff
@@ -0,0 +1,73 @@
+-- CreateTable
+CREATE TABLE `PaybuttonAddress` (
```
Klakurka, line 2: Is it too ambiguous to just use `Address` ?

We're thinking that we'll eventually support other buttons like Cashtab's and not just our own.

prisma/migrations/20220715134600_init/migration.sql
```diff
@@ -0,0 +1,73 @@
...
+    `providerUserId` VARCHAR(255) NULL,
+    `createdAt` DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
+    `updatedAt` DATETIME(3) NOT NULL,
+
+    UNIQUE INDEX `Paybutton_name_providerUserId_unique_constraint`(`name`, `providerUserId`),
+    PRIMARY KEY (`id`)
+) DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+
+-- CreateTable
+CREATE TABLE `PaybuttonAddressOnPaybutton` (
```
Klakurka, line 29: Maybe just `ButtonAddress` ?

Klakurka: Would want to update the other ones below, including the field names, if we change that.

Klakurka (changes requested):

prisma/migrations/20220715134600_init/migration.sql
chedieck: I thought about making this change, but didn't went through with it because I though it would be too time consuming and add a lot more changes to a PR that is already not that small. I can do it in another PR tho.

prisma/migrations/20220715134600_init/migration.sql
chedieck: [Prisma docs](https://www.prisma.io/docs/concepts/components/prisma-schema/relations/many-to-many-relations) use this `FoosOnBars` syntax (I actually implemented with the singular, but they use the plural). What about `AddressesOnButtons`?

prisma/migrations/20220715134600_init/migration.sql
Klakurka: Ok

prisma/migrations/20220715134600_init/migration.sql
Klakurka: Sounds fine to me if the others agree.

Klakurka (approved):

Gattaca1 (approved):
LGTM

PR #176, opened 2022-07-19
[Issue #4] feat: create transactions page

chedieck:
**Description:**
Creates a primitive front-end for transactions. Right now this only works for BCH addresses. The edge-case of dealing with multiple `PaybuttonAddress`es is not being dealt with in this PR; I'll create another PR addressing it after solving #172  and #175.

**Test Plan:**
I will create another PR with the integration tests. For now:
1. Create a paybutton with a bitcash address, then click on it to go to its details.
https://user-images.githubusercontent.com/21281174/179738363-376d1efe-a76d-49d6-a9b0-c1846682c738.mp4
2. Send a request to sync this bch address transactions with for example
`curl -X GET http://localhost:3000/api/transactions/sync/bitcoincash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7q9ayphulp`
3. Check that the trasactions now appear in the button detail page.
image


Klakurka (changes requested):
I get the: `{"statusCode":200,"message":"Successfully synced."}` but no results on the Transactions page.

I tried with `bitcoincash:qzqh7ej2vz26a9xaxq7capzfwgxt5gem9g8rvfxc5t`

Klakurka:
For the sync process, I'd like for us to have 2 components there.

Piece one is what you made, the api we can call to sync up while the user has been offline. This would run at login time.

Piece two would be a socket connection that subscribes to all of the relevant addresses that would execute an update (possibly just running the sync command again) in realtime.

chedieck:
> I get the: `{"statusCode":200,"message":"Successfully synced."}` but no results on the Transactions page.
> 
> I tried with `bitcoincash:qzqh7ej2vz26a9xaxq7capzfwgxt5gem9g8rvfxc5t`

Right now you have to refresh the page, have you tried doing that? This will change when we implement the socket connection you mentioned.

Klakurka:
Ya refresh doesn't work.

Klakurka:
Anyone else able to get this to work?...

johnkuney:
So I was able to get it to work fine, although on the /buttons page, the list of created buttons is disappearing after a refresh or navigating away from the page. Not sure if that is already a known issue and in the scope of this PR or not...

The buttons are still there though, and if I go to /button/1 for example I will see the button details and transactions

Is the plan to have the button details pages just be sequentially numbered? Like button/1 as opposed to using the button name or something? Just curious 

Also curious what is the userId field for when creating a button? Is that meant to be like the supertokens login userId? Like thats how each button is tied to each account? And if so should that be configurable by the user on button creation?

Klakurka:
The user ID will be set automatically for them.

Klakurka (changes requested):
- What's the 'hash' a hash of? It should be the transaction hash that actually appears on the blockchain but it's clearly of something else.\
- Transaction ordering should be by time stamp descending.
- Looks like we need to handle both incoming and outgoing txs from the button address. They'll need to be tagged as such (either `in` or `out`) and we'll need to have the actual value in there for outgoing ones rather than just showing 0.

Klakurka:
Also the button/X page was unexpected -- but we can keep it.

We just want this info onto the payments page as well, all of the added button's txs togther. Only incoming payments as well, so outgoing txs would be filtered out.

chedieck:
>  What's the 'hash' a hash of? It should be the transaction hash that actually appears on the blockchain but it's clearly of something else.

It's the same hash, but the `grpc-bchd-node` lib gives us it encoded as `base64`. I will add a commit making it so that we save it as hex.

>  Transaction ordering should be by time stamp descending. 
>   Looks like we need to handle both incoming and outgoing txs from the button address. They'll need to be tagged as such (either `in` or `out`) and we'll need to have the actual value in there for outgoing ones rather than just showing 0.

Ok.

> We just want this info onto the payments page as well, all of the added button's txs togther. Only incoming payments as well, so outgoing txs would be filtered out.

Ok, created an issue for it.

chedieck:
@Klakurka I addressed the timestamps ordering issues. For the others issues:

- I think the `base64` issue is better addressed in the fork @alcipir is making of `grpc-bchd-node`. This way we don't have to bother about changing it back to hex; we just have to not change it from hex to base64 in the first place, so I can just remove the line that does it from there.

- I had understood that the only transactions that we would save were the ones received by the address; so I also think this would be better addressed in another PR because there would have to be significant changes in the scheme and such for addressing both IN and OUT transactions, so this other PR would even depend in #173.

Klakurka:
> * I think the `base64` issue is better addressed in the fork @alcipir is making of `grpc-bchd-node`. This way we don't have to bother about changing it back to hex; we just have to not change it from hex to base64 in the first place, so I can just remove the line that does it from there.

Ok can you make a task for that?

> * I had understood that the only transactions that we would save were the ones received by the address; so I also think this would be better addressed in another PR because there would have to be significant changes in the scheme and such for addressing both IN and OUT transactions, so this other PR would even depend in [[Issue #172] refactor: `Paybutton` <-> `PaybuttonAddress` relations #173](https://github.com/PayButton/paybutton-server/pull/173).

Being aware of the txs going out is also useful as it lets us confirm our 'balance' calc is correct but also allows us to make a "balance over time" chart if we want to do that.

chedieck:
#180 #181  and #182 created.

Klakurka (dismissed):

services/transactionsService.ts
```diff
@@ -7,14 +7,23 @@ import { fetchPaybuttonAddressBySubstring } from 'services/paybuttonAddressesSer
...
     }
   })
-  return new Prisma.Decimal(totalOutput).dividedBy(10e8)
+  return new Prisma.Decimal(totalOutput).dividedBy(1e8)
 }
 
-export async function saveTransaction (transaction: BCHTransaction.AsObject, receivingAddress: string): Promise<Transaction | undefined> {
+export async function fetchAddressTransactions (address: string): Promise<Transaction[]> {
+  const paybuttonAddress = await fetchPaybuttonAddressBySubstring(address)
+  return paybuttonAddress.receivedTransactions.sort(function (a, b) {
```
Gattaca1, line 19: I'm surprised we don't have it already - lets import lodash and just use `_.sortBy()` or `_.orderyBy()` since I'm sure this is not the first or last time we'll want to sort an array of objects in this project

tests/unittests/transactionsService.test.ts
```diff
@@ -5,12 +5,12 @@ import { mockedTransaction, mockedGrpc, mockedPaybuttonAddress } from '../mocked
...
   it('Return created transaction', async () => {
-    prismaMock.transaction.create.mockResolvedValue(mockedTransaction)
-    prisma.transaction.create = prismaMock.transaction.create
+    prismaMock.transaction.upsert.mockResolvedValue(mockedTransaction)
+    prisma.transaction.upsert = prismaMock.transaction.upsert
     prismaMock.paybuttonAddress.findMany.mockResolvedValue([mockedPaybuttonAddress])
     prisma.paybuttonAddress.findMany = prismaMock.paybuttonAddress.findMany
 
-    const result = await transactionsService.saveTransaction(
+    const result = await transactionsService.upsertTransaction(
```
Gattaca1, line 13: Since the function of this.. function.. has changed, we should add another test to cover the update case.

Gattaca1 (changes requested):
overall looks good, just a couple nit picks

tests/unittests/transactionsService.test.ts
chedieck: This test actually covers both cases: since it's an unittest, the DB is ignored and the prisma upsert function is just mocked to return the updated / created object. The object is neither updated or created, of course, because this test is run without the DB being up.

services/transactionsService.ts
chedieck: Addressed.

Gattaca1 (approved):

Klakurka (approved):

Issue #161, opened 2022-07-10
Update favicon and HTML title


Klakurka:
For the HTML title, I'm fine with either just `PayButton` or `{ page name } | PayButton`. Your choice.

PR #179, opened 2022-07-19
[Issue #161, #162] Add favicon, remove supertokens branding

johnkuney:
Added favicon to a public directory and removed supertokens branding through the frontend config


Klakurka (changes requested):
Looks good.

Can we also remove scrolling on the login page? The `margin-top:100px` is the cause. Perhaps we can change that to a `padding-top:100px`? I'd say we can remove the `margin-top` and just change the `padding-top:25px` on `main` to `100px` (125px would make it the same as it is now but 100 is probably fine -- your call).

johnkuney:
yep good call, updated

chedieck (approved):
Looks good

pages/_app.tsx
```diff
@@ -32,9 +33,15 @@ function App ({ Component, pageProps }: AppProps): React.ReactElement | null {
     return null
   }
   return (
-    <ErrorBoundary>
-        <Component {...pageProps} />
-    </ErrorBoundary>
+    <>
+      <Head>
+        <title>Paybutton</title>
```
Klakurka, line 38: Should be capital `B`.

Issue #137, opened 2022-06-24
Fix excessive processing / system load at startup


Klakurka:
After running `docker-compose up --build -d` on Windows, there's >2 minutes of high CPU usage. After that's over, there's another 10 seconds or so before the server is actually up.

Klakurka:
This no longer seems to happen for me.

Klakurka:
Was clearly fixed at some point last year.

PR #185, opened 2022-07-25
[Issue #137] chore: proper multi-stage docker build

marinacabrera:
Description: related to #137. This improves considerably the efficiency of our containers, using https://github.com/wagoodman/dive, it shows a 99% optimization compared to our previous Dockerfile. The main point, though, is caching dependencies and making it easier to change branches and develop locally without the need to reinstall or recompile sometimes. 

Added an entrypoint.sh script for the container itself to run, instead of relying on multiple chained commands on a docker-compose file.

Test Plan: `docker rmi -f paybutton-server` (removes an existing paybutton-server image if present), `make dev`. First install will be the longest. Subsequent `make dev` runs should NOT take longer than 15s as all dependencies should be cached. 

Unfortunately, will need help from many reviewers as possible on different Operating Systems as possible, as this might have issues that I'm not aware/can't properly test.


johnkuney:
So I ran into an issue with the yarn lockfile. It gave the error:
'error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.'

So I tried yarn install but then got:
'error graphql@16.5.0: The engine "node" is incompatible with this module. Expected version "^12.22.0 || ^14.16.0 || ^16.0.0 || >=17.0.0". Got "14.15.3"

So my node version not working with that graphql version. But tried 'yarn install --ignore-engines' instead and that worked.

Then did 'make dev' and the app ran fine

But it did change the yarn.lock file. I tried deleting the changes but then got the original error again...

Gattaca1:
Screenshot from 2022-08-23 10-30-31
I also get the frozen lockfile error, here is a screenshot of my console during build

Gattaca1:
Lockfile issue is fixed, however I still have the db starting issue. Picture attached 
Screenshot from 2022-08-25 15-12-35

chedieck:
Another thing that may be related: after testing the branch, when switching back to master, I was getting this error:
```
[Error: EACCES: permission denied, unlink '/home/node/src/.next/server/_error.js'] {
  errno: -13,
  code: 'EACCES',
  syscall: 'unlink',
  path: '/home/node/src/.next/server/_error.js'
}
```

solution was to delete the `.next` folder. It looks like it is being created with root as owner.

Klakurka:
Can re-open if we want to pursue this.

Issue #168, opened 2022-07-12
Address/Balance API not working


Klakurka:
https://api.paybutton.org/address/balance/bitcoincash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7q9ayphulp should work

Klakurka:
Just want to get prod updated and tested before closing this.

PR #186, opened 2022-07-25
[Issue #168] feat: add balance api

amphetamarina:
Description: Related to #168. Adds a balance endpoint, with an alias for compatibility with the old API version.

Test Plan: `make dev`, then:

`http://localhost:3000/address/balance/bitcoincash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7q9ayphulp`
`http://localhost:3000/api/balance/bitcoincash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7q9ayphulp`

Both endpoints should work.

### DIFF
--- a/pages/api/balance/[address].ts
+++ b/pages/api/balance/[address].ts
@@ -34,7 +34,7 @@ export default async (req: NextApiRequest, res: NextApiResponse): Promise<void> 
     try {
       const address = parseAddress(req.query.address as string)
       const response = await getBCHBalance(address)
-      const balance = satoshisToUnit(new Decimal(response), xecaddr.Format.Cashaddr)
+      const balance = await satoshisToUnit(new Decimal(response), xecaddr.Format.Cashaddr)
       res.status(200).send(balance)
     } catch (err: any) {
       switch (err.message) {


Klakurka:
For #168

Klakurka:
We do get an output but I'm not sure how to best test this against the client without running it live for a bit.

chedieck:
Some questions/remarks:
1. Isn't there a fast way to check the balance than adding all transactions outputs ourselves? If we try to run the endpoint for some address with a lot of transactions, such as `qzkyhguy920rs3l93kcd27sl9a8ft5hx9g2nvg8e22`, the request takes forever.
2. I think we should return the value as a decimal, and not in satoshis, as we are doing in the DB. I have used `return new Prisma.Decimal(satoshis).dividedBy(10e8)` for that end.

chedieck (changes requested):
Some questions/remarks:
1. Isn't there a fast way to check the balance than adding all transactions outputs ourselves? If we try to run the endpoint for some address with a lot of transactions, such as `qzkyhguy920rs3l93kcd27sl9a8ft5hx9g2nvg8e22`, the request takes forever.
2. I think we should return the value as a decimal, and not in satoshis, as we are doing in the DB. I have used `return new Prisma.Decimal(satoshis).dividedBy(10e8)` for that end.

pages/api/balance/[address].ts
```diff
@@ -0,0 +1,47 @@
...
+
+      return resolve(result)
+    })
+  })
+}
+
+export default async (req: NextApiRequest, res: NextApiResponse): Promise<void>  => {
+    await runMiddleware(req, res, cors)
+  if (req.method === 'GET') {
+    const address = req.query.address as string
```
Comment (chedieck) [inline pages/api/balance/[address].ts:30]: also, probably best to make this line `const address = parseAddress(req.query.address as string) ` so that the default parsing is executed

Klakurka:
I'll ask if there's a call we can make to return the balance

Klakurka:
Doesn't look like BCHD has any way to get the address balances on its own unfortunately.

Klakurka (approved):
I didn't test that this was working for the client but even if it's not working, it gets us closer to the goal. We can test it either live or once https://github.com/PayButton/paybutton/pull/156 is merged.

Klakurka (changes requested):
Let's just make sure tests are passing before merging.

Klakurka (dismissed):

utils/validators.ts
```diff
@@ -14,7 +14,7 @@ import xecaddr from 'xecaddrjs'
 /* Validates the address and adds a prefix to it, if it does not have it already.
  */
 export const parseAddress = function (addressString: string): string {
-  if (addressString === '') throw new Error(RESPONSE_MESSAGES.ADDRESS_NOT_PROVIDED_400.message)
+  if (addressString === '' || addressString === undefined) throw new Error(RESPONSE_MESSAGES.ADDRESS_NOT_PROVIDED_400.message)
```
chedieck, line 17: Since this check is being made, shouldn't the typing for the function be `... = function (addressString: string | undefined)` ?

pages/api/balance/[address].ts
```diff
@@ -0,0 +1,45 @@
...
+    })
+  })
+}
+
+export default async (req: NextApiRequest, res: NextApiResponse): Promise<void> => {
+  await runMiddleware(req, res, cors)
+  if (req.method === 'GET') {
+    try {
+      const address = parseAddress(req.query.address as string)
+      const response = await getBCHBalance(address)
```
Comment (chedieck) [inline pages/api/balance/[address].ts:33]: There is now a utility function to convert from satoshis to unit value, check `utils/index.ts`'s `satoshisToUnit`

chedieck (changes requested):

Klakurka (dismissed):

pages/api/balance/[address].ts
```diff
@@ -0,0 +1,49 @@
...
+      const balance = satoshisToUnit(new Decimal(response), xecaddr.Format.Cashaddr)
```
Comment (chedieck) [inline pages/api/balance/[address].ts:37]: `await` statement missing here, making the response return an empty JSON instead of the balance

Gattaca1 (approved):

chedieck (approved):

PR #178, opened 2022-07-19
[Issue #175] refactor: rename paybutton address

chedieck:
**Description:**
Basically, I ran two commands for string replace across the code:
`'paybuttonAddress' -> 'address'`
`'PaybuttonAddress' -> 'Address'`
... and then solved some minor problems like repeated constant names.

**Test:**
There should be no changes in any functionality.

Depends on:

- [x] #173 
- [x] #176

PR #188, opened 2022-07-26
[Issue #181] refactor: save tx hash as hex

chedieck:
Description:
Solves #181

Test plan:
See #176 test plan. The only difference is that the hash should be encoded in hex, and not base64.
You can check if `https://www.blockchain.com/bch/tx/HEX_HASH` works.

Depends on:

- [x] #178

### DIFF
--- a/components/Paybutton/PaybuttonDetail.tsx
+++ b/components/Paybutton/PaybuttonDetail.tsx
@@ -10,7 +10,7 @@ export default ({ paybutton }: IProps): FunctionComponent => {
       <h3>{paybutton.id} {paybutton.name}</h3>
       <ul>
         {paybutton.addresses.map(item => (
-          <li key={item.address}> {item.address}</li>
+          <li key={item.address.address}> {item.address.address}</li>
         ))}
         {paybutton.buttonData}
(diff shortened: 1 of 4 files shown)


Klakurka (approved):

Gattaca1 (changes requested):
We should probably just use a library for converting from base64 to base16. I checked out a few online conversion tools and it looks like your test case is not correct

chedieck:
> We should probably just use a library for converting from base64 to base16. I checked out a few online conversion tools and it looks like your test case is not correct

I think it's overkill to add another dependency just to do this conversion, and it would probably not work very well because we have to reverse the byte orders before doing the conversion. That is also why the online conversion tools you tested are giving different results, because we are getting the transaction info with the option `reverseHashOrder: true`, as recommended in the [grpc-bchrpc-node readme](https://github.com/simpleledgerinc/grpc-bchrpc-node):

> We use reversedHashOrder: true to automatically reverse the txid endianness because BCHD expects to receive transaction hashes without endianness reversed.

Gattaca1 (approved):

PR #190, opened 2022-07-27
[issue #165] feat:dark mode

johnkuney:
Adding dark mode toggle.
*Known issue, the charts on the dashboard page don't update colors when the button is switched...They will if you refresh or toggle the timeframe, but the current view isnt updating because the chart is already rendered and needs to be re-rendered...I could not figure out a good way to trigger this rerender, so if any ideas let me know.


Klakurka (changes requested):
- Why the logo image change?
- As you said, the issue with the text within the charts is an issue. You could force a reload of the charts. Notice how switching from 1M to 1W fixes it too without having to reload the entire page.

Everything else looks good to me.

johnkuney:
For the image i just updated it to add transparency to the "pay" part of it so the different background colors will show through and we can just use one image with a css filter to switch between light and dark mode. 

Yep I'll see if I trigger a chart reload

johnkuney:
Okay charts should update with theme toggle now. 

I had to rearrange some of the layout, as it made more sense to bring the page component back to the top level and hide the nav bar if not logged in

But now the nav bar is not visible on all the pages except /buttons and /. Just need to make the other pages protected pages and that will be fixed. But figure do that in another PR

Klakurka (changes requested):
Can we do it in this PR? I'd prefer we avoid regressions where possible.

Klakurka:
Or you could create a new PR that relies on this one, forked from this one.

johnkuney:
Yep fair enough, added here should be good now

Klakurka (dismissed):

Gattaca1 (changes requested):
- The contrast between the text and background on the 'Submit' Button on the Buttons page is much too low
- The text in the input boxes on the Buttons page is also still black which is quite hard to see

johnkuney:
for sure updated

Klakurka (approved):

PR #192, opened 2022-07-29
[Issue #191] feat: seed mock paybuttons

chedieck:
Description:
Creates some PayButtons automatically, to simplify developing & testing.
image

Test Plan:
Recreate the containers with `make reset-dev` or simply reset the migrations with `yarn docker mr` and check if the following buttons were created for the default dev user.

### DIFF
--- a/prisma/seed.ts
+++ b/prisma/seed.ts
@@ -1,5 +1,6 @@
 import { PrismaClient } from '@prisma/client'
 import { chains } from './seeds/chains'
+import { paybuttons, addresses, connectors } from './seeds/paybuttons'
 import { createDevUserRawQueryList } from './seeds/devUser'
 const prisma = new PrismaClient()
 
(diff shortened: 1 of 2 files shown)


Klakurka (changes requested):
`http://localhost:3000/api/transactions/sync/bitcoincash:qzqh7ej2vz26a9xaxq7capzfwgxt5gem9g8rvfxc5t` should work right away but doesn't for me somehow.

`{"statusCode":500,"message":"3 INVALID_ARGUMENT: invalid address"}`

chedieck:
> `http://localhost:3000/api/transactions/sync/bitcoincash:qzqh7ej2vz26a9xaxq7capzfwgxt5gem9g8rvfxc5t` should work right away but doesn't for me somehow.
> 
> `{"statusCode":500,"message":"3 INVALID_ARGUMENT: invalid address"}`

Weird, I just tried that after running `make reset-dev` and it worked fine. Did you do that too? I think depending on which branch you are coming from simply `yarn docker mr` is not gonna work. Also, did you have the `.env` file with the `GRPC_NODE_URL`?

Klakurka (approved):
Forgot to connect to a BCHD node.

Maybe we can change the error message to indicate that the node isn't connected instead of invalid address?

Gattaca1 (approved):

Issue #200, opened 2022-08-08
Fix seed addresses networkId


chedieck:
The seed addresses defined in `prisma/seeds/paybuttons.ts` are both indicating `networkId: 1`, but one is for XEC and the other for BCH.

PR #201, opened 2022-08-08
[Issue #200] fix: seeds networkId

chedieck:
**Description:**
Solves #200 by making `ecash` the network with `id == 1` and the BCH seed address to have `networkId: 2`

**Test plan**:
Nothing should change.


Klakurka (approved):
Running `sync` on these the addresses don't work for me but I assume this is expected until it's all merged into #196

Issue #177, opened 2022-07-19
Implement automatic transactions sync


chedieck:
For the sync process, I'd like for us to have 2 components there.

Piece one is what you made, the api we can call to sync up while the user has been offline. This would run at login time.

Piece two would be a socket connection that subscribes to all of the relevant addresses that would execute an update (possibly just running the sync command again) in realtime.

_Originally posted by @Klakurka in https://github.com/PayButton/paybutton-server/issues/176#issuecomment-1189302492_

Klakurka:
Related to https://github.com/PayButton/paybutton-server/issues/88.

PR #203, opened 2022-08-09
[Issue #177] feat: add logic to sync transactions after login

chedieck:
**Description:**
Part of #177 
Added signin / signup callbacks, as described in [the supertokens docs](https://supertokens.com/docs/thirdpartyemailpassword/common-customizations/handling-signinup-success).
Post login now triggers the syncing of the transactions related to each one of the logged in user's addresses.

**Test plan:**
With the BCH / XEC nodes URLS defined in the `.env` file such as:
```
GRPC_BCH_NODE_URL=bchd.fountainhead.cash:443    
GRPC_XEC_NODE_URL=grpc.fabien.cash:8335         
```
...do:
1. `make reset-dev`
2. Log in into the default dev user on http://localhost:3000
3. Go to buttons, select any of them. All their transactions should appear on the button detail page (the syncing of the transactions may take some seconds, from the moment of the login, to happen).


Klakurka (changes requested):
ecash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7quss2vxek doesn't seem to sync for me (no txs showing)

I also can't run the syncs manually, getting a 404.

chedieck:
> ecash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7quss2vxek doesn't seem to sync for me (no txs showing)

I think this is that same bug from before on the XEC GRPC client; I have confirmed that the problem is **not** on the node and am working on fixing it.

> I also can't run the syncs manually, getting a 404.

Indeed I removed the sync endpoint.
Independently from the bug aforementioned, should the post-login hook parse if the sync was indeed successful and then do something about it? This would probably be a separate task but it will be good to think about how to parse it.

Klakurka (changes requested):
It definitely at least works in my testing but I'll leave the technical review / pondering alternatives to better men.

Unit tests are failing though.

chedieck:
Fixed the unittests, also removed `rewire` for good (it was causing more problems than it was solving).

Klakurka (approved):

chedieck:
> @chedieck please look into #208 first as it will remove the need for these spyOn lines I believe.

Indeed, I just tried merging this branch into #208 and there was this single conflict: by selecting your new changes (thus removing all the `spyOn` code) the tests still all passed and the sync still worked.

config/backendConfig.ts
```diff
@@ -64,7 +66,70 @@ export const backendConfig = (): TypeInput => {
...
+              // override the email password sign up API
+              emailPasswordSignUpPOST: async function (input) {
+                if (originalImplementation.emailPasswordSignUpPOST === undefined) {
+                  throw Error('Should never come here')
+                }
+                // pre sign up logic goes here
+
+                const response = await originalImplementation.emailPasswordSignUpPOST(input)
+
+                if (response.status === 'OK') {
```
Gattaca1, line 84: Empty If statement here, any good reason to leave it in?

Gattaca1 (changes requested):

config/backendConfig.ts
chedieck: > Empty If statement here, any good reason to leave it in?

I think so, this was basically copied and pasted from their docs, so it's kind of a template on which we can later implement pre/post login hooks.

Klakurka: Sounds reasonable to me.

Klakurka (approved):

Issue #160, opened 2022-07-10
Add link to docs


Klakurka:
Add a link to `docs.paybutton.org`.

Either under the 'Help' page or at the bottom fo the nav next to the Telegram icon.

johnkuney:
merged

PR #204, opened 2022-08-09
[Issue #160] feat: add docs link to navbar

johnkuney:
Adding a link to docs.paybutton.org in the navbar


Klakurka (changes requested):
Nit: I think it's a different (darker) grey than the twitter and telegram icons are (light mode).

Also tooltips might be nice on these buttons.

johnkuney:
ah good catch yeah should be slightly off-black, and yeah good call added some tooltips

Klakurka:
The dark mode toggle is also using the darker colour if you think they should all be the same.

Klakurka (approved):
I think it's good as it is.

Issue #127, opened 2022-06-22
Inconsistent rules being applied for testing, compiling and linting.


alcipir:
`ts-standard` seems to be following a different set of rules than jest (and possibly next.js)

as it is not finding common type errors that jest complains.

PR #206, opened 2022-08-11
[Issue #127]: add eslint config and fix actions

amphetamarina:
Description: Removes ts-standard and adds eslint with the same rules. In practice is the same thing, but it won't randomly break tests and is compatible with more recent versions of Typescript.

Fixes github action to accomodate those changes as well.

#127 

Test Plan: change a .ts or .tsx file (preferable: a test one), commit and see if lint-staged is working. Check if unit tests and integration tests are working properly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ check-logs-users:
 	docker logs -f paybutton-users-service
 
 lint:
-	npx --yes ts-standard .
+	yarn eslint .
 
 lint-master:
 	$(git_diff_to_master)


ScottMcDermid (commented):
The `lint` make routine still uses `ts-standard`. Do we want both `lint` and `lint-master` to use the new `eslint` command?

Gattaca1 (approved):
Works well locally

ScottMcDermid (approved):

PR #208, opened 2022-08-14
test: mock GrpcClient from grpc-bchrpc-node

alcipir:
Description: related to #169 #195 and many others. This PR adds a mock for GrpcClient, enabling more test cases for the API, removing the need for the `rewire` library or other workarounds.

Test Plan: `make dev`, then `yarn docker test`. All tests should pass. No network integrity error should appear in the logs.

Issue #195, opened 2022-07-31
Don't require address to be saved in the system to respond


Klakurka:
Currently we're returning: `{"statusCode":500,"message":"No address found."}`

PR #209, opened 2022-08-14
[Issue #195]: Don't require address to be saved in the system to respond

alcipir:
Description: fix for #195.

Test Plan: `make dev`, `yarn docker test`. Tests should pass. 
After setting up `.env` with GRPC_NODE_URLs, querying the API directly should return transactions even if this address is not found in the database. This is necessary as the current paybutton clients won't be able to query it.

Example URL for testing: http://[localhost:3000/api/transactions/ecash:qzj37rjup4adk60yn2md68ctr034j2g6g5mzlw0lw7](http://localhost:3000/api/transactions/ecash:qzj37rjup4adk60yn2md68ctr034j2g6g5mzlw0lw7)

Depends on #208


Gattaca1:
#208 is merged, you can pull it in now

Klakurka (approved):

Issue #131, opened 2022-06-23
Unused and heavy dependencies related to frontend


alcipir:
We have several unused dependencies that takes some time to install through yarn:
```
   "@data-driven-forms/mui-component-mapper": "^3.16.10",
    "@data-driven-forms/react-form-renderer": "^3.16.10",
    "@emotion/react": "^11.8.2",
    "@emotion/styled": "^11.8.1",
    "@mui/icons-material": "^5.5.1",
    "@mui/material": "^5.5.3",
```

PR #210, opened 2022-08-14
[Issue #131]: remove unused dependencies mui-material and data-driven-forms

alcipir:
Description: cleaning up some unused dependencies.

Related to #131.

Test Plan: `make stop-dev`, rebuild paybutton image without cache to be sure (might take a while): `docker-compose build --no-cache paybutton`, then `make dev`. Frontend should stay the same.


Klakurka (approved):
Build is definitely a bit faster for me and goes through fewer steps.

It still does the high-cpu thing for a while the first time but it's definitely better than master.

Gattaca1 (approved):

Issue #155, opened 2022-07-06
v1 Networks


Klakurka:
- [x] Show each network that we have in the db, including testnets.
- [x] Show "Last block X minutes ago" or "Disconnected" if we're not hooked up to it.

PR #213, opened 2022-08-17
[Issue #155] feat: Display networks

chedieck:
Description:
Shows the available networks on http://localhost:3000/networks

Test plan:
http://localhost:3000/networks should display:
image


Klakurka (changes requested):
Can we add a config option (default off) to show test networks?

chedieck:
> Can we add a config option (default off) to show test networks?

Addressed, added a `SHOW_TEST_NETWORKS` env var which defaults to  `false` if not present.

Klakurka (approved):

PR #214, opened 2022-08-17
Feat/buttons UI

johnkuney:
General updates of buttons page ui
- style updates
- modal for button creation
- transaction table


Klakurka (approved):

Klakurka (changes requested):
Let's use commas on the amounts.

Klakurka:
ID and time stamp can be updated to something better at some point.

johnkuney:
Okay added

Klakurka (changes requested):
Still an issue for me:

image

Should format the numbers so that 2 decimals are always showing and also using commas.

johnkuney:
hmm okay. Are you seeing these transactions on the dev account? Cause I actually dont see any, so was using some mock data instead but must be strings instead of numbers or something. Will get it sorted though. 

Another thought, should I make the transaction ids link to that transaction on the block explorer (and for bch ones maybe blockchair?)

Klakurka:
Maybe we can get a 2nd reviewer to confirm what they see.

For the transactions, let's use the tx hashes instead of an "ID" as I'm not sure what that is.

johnkuney:
Okay was able to get the transactions working. Yeah my bad, hash makes more sense. Added number formatting, ticker based on address, and links to explorers

components/Paybutton/paybutton.module.css
```diff
@@ -0,0 +1,193 @@
...
+}
+
+.paybutton_card,
+.paybutton_nocard {
+  background-color: var(--secondary-bg-color);
+  border-radius: 10px;
+  padding: 20px;
+  margin-bottom: 15px;
+  border: 1px solid var(--secondary-bg-color);
+  transition: all ease-in-out 200ms;
```
chedieck, line 15: I find this `200ms` too slow

chedieck (changes requested):
LGTM, I just left a remark for the animation transition that I found slow, don't know if others agree

Klakurka (changes requested):
- time stamp should be a date, probably use the the user's local date format or whatever we're using on the site already if we are.
- would be nice if the hash column extended out to show the entire thing without the ellipsis if there's room (on larger screens) so people can more easily copy+paste them. OR add a copy button.

johnkuney:
For sure, can add those 
And sure animation could be a bit quicker, changed timing 100ms

johnkuney:
okay added, times should be local, using moment.js

Klakurka (approved):

Issue #151, opened 2022-07-06
v1 Dashboard


Klakurka:
Display:

- # of buttons
- USD revenue (lifetime)
- # of transactions (lifetime)
- USD revenue (past 30 days)
- # of transactions (past 30 days)
- USD revenue (past 7 days)
- # of transcations (past 7 days)

For the USD revenue numbers, ideally we should be using the market prices on the days of each tx but we can use the current price if we need to for now.

PR #218, opened 2022-08-24
[Issue #151] feat: Create dashboard API endpoint

chedieck:
Description:
Create an API endpoint to retrieve data to be used in the dashboard.

Test plan:
This is not implemented in the front end yet (will be done in a separate PR), but it can be tested by running:
 `curl http://localhost:3000/api/dashboard | jq`

Remarks:
- Dummy prices are being used instead of fetching the actual prices at the time of the transaction, which will be addressed in a future PR.

### DIFF
--- a/pages/api/dashboard/index.ts
+++ b/pages/api/dashboard/index.ts
@@ -1,0 +1,134 @@
+import * as paybuttonsService from 'services/paybuttonsService'
+import * as addressesService from 'services/addressesService'
+import { Prisma, Transaction } from '@prisma/client'
+import moment from 'moment'
+
+const DUMMY_XEC_PRICE = 0.00003918
+const DUMMY_BCH_PRICE = 132
+
+interface ChartData {
+  labels: string[]
+  datasets: [
+    {
+      data: number[] | Prisma.Decimal[]
+      borderColor: string
+    }
+  ]
+}
+
+interface PeriodData {
+  revenue: ChartData
+  payments: ChartData
+  totalRevenue: number
+  totalPayments: number
+}
+
+interface DashboardData {
+  thirtyDays: PeriodData
+  year: PeriodData
+  sevenDays: PeriodData
+  total: {
+    revenue: Prisma.Decimal
+    payments: number
+    buttons: number
+  }
+}
+
+const getChartLabels = function (n: number, periodString: string, formatString = 'M/D'): string[] {
+  return [...new Array(n)].map((i, idx) => moment().startOf('day').subtract(idx, periodString).format(formatString))
+}
+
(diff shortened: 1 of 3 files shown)


johnkuney:
Overall looks good to me. Couple of things though
- Are the dummy prices just placeholder and price calls are coming in another PR? 
- I tried adding a new button but it didnt add the transactions for it...it changed the total button count, but the payments didnt change even though the address I added has some transactions in august. Although its also not showing transactions in the button view for the new button. Only for the dev buttons...is this expected and a work in progress?

chedieck:
>    Are the dummy prices just placeholder and price calls are coming in another PR?

Yes, indeed. Forgot to mention in the description (now updated).
 
>   I tried adding a new button but it didnt add the transactions for it...it changed the total button count, but the payments didnt change even though the address I added has some transactions in august. Although its also not showing transactions in the button view for the new button. Only for the dev buttons...is this expected and a work in progress?

Right now, the transactions are only updated at login time. Try logging out and back again, see if they appear.

johnkuney:
Thanks for the clarification. ah thats right we want to reflect the usd values at the time for each transaction

Logging out and in didnt change it for me... Can I test with any xec or bch address?

chedieck:
> Logging out and in didnt change it for me... Can I test with any xec or bch address?

Aren't you missing the `.env` file with the XEC and BCH addresses?

johnkuney:
oh shoot my bad. Yes it works when I config the .env. Is there a reason we havent added those to the codebase by default? Whenever I switch branches I have to remember to add them in

But yeah working better. Although I dont think its showing every payment from my test addresses. Theyre both ecash addresses and only have very small transaction amounts in them. Is there like a minimum amount that gets filtered out or something and doesnt register?

Klakurka:
> oh shoot my bad. Yes it works when I config the .env. Is there a reason we havent added those to the codebase by default? Whenever I switch branches I have to remember to add them in
> 
> But yeah working better. Although I dont think its showing every payment from my test addresses. Theyre both ecash addresses and only have very small transaction amounts in them. Is there like a minimum amount that gets filtered out or something and doesnt register?

They aren't our APIs so not a good idea to put them in as defaults.

Klakurka (approved):

chedieck:
> Although I dont think its showing every payment from my test addresses. Theyre both ecash addresses and only have very small transaction amounts in them. Is there like a minimum amount that gets filtered out or something and doesnt register?

I think this is the bug I have addressed on #220. Can you try it there?

johnkuney:
Well actually the button view was showing all the transactions, but the payment count in the dashboard api was only going up by one. Either way, Im sure there will be gremlins to sort out, probably better to get this and #220 in, and get the dashboard frontend tied in, and then do more testing

johnkuney (approved):

PR #221, opened 2022-08-26
[Issue #151] feat: use real data on dashboard front end

chedieck:
Description:
Connects the "Dashboard" view in the frontend with the dashboard API endpoint from #218. Furthermore, the endpoint now checks the user session and gets the `userId` from it.

Test Plan:
Go to http://localhost:3000/dashboard, see if the data being displayed corresponds to the expected data, considering the user's paybuttons.
Also, `curl http://localhost:3000/api/dashboard | jq` should not work anymore (it should show "unauthorized"), since there is not going to be an active session for this request.


johnkuney:
Looks great!
- Can you round the revenue numbers to two decimals?
- We did have another color for the revenue line (#669cfe). It helped differentiate the two charts. Can we get that back in?

Klakurka (changes requested):
Same feedback as John plus:

- Possible to set the minimum values on the charts to 0 rather than allowing them to go negative? Probably out of scope.
- On the 1-year chart, there are a bunch of negative values... where are those coming from?

johnkuney:
For the chart, the y-axis is automatically configured by the plugin based on the dataset range you feed it. So can't really mess with that. 

I believe the negative numbers are coming from the dev addresses have some large send amounts in them. It looks to line up with the transaction table dates

That brings up a point though, we probably dont want to include send amounts in this do we? Users probably only want to see how much has come into the address wouldnt they? Like they may collect payments through the button and then send the balance somewhere else every month or whatever, but then it would reflect a negative or 0 revenue in the dashboard....

Klakurka:
> That brings up a point though, we probably dont want to include send amounts in this do we? Users probably only want to see how much has come into the address wouldnt they? Like they may collect payments through the button and then send the balance somewhere else every month or whatever, but then it would reflect a negative or 0 revenue in the dashboard....

Yep. Only incoming payments should show up here.

chedieck:
>    - Can you round the revenue numbers to two decimals?
>    - We did have another color for the revenue line (#669cfe). It helped differentiate the two charts. Can we get that back in?
>    - Yep. Only incoming payments should show up here.

All addressed.

Klakurka (approved):
Really cool -- this is actually pretty useful just by itself.

johnkuney (approved):
Awesome, looks good

Issue #222, opened 2022-08-26
Don't send every tx for an address on each poll to the client


Klakurka:
It's a lot of data when there are excessive txs to an address.

eg. http://localhost:3000/address/transactions/bitcoincash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7q9ayphulp

chedieck:
Considering the [parameters](https://www.grpc.cash/docs/ts/classes/grpc_bchrpc_browser.grpcclient.html#getaddresstransactions) that we can use when calling `getAddressTransactions`:
![image](https://user-images.githubusercontent.com/21281174/187698994-0d8caad1-d5ca-434c-b6a6-41373075ee67.png)


My idea on how to implement this is:
- Add a column `blockHeight` to the `Transaction` table.
- When syncing the transactions for address A:
  - Let H be the greatest `blockHeight` found in a `Transaction` that points to A.
  - Let S be the number of `Transaction` objects that point to A and have `blockHeight == H`.
  - Call `getAddressTransactions` on the address with `nbSkip: S`, `heigth = H -1` and `nbFetch = 50` 



This way, each call to the endpoint should insert 50 more transactions in the database, at a time.

chedieck:
Update: there is already a default for getting at most 50 transactions for a `bitcoincash` address request and 100 transactions for a `ecash` address request (the most recent ones).

E.g: Running `curl http://localhost:3000/address/transactions
/qr8ger8kn2fz5cr73cp7ylkqznauyjyzuqhrr7863u | jq '(.confirmedTransactionsList) | length'` gives us `100`, even though this ecash address has +20k txs 


So, the plan is still the same, but is is now about solving the problem of only getting the 100 or 50 latest transactions instead of getting all of them (incrementally).

chedieck:
This is actually much simpler than I initially thought. A while loop for getting the transactions that uses `nbFetch: 50` and  `nbSkip` equal to the number of transactions already fetched is enough to sync all transactions incrementally.

Klakurka:
So we'll need to make sure this is possible in chronik too?

chedieck:
> So we'll need to make sure this is possible in chronik too?

I think it is, according to [this](https://e.cash/blog/how-chronik-indexer-improves-cashtab-wallet):

![image](https://user-images.githubusercontent.com/21281174/187978105-ea238ec4-2587-4994-a519-40e8b0d113ab.png)

PR #229, opened 2022-08-31
[Issue  #222] feat: add transaction blockHeight field

chedieck:
Description:
Related to  #222, adds the `blockHeight` field to the `Transaction` table.

Test Plan:
`make reset-dev` then, after logging in in the front end to sync the transactions, run `curl http://localhost:3000/address/transactions/ecash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7quss2vxek| jq` and check if the `blockHeight` is also included.

### DIFF
--- a/prisma/migrations/20220831141824_transaction_block_height/migration.sql
+++ b/prisma/migrations/20220831141824_transaction_block_height/migration.sql
@@ -1,0 +1,8 @@
+/*
+  Warnings:
+
+  - Added the required column `blockHeight` to the `Transaction` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- AlterTable
+ALTER TABLE `Transaction` ADD COLUMN `blockHeight` INTEGER NOT NULL;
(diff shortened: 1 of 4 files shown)


Klakurka:
Is there something we want to do that requires this?

prisma/migrations/20220831141824_transaction_block_height/migration.sql
```diff
@@ -0,0 +1,8 @@
+/*
+  Warnings:
+
+  - Added the required column `blockHeight` to the `Transaction` table without a default value. This is not possible if the table is not empty.
```
Gattaca1, line 4: I'm not sure how much this matters as the branch seems to work fine to me, but it looks like it is possible to add a column with a default value even if the table has data

https://stackoverflow.com/questions/92082/add-a-column-with-a-default-value-to-an-existing-table-in-sql-server

Gattaca1 (approved):
LGTM

PR #233, opened 2022-09-06
[Issue #222] feat: sync transactions in chunks

chedieck:
Description:
Part of #222. Makes so that all transactions of some address are fetched and added to our DB in chunks.

Test plan:
- Create a button for some address with significantly many transactions, e.g: [this address](https://ecashexplorer.com/address/ecash:qrv5hwntl5h4mypkg5kek6cj5f2d764txysuu0wffd) has 399 transactions,
- Log in and out to make it sync the transactions, and wait for it.
- Check the database to see if the numbers match:
  - Enter the docker container database with `yd db`,
  -   Get all transactions for newly created address:
```
MariaDB [paybutton]> SELECT * FROM Transaction where addressId=3;
[...]
399 rows in set (0.001 sec)
```

### DIFF
--- a/services/transactionsService.ts
+++ b/services/transactionsService.ts
@@ -74,6 +74,20 @@ export async function upsertTransaction (transaction: BCHTransaction.AsObject, a
   })
 }
 
+export async function getAllTransactions (addressString: string): Promise<void> {
+  let newTransactionsCount = -1
+  let totalFetched = 0
+  while (newTransactionsCount !== 0) {
+    const nextTransactions = (await grpcService.getAddress({
+      address: addressString,
+      nbFetch: 50,
+      nbSkip: totalFetched
+    })).confirmedTransactionsList
+    newTransactionsCount = nextTransactions.length
+    totalFetched += newTransactionsCount
+  }
+}
+
 export async function syncTransactions (addressString: string): Promise<void> {
   const address = parseAddress(addressString)
   if (address === '' || address === undefined) {
--- a/services/transactionsService.ts
+++ b/services/transactionsService.ts
@@ -6,7 +6,7 @@ import { parseAddress } from 'utils/validators'
 import { satoshisToUnit } from 'utils/index'
 import { fetchAddressBySubstring } from 'services/addressesService'
 import _ from 'lodash'
-import { RESPONSE_MESSAGES } from 'constants/index'
+import { RESPONSE_MESSAGES, FETCH_N, FETCH_DELAY } from 'constants/index'
 import xecaddr from 'xecaddrjs'
 
 const { ADDRESS_NOT_PROVIDED_400 } = RESPONSE_MESSAGES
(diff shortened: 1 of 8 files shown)


Klakurka (changes requested):
I added the current IFP address but it didn't seem to sync on the 1st attempt. I re-logged out and back in and it started going but it was quite a lot of processing even on my PC. Obviously it froze the UI until it was done, which took a few minutes.

Possible to have it "take a break" in between chunks so that stuff still loads for ppl while it's working in the bg?

Klakurka:
Hmm... not getting any blockchain data on the main dev account.

Klakurka (changes requested):
Not getting any blockchain data no matter what I try.

Switching back to master works.

chedieck:
Fixed, now it doesn't try to commit all transactions at once, it puts them in the database while fetching, 100 at a time.

The sleep time is of one second. Both this values, the amount of transactions to fetch at a time & the sleep time between each fetching would be better implemented as custom variables. I can address this in a next PR, but I am not sure if this should go as env variables or in another manner.

Another idea is to have someway of knowing that the syncing of some address is not done. E.g; when syncing the transactions for some address with 20k txs, it will take some time to finish it. In the meanwhile, we could have a "Syncing in progress" message or something alike. Knowing if the syncing concluded successfully or if it preemptively ended would also be useful.

Klakurka (approved):
Works way better now. I added a new task for doing paging on the transaction list as it's a lot of data to grab all at once if you've added a big address like the IFP one.

services/transactionsService.ts
```diff
@@ -73,13 +72,41 @@ export async function upsertTransaction (transaction: BCHTransaction.AsObject, a
...
+}
+
+export async function fetchAllTransactions (addressString: string): Promise<boolean> {
+  const address = await fetchAddressBySubstring(addressString)
+  let newTransactionsCount = -1
+  let seenTransactionsCount = 0
+  while (newTransactionsCount !== 0) {
+    const nextTransactions = (await grpcService.getAddress({
+      address: address.address,
+      nbFetch: 100,
```
ScottMcDermid, line 95: Magic number here. Maybe add a page-fetch-size constant.

services/transactionsService.ts
```diff
@@ -73,13 +72,41 @@ export async function upsertTransaction (transaction: BCHTransaction.AsObject, a
...
+      nbSkip: seenTransactionsCount
+    })).confirmedTransactionsList
+    newTransactionsCount = nextTransactions.length
+    seenTransactionsCount += newTransactionsCount
+    await upsertManyTransactions(nextTransactions, address)
+    await new Promise(resolve => setTimeout(resolve, 1000))
```
ScottMcDermid, line 101: Magic number here. Add a fetch-delay/transaction-delay constant.

ScottMcDermid (changes requested):
Just a few nits. The rest looks good.

Klakurka (dismissed):
Double checked some things and it looks good -- just want to get the syncing process running initially when a new address is added to the account.

services/transactionsService.ts
chedieck: Addressed

Issue #154, opened 2022-07-06
v1 Wallets


Klakurka:
- [x] Have all addresses be attached to a wallet ID. Wallets can be keyed on [wallet ID] + [user ID], so everyone will have a wallet with ID 0 to start.
- [x] When adding a new paybutton, allow them to select the desired wallet to attach it to in a dropdown. If they only have 1 wallet, grey out the dropdown so it just shows as being attached to the default wallet.
- [x] List all wallets on the Wallets page.
- [x] Allow creation of a new wallet, where all they do is choose the name and network.
- [x] Wallets will have an (wallet) ID, user ID, name, network, createdAt, updatedAt (new DB table).
- [x] When a user account is created, they're assigned a wallet call "Default Wallet" with an ID (index) of 0.
- [ ] Allow existing wallets to have their names changed.

chedieck:
Should `Address` <-> `Wallet` be one-to-many? As in: it is possible for different users have different wallets pointing to the same address; but each wallet will only point to one address.

Klakurka:
- many to many for the addresses <-> wallets feels right to me -- give people more freedom to do what they want.
- also, allowing them to have more than 1 currency in a wallet also seems totally fine to me. So for example for our company, we'd have a "Coin Dance" wallet which has both XEC and BCH (addresses) in it.

PR #238, opened 2022-09-08
[Issue #154] feat:wallet ui

johnkuney:
Adding some UI for the wallets page. Using mock data for now. Basically have 
-wallet name 
- network
- Current balance (wasnt sure if we wanted this, or more like on the dashboard where its total received?)
- payments
- Button list, and each one links to the specific button page
- edit button  
- create new wallet button


Klakurka (changes requested):
We're going to break a crypto convention here intentionally and allow them to have any number of currencies within a wallet (in the same way you can in a physical fiat wallet).

So instead of showing the 'eCash Network' text, let's show the currency icons for whatever networks they have payments on in the wallet on the buttons. So they won't need to "select" a currency when making a new wallet. The button's addresses contained within the wallet dictate what networks we show there. If they're on both XEC and BCH, just put both icons next to each other I guess unless you have a better idea.

Klakurka:
Let's also add an actual 'Default Wallet' marker that can set in the wallet edit dialog.

johnkuney:
For sure, added network icons, and separated out the balances

pages/wallets/index.tsx
```diff
@@ -34,14 +36,53 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
...
   )
 }
+
+class ProtectedPage extends React.Component<WalletsProps, WalletsState> {
+  constructor (props: WalletsProps) {
+    super(props)
+    this.state = {
+      wallets: [
+        {
+          name: 'Default Wallet',
```
Klakurka, line 61: We should have an actual field here for 'default_wallet' or whatever which is a true/false field.

Can we also have some sort of indicator somewhere that tells people that this is the default wallet (even if they change the wallet name)?

Klakurka (changes requested):
Nit but I think the vertical bar to the left of the icons should be the same font colour as the text.

Everything else looks good.

Klakurka:
I see that the vertical bar shows up in a few places. Can we go for something really subtle? The light theme is fine I think but the dark theme is a blue colour which I think should be changed to a dark grey or something so that it's not what the eyes are drawn to.

johnkuney:
Cool, added. Added edit wallet form too

Klakurka (approved):
Looks good.

I overlooked something though that we may need to sort out at some point...

We might need to do some brainstorming about how we want to approach this but basically we need to have a default for each currency I think.

Let's cross that bridge later as we run into actual issues.

PR #250, opened 2022-09-30
[Issue #154] feat: wallets scheme, migrations & seed

chedieck:
Description:
Create the `Wallet` model, migrations and the wallet objects to seed.

Test Plan:
Run `make reset-dev`
Enter the database with `yarn docker db`
See if the wallet table exists and is not empty: `SELECT * FROM Wallet;`


Klakurka:
I'm not sure how to test this one Windows.

ScottMcDermid (approved):
New table shows up. Code looks good!

PR #255, opened 2022-10-04
[Issue #154] feat: add wallet services

chedieck:
Description:
Add a `walletService.ts` file and tests for it.

Test Plan:
Should be covered by `tests/unittests/walletsService.test.ts`

Depends on:

- [x] #254

PR #251, opened 2022-09-30
Refactor/singular services name

chedieck:
Description:
Rename services so that they refer to the model in the singular. More specifically, renames:
- `addressesService` -> `addressService`
- `networksService` -> `networkService`
- `paybuttonsService` -> `paybuttonService`
- `transactionsService` -> `transactionService`

Test Plan: Nothing should change.

PR #256, opened 2022-10-04
Refactor/singular service tests name

chedieck:
Description:
Continuation of #251, in which I forgot to change the name of the test files.

Test plan:
Nothing should change.

Depends on:
- [ ] #255


constants/index.ts
```diff
@@ -9,9 +9,13 @@ export const RESPONSE_MESSAGES = {
   NAME_NOT_PROVIDED_400: { statusCode: 400, message: "'name' not provided." },
   NAME_ALREADY_EXISTS_400: { statusCode: 400, message: 'Button name already exists for this user.' },
   ADDRESSES_NOT_PROVIDED_400: { statusCode: 400, message: "'addresses' not provided." },
+  BUTTON_IDS_NOT_PROVIDED_400: { statusCode: 400, message: "'paybuttonIdList' not provided." },
   TRANSACTION_ID_NOT_PROVIDED_400: { statusCode: 400, message: "'transactionId' not provided." },
   INVALID_NETWORK_SLUG_400: { statusCode: 400, message: 'Invalid network slug.' },
   INVALID_BUTTON_DATA_400: { statusCode: 400, message: "'buttonData' is not valid JSON." },
+  PAYBUTTON_ALREADY_BELONGS_TO_WALLET_400: { statusCode: 400, message: 'Some of the paybuttons already belong to another wallet.' },
```
Klakurka, line 16: `One or more buttons already belong to another wallet.`

constants/index.ts
```diff
@@ -9,9 +9,13 @@ export const RESPONSE_MESSAGES = {
...
+  WALLET_CREATION_FAILED_400: { statusCode: 400, message: 'Wallet creation failed.' },
+  ADDRESS_ALREADY_BELONGS_TO_WALLET_400: { statusCode: 400, message: "Some of the paybuttons' addresses already belong to another wallet." },
```
Klakurka, line 18: `One or more button addresses already belong to another wallet.`

Klakurka (changes requested):

constants/index.ts
chedieck: This is refering to #255 changes; addressed.

Klakurka (approved):

Klakurka (changes requested):
Since we'll probably be supporting alt buttons in the future, can we have all of these constants use just the word 'button' instead of 'paybutton'?

Klakurka (approved):

PR #259, opened 2022-10-06
Refactor/ rename error message constant

chedieck:
Description:
Renames variable
`NAME_ALREADY_EXISTS_400` -> `PAYBUTTON_NAME_ALREADY_EXISTS_400`: because there will also be a `WALLET_NAME_ALREADY_EXISTS_400` in a future PR, so this makes it all more clear.

Test plan:
Nothing should change.


ScottMcDermid (approved):

chedieck:
> Since we'll probably be supporting alt buttons in the future, can we have all of these constants use just the word 'button' instead of 'paybutton'?

Yes, I was also feeling annoyed by this, so I'll address this in a next PR.

Klakurka (approved):

PR #266, opened 2022-10-10
[Issue #154] feat: wallet front end services

chedieck:
Description:
Creates new services that will be used by the wallet front-end in a next PR.

Test plan:
Should be covered by the unittests & integration tests changes.

### DIFF
--- a/services/walletService.ts
+++ b/services/walletService.ts
@@ -3,7 +3,7 @@ import * as paybuttonService from 'services/paybuttonService'
 import * as addressService from 'services/addressService'
 import { Prisma, Wallet } from '@prisma/client'
 import prisma from 'prisma/clientInstance'
-import { RESPONSE_MESSAGES } from 'constants/index'
+import { RESPONSE_MESSAGES, XEC_NETWORK_ID, BCH_NETWORK_ID } from 'constants/index'
 
 export interface CreateWalletInput {
   userId: string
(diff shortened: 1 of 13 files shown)


services/walletService.ts
```diff
@@ -74,6 +76,31 @@ export async function fetchWalletById (walletId: number | string): Promise<Walle
...
+
+export async function getWalletBalance (wallet: WalletWithAddressesAndPaybuttons): Promise<WalletPaymentInfo> {
+  const ret: WalletPaymentInfo = {
+    XECBalance: new Prisma.Decimal(0),
+    BCHBalance: new Prisma.Decimal(0),
+    paymentCount: 0
+  }
+  for (const addr of wallet.addresses) {
+    const addrBalance = await addressService.getAddressPaymentInfo(addr.address)
+    if (addr.networkId === 1) {
```
ScottMcDermid, line 93: Server shouldn't have to rely on the network data for eCash having a specific ID.

ScottMcDermid (changes requested):
Tests work well. Just one code comment.

services/walletService.ts
chedieck: What do you think if I set `ECASH_NETWORK_ID = 1` on the `constants.ts` file?
I know is not ideal, but the other option would be accessing the database to check which network is which, example:
`if (addr.networkId == networkService.fetchNetworkIdBySlug("ecash")) { ...`

...which in my view is not great because it will always access the database to get some information that never changes.

Btw, the `id` is fixed because the networks are created in the `prisma/seeds/networks.ts` file, always in the same order.

I will later update this file to include each network ID, just for better clarity (in another PR).

chedieck: Another thing: this check based on the hardcoded networkId is happening not only here but in other places too, so after we agree on a solution to this I think the best approach is to let this slide for these 2 open PRs we have now, and then I will open a PR addressing it at all places at once

ScottMcDermid: Alright thanks for the explanation. In that case, let's go with the new constant in `constants.ts`

Gattaca1 (approved):

Klakurka:
@chedieck has conflicts.

PR #263, opened 2022-10-07
[Issue #154] feat: wallet api endpoints

chedieck:
Description:
Creates 3 endpoints for wallets: creating a wallet, fetching a wallet by `id` and fetching all user wallets (just like the paybutton endpoints). Also adds  the tests for these endpoints.

 

Test plan:
The integration tests should be covering the new funcitonalities; but one can test the endpoints manually by using some `curl`-like command, examples:
```
curl -X GET http://localhost:3000/api/wallet/3
curl -X GET 'http://localhost:3000/api/wallets?userId=dev-uid'
curl -X POST localhost:3000/api/wallet -H "Content-Type: application/json"  -d '{"userId": "dev-uid", "name":"uhul", "paybuttonIdList":[1,2] }'
```

Depends on:
- [x] #262
- [x] #255

PR #267, opened 2022-10-10
[Issue #154]: Wallet front end list

chedieck:
**Description:**
http://localhost:3000/wallets now shows the real wallets in the database.

**Test plan:**
Should be covered by the tests.
Nonetheless, the reviewer can create paybuttons and then wallets and see if the information is making sense. Check #263 `curl` POST example to see how to create a wallet.

**Remarks:**
- The function to edit the wallets, accessible when clicking in the pencil icon, has not been addressed in this PR and is not expected to work;
- The `default_wallet` is not implemented, so the `wallets` endpoint is returning all wallets as if this property was `false`. This also will be addressed in a next PR.

**Depends on:**
- [x] #266

### DIFF
--- a/components/Wallet/WalletCard.tsx
+++ b/components/Wallet/WalletCard.tsx
@@ -26,8 +26,8 @@ export default ({ wallet, paymentInfo }: IProps): FunctionComponent => {
           </div>
         </div>
         <div className={style.edit_button_ctn}>
-          {wallet.userProfile?.isDefaultForNetworkId === 1 && <div className={style.default_wallet}>Default XEC Wallet</div>}
-          {wallet.userProfile?.isDefaultForNetworkId === 2 && <div className={style.default_wallet}>Default BCH Wallet</div>}
+          {wallet.userProfile?.isDefaultForNetworkId === XEC_NETWORK_ID && <div className={style.default_wallet}>Default XEC Wallet</div>}
+          {wallet.userProfile?.isDefaultForNetworkId === BCH_NETWORK_ID && <div className={style.default_wallet}>Default BCH Wallet</div>}
           <EditWalletForm wallet={wallet} />
         </div>
       </div>
--- a/components/Wallet/WalletCard.tsx
+++ b/components/Wallet/WalletCard.tsx
@@ -6,6 +6,7 @@ import XECIcon from 'assets/xec-logo.png'
 import BCHIcon from 'assets/bch-logo.png'
 import EditWalletForm from './EditWalletForm'
 import { WalletWithAddressesAndPaybuttons, WalletPaymentInfo } from 'services/walletService'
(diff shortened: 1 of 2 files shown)


components/Wallet/WalletCard.tsx
```diff
@@ -5,50 +5,54 @@ import Image from 'next/image'
...
   return (
     <div className={style.wallet_card}>
       <div className={style.wallet_card_header_ctn}>
         <div className={style.wallet_card_header}>
-          <h4>{walletInfo.name}</h4>
+          <h4>{wallet.name}</h4>
           <div className={style.walletcard_icons}>
-            {walletInfo.xec_balance > 0 && <div><Image src={XECIcon} alt='XEC' /></div>}
-            {walletInfo.bch_balance > 0 && <div><Image src={BCHIcon} alt='BCH' /></div>}
+            {networks.includes(1) && <div><Image src={XECIcon} alt='XEC' /></div>}
```
ScottMcDermid, line 23: I'd feel better if we had a more reliable check than `networkId === 1` to indicate the currency of an address.
The client shouldn't have to know network ID's

Klakurka:
We'll want to get the default wallet situation sorted out later.

components/Wallet/WalletCard.tsx
ScottMcDermid: This isn't resolved yet. Just wanted to re-open this in case you missed it.

Klakurka:
Conflicts.

Klakurka:
Conflicts

components/Wallet/WalletCard.tsx
chedieck: Indeed I had missed it, thanks. Solved.

pages/wallets/index.tsx
```diff
@@ -81,7 +81,21 @@ class ProtectedPage extends React.Component<WalletsProps, WalletsState> {
...
         <div>
-        {this.state.walletsWithPaymentInfo.sort((a, b) => Number(b.wallet.default_wallet) - Number(a.wallet.default_wallet)).map(walletWithPaymentInfo => {
+        {this.state.walletsWithPaymentInfo.sort((a, b) => {
+          /* Sorts in the following order, from first to last, if they exist:
+           * Default XEC Wallet, Default BCH Wallet, other wallets.
+           */
+          const aNetworkId = Number(a.wallet.userProfile?.isDefaultForNetworkId)
+          const bNetworkId = Number(b.wallet.userProfile?.isDefaultForNetworkId)
+          switch (aNetworkId | bNetworkId) {
+            case 1 | 2:
```
ScottMcDermid, line 91: I realize this is pulled in from master, but we may as well update these `case` statements to use the network constants

ScottMcDermid (changes requested):

Klakurka (approved):

ScottMcDermid (approved):

PR #270, opened 2022-10-13
[Issue #154] feat: default wallet for network

chedieck:
Description:
Replaces the mocked data on the [wallets page](http://localhost:3000/wallets) with the real data.

Test plan:
Editing the wallets from the front end is still not implemented; but one can create paybuttons and then wallets using e.g.
```
curl -X POST localhost:3000/api/wallet -H "Content-Type: application/json"  -d '{"userId": "dev-uid", "name":"eg", "paybuttonIdList":[12] }'
```
 ... and then see if the payment info is correct.

Depends on:
- [x] #269


Klakurka (approved):

johnkuney (approved):
Was able to create a new wallet using the command, and it updated and had correct info. Looks good to me!

Issue #272, opened 2022-10-14
Back button should point to the page you actually came from


Klakurka:
1. Go to the Wallets page.
2. Click on a button link.
3. Click Back.
4. Notice how instead of taking you back to Wallets it takes you to Buttons.

Issue #271, opened 2022-10-14
Login box is off to the left


Klakurka:
Should be centered.

PR #276, opened 2022-10-17
[Issue #271, #272] bug: center login box, change button back behavior

johnkuney:
A couple edits to center the login box and make the back button on the /button/id page go back to the page you came from instead of always back to the /buttons page


ScottMcDermid (approved):
Looks good!

Klakurka (approved):

Issue #275, opened 2022-10-14
Have a wallet be the default for multiple networks.


chedieck:
I would like for there to only be only 1 'Default Wallet' marker if a wallet is the default for all (2) currencies but I'm fine with this as is for now.

_Originally posted by @Klakurka in https://github.com/PayButton/paybutton-server/issues/270#issuecomment-1278777368_

PR #280, opened 2022-10-24
[Issue #154, #275] feat: multiple default wallets

chedieck:
**Description:**
A wallet now can be the default for both XEC and BCH.

**Test Plan:**
The edit wallet front end is not yet implemented, so for now one can change the `prisma/seeds/walletUserConnectors.ts` file such as in the print below:
image

... and then after running `yarn docker mr` to reset the database, the following should be expected
image

### DIFF
--- a/components/Layout/layout.module.css
+++ b/components/Layout/layout.module.css
@@ -16,6 +16,7 @@
   max-height: 100vh;
   overflow-y: scroll;
   width: calc(100% - 220px);
+  margin: 0 auto;
 }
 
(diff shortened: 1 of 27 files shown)


ScottMcDermid (approved):
Code looks good. Default sorting works for me, although I had to update the test data so that both defaults were on a single wallet through `yarn docker db`

johnkuney (approved):
Nice. Test plan worked for me, except the yarn docker mr. That container kept restarting for some reason. Just reset everything though, and that worked and I saw the two defaults

Issue #246, opened 2022-09-28
Add 'All' button if there's more than 1 year of data


Klakurka:
For now, just keep it to monthly values but perhaps we can have it automatically switch to yearly values if there are more than X months of data.

PR #286, opened 2022-10-26
[Issue #246] feat: 'All' option for dashboard charts

johnkuney:
Adding an all option to the dashboard charts. Also fixed a few bugs that were causing sub $0.01 transactions to not register, and causing some date misalignments in the charts 

Test plan: Run the app and test out the all button, and see if the transactions line up on the chart


Klakurka (changes requested):
Definitely better.

Only issue I saw was this when there were no data points >0 in the range (showing negative values in the Y-axis:

image

johnkuney:
Good call, should be good now

ScottMcDermid (approved):
Looks good!

Klakurka (approved):

PR #254, opened 2022-10-04
[Issue #154] chore: fix paybutton service typing and tests

chedieck:
Description:
Some changes related to typing the `Paybutton` objects returned by functions in `services/paybuttonService.ts`. The format of the returned object was fixed in the `tests/mockedObjects.ts` file.

Test Plan:
Nothing should change.

PR #287, opened 2022-10-27
[Issue #254] feat: wallet edit back end defaults

chedieck:
WIP

Implements changing default wallet from the frontend.

### DIFF
--- a/components/Wallet/EditWalletForm.tsx
+++ b/components/Wallet/EditWalletForm.tsx
@@ -11,9 +11,10 @@ import { WalletWithAddressesAndPaybuttons } from 'services/walletService'
 interface IProps {
   wallet: WalletWithAddressesAndPaybuttons
   userPaybuttons: Paybutton[]
+  refreshWalletList: Function
 }
 
-export default function EditWalletForm ({ wallet, userPaybuttons }: IProps): ReactElement {
+export default function EditWalletForm ({ wallet, userPaybuttons, refreshWalletList }: IProps): ReactElement {
   const { register, handleSubmit, reset } = useForm<WalletPATCHParameters>()
   const [modal, setModal] = useState(false)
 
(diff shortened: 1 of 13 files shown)


johnkuney (changes requested):
Nice, worked for me.
Only thing I noticed is I was able to uncheck the current default wallet, making it so there is no default wallet...
Maybe an easy fix would be just hide that option in the dialog box if it is currently the default wallet?
Along those lines would also be better UI I think if we hide the XEC default option if its a BCH wallet and vice versa.

chedieck:
> Only thing I noticed is I was able to uncheck the current default wallet, making it so there is no default wallet...
I thought this was the desired behavior, we shouldn't allow no default wallets?
> Along those lines would also be better UI I think if we hide the XEC default option if its a BCH wallet and vice versa.
Indeed, I'll work on that.

johnkuney (commented):
Cool, yeah thats better. 
Hmm yeah im not sure, I thought the idea was there always had to be a default wallet

Klakurka:
Ya we always want 1 default wallet for each currency because this is the wallet that will be set initially in the "create new button" dialog. They can change it then or even after they create the button but it's better to that than to use some random wallet right?

Klakurka (changes requested):
- Don't allow removing the default wallet, only replacing it with another one
- consolidate labels like John said

chedieck:
Addressed, and now the checkbox is disabled when trying to
- unset a wallet as default;
- setting a wallet with no XEC(BCH) addresses as default for XEC(BCH)

Klakurka (changes requested):
I'm not seeing the merged labels for some reason but the checkboxes work.

Klakurka (approved):
Nevermind I misread.

We can do it in either branch later as it's not important.

In was talking about showing "Default Wallets" instead of 2 separate labels for each currency if the wallet is the default for all currencies.

johnkuney (approved):

PR #290, opened 2022-10-31
[#154 Issue] feat: wallet edit back end name

chedieck:
Description:
Editing a wallet's name also works now.

Test plan:
Go to http://localhost:3000/wallets and try to change the wallets' names. 

Depends on:
- [x] #287

### DIFF
--- a/components/Wallet/EditWalletForm.tsx
+++ b/components/Wallet/EditWalletForm.tsx
@@ -8,6 +8,8 @@ import s from '../Wallet/wallet.module.css'
 import EditIcon from 'assets/edit-icon.png'
 import { WalletWithAddressesAndPaybuttons } from 'services/walletService'
 import { XEC_NETWORK_ID, BCH_NETWORK_ID } from 'constants/index'
+import axios from 'axios'
+import { appInfo } from 'config/appInfo'
 
 interface IProps {
   wallet: WalletWithAddressesAndPaybuttons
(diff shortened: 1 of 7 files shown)


johnkuney (changes requested):
Sweet, worked well for me. I noticed though if you try using an existing name for instance the api responds with an error, but nothing happens on screen. We should show any error messages on screen

chedieck:
> Sweet, worked well for me. I noticed though if you try using an existing name for instance the api responds with an error, but nothing happens on screen. We should show any error messages on screen

Indeed, we don't support showing errors yet. We definitively should do that, but I think in a next PR, since this problem refers to many other pages besides `/wallets`

johnkuney (dismissed):
Okay cool fair enough.  I think I did actually add some error message handling on the create button form...but probably not done in the best way/shareable with other forms. But yeah fine to tackle in another PR

components/Wallet/EditWalletForm.tsx
```diff
@@ -27,7 +28,13 @@ export default function EditWalletForm ({ wallet, userPaybuttons, refreshWalletL
     if (params.name === '' || params.name === undefined) {
       params.name = wallet.name
     }
-    console.log('edit wallet params', params)
+    void await fetch(`/api/wallet/${wallet.id}`, {
```
ScottMcDermid, line 31: We should be using the `websiteDomain` environment variable for api requests.

components/Wallet/EditWalletForm.tsx
```diff
@@ -27,7 +28,13 @@ export default function EditWalletForm ({ wallet, userPaybuttons, refreshWalletL
...
+      method: 'PATCH',
```
ScottMcDermid, line 32: Optionally: `axios` would be more succinct. We're using it in `pages/api/resolvers.index.ts`, at least.
Something like:

```js
void await axios.patch(`${websiteDomain}/api/wallet/${wallet.id}`, params );
```

Should accomplish the same thing I think.

ScottMcDermid (changes requested):

components/Wallet/EditWalletForm.tsx
chedieck: Addressed, also made sure this wouldn't be `undefined`.

components/Wallet/EditWalletForm.tsx
chedieck: Addressed (check last commit)

ScottMcDermid (approved):
Looks good!

Klakurka (approved):

Issue #223, opened 2022-08-29
Process for keeping currency values up to date (and retained for each day)


Klakurka:
Related to #63 and #105.

Basically the only requirements for this are:

- We have a price for every day that we have a transaction for (for each currency); not necessarily a price for every day ever.
- We have a recent (maybe updated every 5-15 minutes) price for each currency. "What's the price now?"

chedieck:
My idea to implement this, so far:
-  Create `Price` table with columns: id, quote, date, value & chainId. (quote will only be USD for now, later we can add e.g. BTC)
-  Script to populate the `Price` table using the `Transaction` timestamps.

The `date` field could be simply a YYYY-MM-DD string, since we want only one `Price` for each day.

Another option is to have one `Price` for each timestamp (and thus have a `timestamp` column instead of `date`), which will increase the time needed to fetch & save all prices, but give us more accurate information.

Klakurka:
Having a price for each time stamp actually does sound better but if we did that then in reality we'd have a mix of "accurate" prices and "daily" prices on use because most txs we'd probably not actually see in the system until they log in later.

Klakurka:
It's simpler to just go with a daily price though.

chedieck:
> Having a price for each time stamp actually does sound better but if we did that then in reality we'd have a mix of "accurate" prices and "daily" prices on use because most txs we'd probably not actually see in the system until they log in later.

I am not sure if I understand; but I think we could have `Price`s with a `timestamp` field for the timestamps relevant to the `Transaction`s; and on top of that `Price`s with `timestamp == 0` representing the current price (updated every N minutes).

chedieck:
Solved with #300

PR #292, opened 2022-11-02
[Issue #223] feat: create prices tables

chedieck:
Description:
Create one table for prices, another one for their quotes (BTC and USD for now) and another one connecting prices to transactions.

Test plan:
Nothing yet, only the models are being created.

### DIFF
--- a/prisma/migrations/20220715134600_init/migration.sql
+++ b/prisma/migrations/20220715134600_init/migration.sql
@@ -100,6 +100,43 @@ CREATE TABLE `WalletsOnUserProfile` (
     PRIMARY KEY (`walletId`, `userProfileId`)
 ) DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
 
+-- CreateTable
+CREATE TABLE `Price` (
+    `id` INTEGER NOT NULL AUTO_INCREMENT,
+    `value` DECIMAL(64, 16) NOT NULL,
+    `createdAt` DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
+    `updatedAt` DATETIME(3) NOT NULL,
+    `timestamp` INTEGER NOT NULL,
+    `networkId` INTEGER NOT NULL,
+    `quoteId` INTEGER NOT NULL,
+
+    UNIQUE INDEX `Price_timestamp_quoteId_networkId_key`(`timestamp`, `quoteId`, `networkId`),
+    PRIMARY KEY (`id`)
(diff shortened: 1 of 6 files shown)


prisma/seeds/quotes.ts
```diff
@@ -0,0 +1,18 @@
...
+    id: 1,
+    slug: 'usd',
+    ticker: 'USD',
+    title: 'US Dollars',
+    createdAt: new Date(),
+    updatedAt: new Date()
+  },
+  {
+    id: 2,
+    slug: 'bitcoin',
```
Klakurka, line 12: I don't think we need BTC-anything in the repo, right?

prisma/migrations/20220715134600_init/migration.sql
```diff
@@ -100,6 +100,43 @@ CREATE TABLE `WalletsOnUserProfile` (
     PRIMARY KEY (`walletId`, `userProfileId`)
 ) DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
 
+-- CreateTable
+CREATE TABLE `Price` (
+    `id` INTEGER NOT NULL AUTO_INCREMENT,
+    `value` DECIMAL(64, 16) NOT NULL,
```
Klakurka, line 106: Probably could get away with 8 decimals as I doubt we're going to be touching BTC at all.

prisma/seeds/prices.ts
```diff
@@ -0,0 +1,35 @@
...
+export const prices = [
+  { // XECUSD
+    value: new Prisma.Decimal('0.00004095051166'),
+    createdAt: new Date(),
+    updatedAt: new Date(),
+    timestamp: 166459320,
+    networkId: 1,
+    quoteId: 1
+  },
+  { // XECBTC
```
Klakurka, line 11: Half of the team is Canadian and we do sort of need CAD conversions (which the API does have rn) so I'd rather we just do that instead of BTC here and below.

prisma/seeds/prices.ts
```diff
@@ -0,0 +1,35 @@
...
+  },
+  { // BCHUSD
+    value: new Prisma.Decimal('117.90816992278000'),
+    createdAt: new Date(),
+    updatedAt: new Date(),
+    timestamp: 166459320,
+    networkId: 2,
+    quoteId: 1
+  },
+  { // BCHBTC
```
Klakurka, line 27: Same.

Klakurka (changes requested):

prisma/seeds/prices.ts
chedieck: Addressed this and the other remarks (check last commit)

Klakurka (approved):

PR #294, opened 2022-11-03
[Issue #223] feat: price services

chedieck:
Description:
Creates the price services & syncs prices together with transactions.

Test plan:
1. Delete the transactions from the database:
```
$ yarn docker db
...
> DELETE FROM Transaction;
```
2. Log out and back in to resync them with the prices
3. Check if the tables have been populated
```
...
> SELECT * FROM Price;
> SELECT * FROM PricesOnTransactions;
``` 

Depends on:
- [x] #293

### DIFF
--- a/services/priceService.ts
+++ b/services/priceService.ts
@@ -1,0 +1,144 @@
+import { fetchAllTransactions, TransactionWithAddressAndPrices } from 'services/transactionService'
+import axios from 'axios'
+import { appInfo } from 'config/appInfo'
+import { Prisma } from '@prisma/client'
+import prisma from 'prisma/clientInstance'
+import { RESPONSE_MESSAGES, XEC_NETWORK_ID, BCH_NETWORK_ID, USD_QUOTE_ID, CAD_QUOTE_ID } from 'constants/index'
+
+function dateStringFromTimestamp (timestamp: number): string {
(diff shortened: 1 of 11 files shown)


services/priceService.ts
```diff
@@ -0,0 +1,159 @@
...
+    create: {
+      transactionId: params.transactionId,
+      priceId: cadPrice.id
+    },
+    update: {}
+  })
+}
+
+export async function syncTransactionPrices (params: SyncTransactionPricesInput): Promise<void> {
+  if (appInfo.priceAPIURL === undefined) {
```
ScottMcDermid, line 128: I think environment variables are typically strings. This isn't triggering for me when I leave the PRICE_API_URL blank

ScottMcDermid (changes requested):
Just one comment. Otherwise tests look good

services/priceService.ts
chedieck: You probably left it as `PRICE_API_URL=`, which will set it to an empty string. If you don't mention the variable at all it will come as `undefined`.

ScottMcDermid: Ya that was it. All good then!

ScottMcDermid (approved):

Issue #299, opened 2022-11-06
Display prices/values consistently


Klakurka:
Right now we're inconsistent with displaying prices/values/amounts throughout the site.

IMO we should stick with something like:

Amount: crypto quantity
Value: fiat conversion at the time of the transaction

Then, add an on-hover tooltip on 'Value' which shows the current fiat conversion. This would allow somebody to check how much they paid in "today's money" for their favorite slice of pizza from 6 years ago.

Klakurka:
And of course the 'Value' column would use whatever fiat currency the user has selected in their settings.

PR #300, opened 2022-11-07
[Issue  #299] feat: save current price

chedieck:
Description:
Part of #299. Implements current prices and sync them upon login.

Test plan: 
1. Log out and back in to sync the prices,
2. Run:

```
$ yarn docker db
...
> SELECT * FROM Price WHERE timestamp=0;
```

The result should be 4 prices (XECUSD, BCHUSD, XECCAD, BCHCAD), as in the pic:

image

Depends on: 
- [x] #297


Klakurka:
Maybe let's make a new task to sync on some other cron process or something because this sounds like a lot of requests if we have a bunch of users.

Klakurka:
On the dashboard, is the "Lifetime Total Revenue" using this or the historical prices? We want it using the historical prices if possible.

Klakurka (changes requested):
Actually I get this error sometimes when logging back in:

```
paybutton-dev    | error - services/priceService.ts (149:29) @ eval
paybutton-dev    | TypeError: Cannot read properties of undefined (reading 'usd')
paybutton-dev    |   147 |       },
paybutton-dev    |   148 |       create: {
paybutton-dev    | > 149 |         value: params.values.usd,
paybutton-dev    |       |                             ^
paybutton-dev    |   150 |         timestamp: params.timestamp,
paybutton-dev    |   151 |         network: {
paybutton-dev    |   152 |           connect: {
```

Any idea what that is?

Also when I log out and log back in, no new pricing entries go in. Is that expected?

ScottMcDermid (changes requested):
Code looks good! The 0-timestamp prices populate as expected.
I get the same `Cannot read properties of undefined (reading 'usd')` error as david.

chedieck:
The error problem should be solved now.

> On the dashboard, is the "Lifetime Total Revenue" using this or the historical prices? We want it using the historical prices if possible.

Historical prices, this is not being used anywhere actually.

> Also when I log out and log back in, no new pricing entries go in. Is that expected?

I'm not sure if I understood, but yes, logging out and back in will only update the current prices, not create new ones.

Klakurka (commented):
I get this on first load. Is this expected?

```
paybutton-dev    | error - (api)/services/transactionService.ts (51:55) @ Module.getTransactionValue
paybutton-dev    | Error: txid1, ts1666078528 Missing price for transaction.
paybutton-dev    |   49 |     cad: new Prisma.Decimal(0)
paybutton-dev    |   50 |   }
paybutton-dev    | > 51 |   if (transaction.prices.length !== N_OF_QUOTES) throw new Error(`txid${transaction.id}, ts${transaction.timestamp} ${RESPONSE_MESSAGES.MISSING_PRICE_FOR_TRANSACTION_400.message}`)
paybutton-dev    |      |                                                       ^
paybutton-dev    |   52 |   for (const p of transaction.prices) {
paybutton-dev    |   53 |     if (p.price.quoteId === USD_QUOTE_ID) {
paybutton-dev    |   54 |       ret.usd = ret.usd.plus(p.price.value.times(transaction.amount))
paybutton-dev    | event - compiled client and server successfully in 893 ms (519 modules)
paybutton-dev    | grpc-bchrpc-node: failed network integrity check
```

chedieck:
> I get this on first load. Is this expected?

This may happen if you refresh the page while the prince sync didn't happen yet. Maybe in the future we can have the "syncing state" programmed in some way that we avoid this error message on this particular case, since it's already expected.

Klakurka (changes requested):
I'm seeing some weirdness in the `timestamp` column in the DB:

image

Notice the missing 0 on the first few, then just 0s for a bunch, then what I assume are correct time stamps for the rest.

Klakurka:
That was on the main `dev` account with no extra buttons.

chedieck:
> I'm seeing some weirdness in the timestamp column in the DB:
Indeed, the first prices (the ones with the missing zero) were seeded prices that I manually input this incorrect timestamp (fixed in the last commit).
The ones with 0 for `timestamp` are the current prices.

Klakurka (approved):

Klakurka:
Actually, any reason why we're not just using the current time stamp instead of 0 for those? @chedieck

ScottMcDermid (approved):

ScottMcDermid (changes requested):
I'm changing my approval status just to ask a few questions

1. Why use a special `timestamp` value for current prices, instead of querying for the latest price?
2. Why not use a dedicated `CurrentPrice` table, which can store the current price's timestamp?

I'm down to approve this, and create some followup tasks. Especially if it's mostly for the sake of efficiency. Just want to know your thoughts!

chedieck:
> Actually, any reason why we're not just using the current time stamp instead of 0 for those? @chedieck

The current timestamp is immediately changing, so if I save the current price with now's timestamp and 2 seconds later I want to fetch it on another place, to use in some computation: I wouldn't know how to fetch it, since neither it's timestamp or id is known.

Having the 0 timestamp makes it clear that that is the current price, and since the timestamp is unique, there can always be only one current price (as it should be); and we always update this same entry with the new current price (instead of creating many entries, one for each timestamp of the time we asked to get the current price).

chedieck:
>  Why use a special `timestamp` value for current prices, instead of querying for the latest price?

Because we only create prices for the transactions we have on the DB, if the latest transaction is old, the latest price will also be old and not reflect the current price.

>  Why not use a dedicated `CurrentPrice` table, which can store the current price's timestamp?

In my vision this would be an overkill (a whole new table for N objects, where N is the number of supported currencies, currently equal to `2`), but I am also not seeing why we would like to keep history of what we have already considered being the "current price": for each transaction we have the price at the moment it happened; and we also have the price of right now. Is there any reason why I would want to know what was the 'price right now' 15 minutes ago? If there was a transaction then, the price will be there. But if not, is this information useful for anything?

ScottMcDermid (approved):
> if the latest transaction is old, the latest price will also be old and not reflect the current price.

Right ya.. That would be an issue.

> Is there any reason why I would want to know what was the 'price right now' 15 minutes ago?

In the future, we'll want to occasionally synchronize the latest price. Either with some service, or when we visit a particular page. Knowing how stale the `currentPrice` value is, would be useful for such a service.

For now, I think this implementation works great. So I'm approving for now.

chedieck:
> In the future, we'll want to occasionally synchronize the latest price. Either with some service, or when we visit a particular page. Knowing how stale the `currentPrice` value is, would be useful for such a service.

We can use the info on the `updatedAt` / `createdAt` fields to know that

PR #303, opened 2022-11-08
[Issue #245] feat: delete paybutton

chedieck:
Description:
Solves #245. Users can delete paybuttons from the frontend, on the button detail page.

Test plan:
Go to http://localhost:3000/buttons, select any button and try to delete it by clicking on the pencil. Note that the "Edit" function is not implemented yet, only the "Delete".

Depends on:
- [x] #302 
- [x] #309

### DIFF
--- a/components/Paybutton/EditButtonForm.tsx
+++ b/components/Paybutton/EditButtonForm.tsx
@@ -1,5 +1,5 @@
 import React, { ReactElement, useState, useEffect } from 'react'
-import  { PaybuttonWithAddresses } from 'services/paybuttonService'
+import { PaybuttonWithAddresses } from 'services/paybuttonService'
 import { useForm } from 'react-hook-form'
 import { paybuttonPOSTParameters } from 'utils/validators'
 import Image from 'next/image'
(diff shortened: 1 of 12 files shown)


johnkuney (changes requested):
I'm seeing some display issues on the buttons page.
And actually there should already be some UI in place for deleting. If you you go to the button detail page and click the pencil icon, there is a delete button. If we can get this functionality there, that'd be ideal
Screen Shot 2022-11-08 at 11 27 16 AM
Screen Shot 2022-11-08 at 11 27 10 AM
Screen Shot 2022-11-08 at 11 26 51 AM

chedieck:
> I'm seeing some display issues on the buttons page. 

Indeed, I didn't care about that however; I just tried to make a pipeline that worked and leave the front end for you :shrug: (it would take a lot longer for me to brute-force the CSS into making it work)

> And actually there should already be some UI in place for deleting. If you you go to the button detail page and click the pencil icon, there is a delete button. If we can get this functionality there, that'd be ideal

Nice, I didn't know that. In this case, I will remove the ugly buttons from the button list and move this logic there.

chedieck:
Ready for review after dependency PRs are merged.

Klakurka (approved):

PR #305, opened 2022-11-08
chore: fix github action warnings

chedieck:
Description:
Update github actions, to stop getting this warning:

untitled(1)

Test plan:
The github action results, for this PR, should not contain any warning, as the ones for #303, for example, do.


johnkuney (commented):
I'm still seeing the same warning on this PR I believe 
Screen Shot 2022-11-08 at 11 10 11 AM

Klakurka (changes requested):
Seeing same as John.

chedieck:
> I'm still seeing the same warning on this PR I believe

Indeed, I think this didn't run when I created the draft PR. Now, this check comes from https://github.com/gregsdennis/dependencies-action; so we either fork and update that, we ignore it, or we even delete the check: every PR with dependencies I create I do so by stating it in the following way:

Depends on:
- [ ] #X


... which is not supported by the check, so it always passes even if the PR has dependencies.

johnkuney:
Oh my bad, didnt see its a new warning for the other check. hmm seems like we don't really need the dependencies check then? Or if we keep it should use it so it actually catches errors...I would say probably not worth maintaining a fork just to clear a warning

chedieck:
> Oh my bad, didnt see its a new warning for the other check. hmm seems like we don't really need the dependencies check then? Or if we keep it should use it so it actually catches errors...I would say probably not worth maintaining a fork just to clear a warning

I agree with all you said, and I think the way to go is to just remove the action.

chedieck:
Btw, here I think is another reason to axe it:

image

The red X on the pic above  makes it look like tests didn't pass and that the PR is broken; when the tests have actually passed, only a dependency PR has not yet been merged. I think having the "incomplete tasks", that github shows when there are items in a check list on the PR description that haven't been checked yet is more clear about "the tests have passed, but there still other things that this depends on that are incomplete". Example:

Incomplete:
image

Complete:
image

ScottMcDermid (approved):

PR #307, opened 2022-11-08
[Issue #306] feat: create default wallet for user

chedieck:
Description:
Upon registering, new users have now a wallet being created for them, set as the default for both XEC & BCH.

Test plan:
Log out, create a new user, confirm the email. A empty wallet should be there:
image


johnkuney (approved):
Looks good to me. Test plan worked

ScottMcDermid (approved):
Works for me!

PR #308, opened 2022-11-09
[Issue #155] feat: showing networks connection info

chedieck:
Description:
Solves #155. Shows network info in the front end.
image

Test plan:
Go to http://localhost:3000/networks and see if the info is there. Then mess up something of the `.env` file, run `make reset-dev` and see if it shows "Disconnected" for the messed up network(s).

### DIFF
--- a/components/Network/NetworkList.tsx
+++ b/components/Network/NetworkList.tsx
@@ -1,8 +1,8 @@
 import React, { FunctionComponent } from 'react'
-import { Network } from '@prisma/client'
+import { NetworkWithConnectionInfo } from 'services/networkService'
 import style from './network.module.css'
 
-interface IProps { networks: Network[] }
+interface IProps { networks: NetworkWithConnectionInfo[] }
 export default ({ networks }: IProps): FunctionComponent<IProps> => {
   return (
     <div className={style.network_ctn}>
@@ -10,8 +10,8 @@ export default ({ networks }: IProps): FunctionComponent<IProps> => {
         <div key={network.id} className={style.card_wrapper}>
             <div className={style.network_card_text}>
               <div className={style.cardTitle}>{network.title}</div>
-              <div>Last block: X time ago</div>
-                {network.title === 'eCash' ? 
+              <div>Last block: {network.minutesSinceLastBlock ?? '-'} time ago</div>
+                {network.connected === true ? 
                   <div className={style.cardStatus} style={{ color: '#04b504' }}>Connected</div> :
                   <div className={style.cardStatus}>Disconnected</div>
                 }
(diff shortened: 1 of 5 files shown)


components/Network/NetworkList.tsx
```diff
@@ -1,17 +1,17 @@
...
 export default ({ networks }: IProps): FunctionComponent<IProps> => {
   return (
     <div className={style.network_ctn}>
       {networks.map(network => (
         <div key={network.id} className={style.card_wrapper}>
             <div className={style.network_card_text}>
               <div className={style.cardTitle}>{network.title}</div>
-              <div>Last block: X time ago</div>
-                {network.title === 'eCash' ? 
+              <div>Last block: {network.minutesSinceLastBlock ?? '-'} time ago</div>
```
ScottMcDermid, line 13: Change to `minutes ago`.
In the future, we could pass the time of the last block, and let `moment()` do the rest.

ScottMcDermid (changes requested):
Tested, and works as expected. Just one code comment.

Klakurka (changes requested):
The label says "X time ago". IDK what that means. Should 'time' be 'minutes"? If so, would be nice if it said 'minute' instead of 'minutes' when it was 1 minute ago to be grammatically correct.

Scott addressed it earlier; missed that.

components/Network/NetworkList.tsx
chedieck: Addressed. I don't understand the second part of the comment tho

components/Network/NetworkList.tsx
```diff
@@ -1,23 +1,25 @@
...
             <div className={style.network_card_text}>
               <div className={style.cardTitle}>{network.title}</div>
-              <div>Last block: X time ago</div>
-                {network.title === 'eCash' ? 
-                  <div className={style.cardStatus} style={{ color: '#04b504' }}>Connected</div> :
-                  <div className={style.cardStatus}>Disconnected</div>
+                {network.connected
+                  ? <>
+                    <div className={style.cardStatus} style={{ color: '#04b504' }}>Connected</div>
+                    <div>Last block: {network.minutesSinceLastBlock ?? '-'} {network.minutesSinceLastBlock === 1 ? 'minute' : 'minutes'} ago</div>
```
Klakurka, line 16: Maybe we can make a new task to use a lib or custom function to display a nice bit of text here?

Something that covers:

 - 0 minutes: "just now"
 - >=60 minutes (switch to 1 hour / x hours)

Klakurka (approved):

components/Network/NetworkList.tsx
Klakurka: John and Scott have some experience with doing this in the past on other projects.

components/Network/NetworkList.tsx
ScottMcDermid: `moment.js` is one of our projects' dependencies. It has a cool feature that accepts a date, and returns a string like `X minutes ago`
https://momentjs.com/docs/#/displaying/fromnow/

I'll make a separate task for it

ScottMcDermid (approved):

Issue #252, opened 2022-09-30
Add new tx when a new one is found during client polling 


Klakurka:
So if somebody is logged in, they will just have to refresh the page / navigate to another page to see their tx show up.

Later this will be done via a socket connection (#88) but it may be a while before we will have this working. Perhaps we can still set up a socket connect to do this with bchd/xecd?... Whatever works.

Klakurka:
Should be working fine now.

PR #314, opened 2022-11-14
[Issue #252] feat: sync txs on txs views

chedieck:
Description:
Accessing the button detail view and the dashboard now sync transactions.

Test plan:

Log in, go to the dashboard or the button detail view. Both should work normally.

Then delete all transactions by running:
```
$ yarn docker db
> DELETE FROM Transaction;
```
And access one of the pages. The transactions should appear back again.

### DIFF
--- a/config/backendConfig.ts
+++ b/config/backendConfig.ts
@@ -4,7 +4,7 @@ import { appInfo } from './appInfo'
 import { TypeInput } from 'supertokens-node/types'
 import * as addressService from 'services/addressService'
 import * as walletService from 'services/walletService'
-import { syncTransactions } from 'services/transactionService'
+import { syncTransactionsAndPricesForAddress } from 'services/transactionService'
 import { syncCurrentPrices } from 'services/priceService'
 
 const getSocialLoginProviders = (): array => {
@@ -103,7 +103,7 @@ export const backendConfig = (): TypeInput => {
                 if (response.status === 'OK') {
                   // post sign in logic goes here
                   (await addressService.fetchAllUserAddresses(response.user.id)).forEach((addr) => {
-                    void syncTransactions(addr.address)
+                    void syncTransactionsAndPricesForAddress(addr.address)
                   })
                   void syncCurrentPrices()
                   return response
@@ -125,7 +125,7 @@ export const backendConfig = (): TypeInput => {
                   } else {
                     // post sign in logic goes here
                     (await addressService.fetchAllUserAddresses(response.user.id)).forEach((addr) => {
-                      void syncTransactions(addr.address)
+                      void syncTransactionsAndPricesForAddress(addr.address)
                     })
                     void syncCurrentPrices()
                   }
--- a/pages/api/dashboard/index.ts
+++ b/pages/api/dashboard/index.ts
@@ -132,6 +132,7 @@ interface Payment {
 const getUserDashboardData = async function (userId: string): Promise<DashboardData> {
   const buttons = await paybuttonService.fetchPaybuttonArrayByUserId(userId)
   const addresses = await addressService.fetchAllUserAddresses(userId, true)
+  void await transactionService.syncTransactionsAndPricesForAddressList(addresses.map(addr => addr.address))
   const XECAddressIds = addresses.filter((addr) => addr.networkId === XEC_NETWORK_ID).map((addr) => addr.id)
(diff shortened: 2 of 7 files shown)


johnkuney:
Hmm the app isnt working for me...getting a 500 error on /api/dashboard. Not sure what is causing it. Any ideas? I have tried deleting and restarting everything...Is there a new .env config or something Im missing?

Klakurka:
Tests failing.

chedieck:
> Hmm the app isnt working for me...getting a 500 error on /api/dashboard. Not sure what is causing it. Any ideas? I have tried deleting and restarting everything...Is there a new .env config or something Im missing?

Does the error come with any message? It should take a couple of seconds to load, since it will be syncing the transaction again. The env file should have the three URLs, then you can just run `make reset-dev` in the branch and test it.

johnkuney:
Also seeing these
grpc-bchrpc-node: failed network integrity check
error - Error: [DecimalError] Invalid argument: undefined

johnkuney:
Working now, but Im not following the test plan...Do I need to install something or something? Whats yd?

chedieck:
> 

Sorry, `yd` is `yarn docker`, is just an alias that I use (so much that I forgot it was not the actual name), fixed in the description.

johnkuney:
Okay sweet thanks. Yeah looks good to me!
Test plan worked, was able to delete the rows and then they came back on page view. 

I think we should definitely add some loading screens though. Not sure if there is already a task for it, but I can add and tackle in another pr if not

johnkuney (approved):

Klakurka:
We don't have any tasks for loading screens / spinners/ etc but I was just thinking about that last week. Maybe make one for whatever scope you think is reasonable for a single branch and we can start there.

Klakurka (changes requested):
I think the performance hit on this one is just too much to be ok with.

Maybe @ScottMcDermid or @Gattaca1 have some ideas for a better solution?

Like if I was to approach this, I might do something like:

- Sync address txs when accessed from the API (every time, before returning results). Remember that this will be protected behind AWS so we can choose how often the queries actually get through to the server.
- Have a process in the bg that tracks who's "online" and continually sync address txs for all of their button addresses.
- No sync action on page view.

But this is just my initial thought and I imagine you guys might be able to come up with something better.

marinacabrera:
some ideas (in the short term):

on the client-side (dashboards): https://nextjs.org/docs/basic-features/data-fetching/client-side#client-side-data-fetching-with-swr

on the server-side (API Routes): add a redis service, let critical endpoints hit it first, if not, populate it after fetching

chedieck:
> I think the performance hit on this one is just too much to be ok with.

yeah, I agree.
> Maybe @ScottMcDermid or @Gattaca1 have some ideas for a better solution?
> 
> Like if I was to approach this, I might do something like:
> 
>     * Sync address txs when accessed from the API (every time, before returning results). Remember that this will be protected behind AWS so we can choose how often the queries actually get through to the server.

The endpoint `transactions/sync/[address]` does that.

I think that the real solution here is using chronik, which will solve this once for all by syncing exactly the necessary & sufficient amount. Until then, I don't think we should spend too much time in another sub-optimal solution, so my suggestion is to just implement some button that does the sync, so that there is no impact on the performance in any views, and if the user feels like something is out of sync, they can confortably hit a button and know that the sync is happening.

Klakurka:
I don't like the button idea and also we can't switch BCH to chronik so we do need a better solution anyway.

The solution needs to work for the client too - as in every time the API is hit, that needs to be showing live data (even more important than making sure the server is up to date technically).

chedieck:
I think we can close this, I will reimplement it by creating a task that uses https://www.grpc.cash/docs/bchrpc/#pb.SubscribeTransactionsRequest with the DB addresses, so that we get transactions notifications.

> The solution needs to work for the client too - as in every time the API is hit, that needs to be showing live data (even more important than making sure the server is up to date technically).

I think in this case it suffices for the client to use the API endpoint I mentioned above, which will first sync all transactions before sending the data. I will also create a column `lastSyncBlockHeight` on the `Network` table so that we don't redundantly resync transactions that were already synced.

Klakurka:
Ok. Sounds like a better way that will be closer to what we do with Chronik anyway.

PR #315, opened 2022-11-14
refactor: remove alternative address routes

chedieck:
**Description:**
Remove alternative routes.

**Test plan:**
On master, running:
```
curl http://localhost:3000/balance/[address]
```
or
```
curl http://localhost:3000/address/balance/[address]
```
have the same effect. The same applies to:
```
curl http://localhost:3000/transactions/[address]
curl http://localhost:3000/address/transactions/[address]
```

Now, only the first ones should work.


Klakurka (changes requested):
On this branch when I try to access /transactions/xxx for an address that we don't have yet, it just returns this instead of returning the latest data:

`{"statusCode":404,"message":"No address found."}`

Is this what you're seeing? If that's the case then it'll break the client setup unless there another way to get the data.

Klakurka:
Also I know it might be annoying to swap them but was there reason you chose the shorter URLs over the longer ones that included /address/ ? I think the client is expecting /address/ and also do you think this might be more clear to use the longer URL as the project grows and new pages are added?

chedieck:
> Also I know it might be annoying to swap them but was there reason you chose the shorter URLs over the longer ones that included /address/ ? I think the client is expecting /address/ and also do you think this might be more clear to use the longer URL as the project grows and new pages are added?

I chose the short ones merely for convenience, I will swap them.

Klakurka:
I just triple checked and the client uses the following:

- https://api.paybutton.org/address/transactions/[address]
- https://api.paybutton.org/address/balance/[address]

chedieck:
Update:
I left the route rewriting out and simply moved the files into the `api/address` directory, so that now the URLs that are expected are the ones that start with `address`.

Klakurka (approved):

ScottMcDermid (approved):
Code looks good

Issue #317, opened 2022-11-21
Edit paybuttons


chedieck:
Implement logic to edit exists paybuttons from the front end

PR #318, opened 2022-11-21
[Issue #317] feat: show button data on button detail

chedieck:
Description:
Show the button data in the paybutton detail view.
image

Part of #317 so that the editing of this data is visible.

Test plan:
Go to any button detail view and see if the info shown above is present.

### DIFF
--- a/components/Paybutton/PaybuttonDetail.tsx
+++ b/components/Paybutton/PaybuttonDetail.tsx
@@ -8,7 +8,6 @@ interface IProps {
   onDelete: Function
 }
 export default ({ paybutton, onDelete }: IProps): FunctionComponent => {
-  console.log(paybutton)
   return (
     <div className={style.paybutton_list_ctn}>
       <div className={`${style.paybutton_card} ${style.paybutton_card_no_hover}`}>
--- a/components/Paybutton/PaybuttonDetail.tsx
+++ b/components/Paybutton/PaybuttonDetail.tsx
@@ -1,5 +1,5 @@
 import React, { FunctionComponent } from 'react'
-import  { PaybuttonWithAddresses } from 'services/paybuttonService'
+import { PaybuttonWithAddresses } from 'services/paybuttonService'
 import style from './paybutton.module.css'
 import EditButtonForm from './EditButtonForm'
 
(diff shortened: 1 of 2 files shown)


chedieck:
@johnkuney I imagine you have ideas of making this visually better, if that's the case you can edit this PR directly.

johnkuney:
What is the button data exactly? Is this something the user will be able to add through the frontend? Is it always a json object? But it can have any keys and values?  

Test plan is working for me here btw, just trying to understand better for potential design improvements

johnkuney:
Just adjusted the margins a bit and added a check to not display it if it is an empty object. Well and I guess its actually a string?

chedieck:
> What is the button data exactly? Is this something the user will be able to add through the frontend? Is it always a json object? But it can have any keys and values?

Those are all good questions, for which I don't know the answer, except that the button data is a JSON object that we save as a long string (due to how Prisma works with MySQL).

johnkuney:
Okay for sure, maybe @Klakurka can clarify. But other than that looks good to me

johnkuney (approved):

Klakurka:
Umm... Let me think about this one. I forgot what we wanted to do with this one but it might have been made obsolete.

Klakurka (changes requested):
Ok so my plan for this one is a bit different from what we're doing here.

Here's the client task for it: https://github.com/PayButton/paybutton/issues/66

It's supposed to somehow be attached to a payment. Imagine how a button might be in a particular state on a webpage that the person/business wants to keep track of. We already know the currency, address, and amount, but for example they might want to know which page it was on, which item(s) were being purchased, etc. You can see in the original client task that we're thinking of allowing them to actually generate effectively a form to fill in data (like name or something) but we can discuss this one later and it's really only a concern on the client if we do this in a generic enough way on the server.

So for now, I think what we should do is move this into `payments` (so URL could be `payments/1` where we show everything including json data. BUT for now hide the page/link -- it'll just be for later when we need it.

Klakurka:
There's also the question of how do we attach this data to a particular payment. There are a few ways to go about it but for now we'll not worry about that as this is a pretty low prio task.

Klakurka:
It's per payment.

"What state was the button in when the payment happened?" Is what this is.

- no
- no

chedieck:
>It's per payment.

Then `buttonData` should not even be a column on the `Paybutton` table, as it is now. I'll remove it from the edit logic in this PR and then create other PRs deleting the column.

Klakurka:
Yep sounds good.

chedieck:
> I'll remove it from the edit logic in this PR 

Though I was at #319. Closing this, since it no longer makes sense.

PR #319, opened 2022-11-22
[#Issue 317] feat: edit paybuttons backend

chedieck:
Description:
Create API endpoint to edit paybuttons. No front end for accessing it implemented yet.

Test plan:
Tests should be covering it. There will be a next PR implementing the front end logic that depends on this one and makes testing from the front end possible; `curl` commands won't work because of the supertokens session authentication.

~Depends on:~ 
~- [ ] #318~


johnkuney (approved):
Looks good to me

chedieck:
I found some missing logic when finishing the front end; will mark as ready for review when the front end PR is also submitted.

ScottMcDermid (approved):
Code looks good

PR #320, opened 2022-11-23
[Issue #317] feat: edit paybuttons frontend

chedieck:
Description:
Implements the front end for editing paybuttons.

Test plan:
Go to http://localhost/button/1 or any other button detail view and try to edit it.

Depends on:
- [x] #319


Klakurka (changes requested):
I've never seen an edit form use the form labels showing the past value but I'm good with that.

Issues:
 - Entering an invalid address throws a 500 error.
 - Address placeholder label could use some extra room to show the entire address rather than just a bit that barely helps in knowing which address is entered in there.

chedieck:
> I've never seen an edit form use the form labels showing the past value but I'm good with that.

Idd, maybe in the future we add a per-field edit such that if the user only want to edit one field, it's not ambiguous what will happen to the other ones.

> Entering an invalid address throws a 500 error.

Fixed

> Address placeholder label could use some extra room to show the entire address rather than just a bit that barely helps in knowing which address is entered in there.

Yeah, here I made and adaptation in the case that "single address" is selected; because then there is only one available line. If you select "multiple addresses", then the multiple addresses are shown one per line.

My suggestion would be to actually drop this whole multiple/single address logic (here and in the create button view) and just use a multi-line field anyway, since parsing one line or multiple lines is basically the same. We could then have a hint button explaining "enter one address per line", so if it's only one address, it's just one line.

johnkuney:
Looks good! One thing I noticed is looks like pressing enter will actually trigger the delete button dialog instead of submit. I'm fine to address that in another pr if you dont want to mess with it rn though

_Yeah, here I made and adaptation in the case that "single address" is selected; because then there is only one available line. If you select "multiple addresses", then the multiple addresses are shown one per line._

Im not sure I follow why cutting the address off is needed for the single address?

chedieck:
> Looks good! One thing I noticed is looks like pressing enter will actually trigger the delete button dialog instead of submit. I'm fine to address that in another pr if you dont want to mess with it rn though

Well noted, I think it would be best addressed in a separate PR.


> Im not sure I follow why cutting the address off is needed for the single address?

Only in the case there is more than one addresses, because then it would not fit in the single line width.

chedieck:
I've set changed it to always accept multiple addresses and thus always display all addresses separated by a line break on the field placeholder.

Klakurka (dismissed):
We can do this in another branch but when you change your address(es), the page will still list the transactions and such from the original address(es) until you refresh.

ScottMcDermid (approved):
Looks good

Klakurka:
Conflicts

Klakurka (approved):

PR #326, opened 2022-11-26
[Issue #324] feat: dependencies for current price endpoint

chedieck:
Description:
Dependencies for the implementation of #324. Some refactors and the creation of constants.

Test plan:
Nothing should change.

Issue #324, opened 2022-11-25
Price API endpoint


Klakurka:
The button will request a particular currency (BCH or XEC) valued in one fiat currency (currently only USD and CAD are available).

Assume they always want the latest unless it makes sense to re-use an API we're already using for the server if we're doing that.

eg. `/price/xec/cad`

Default to USD if only /xec or /bch is entered.

PR #327, opened 2022-11-26
 [Issue #324] feat: current price endpoint

chedieck:
**Description:**
Creates an API endpoint to get current prices from the DB.

**Test plan:**
Tests should be covering. Experiment with requests such as:
```
curl http://localhost:3000/price/bitcoincash
```
should return `bitcoincash`'s current price in USD,
```
curl http://localhost:3000/price/ecash/cad
```
should return `ecash`'s current price in CAD

----

Depends on:
- [ ] #326


Klakurka (approved):

ScottMcDermid (approved):
Works for me! Code looks good

Issue #322, opened 2022-11-25
Sync only new txs by subscribing addresses


chedieck:
Detailing of #252.
---
I think we can close this, I will reimplement it by creating a task that uses https://www.grpc.cash/docs/bchrpc/#pb.SubscribeTransactionsRequest with the DB addresses, so that we get transactions notifications.

> The solution needs to work for the client too - as in every time the API is hit, that needs to be showing live data (even more important than making sure the server is up to date technically).

I think in this case it suffices for the client to use the API endpoint I mentioned above, which will first sync all transactions before sending the data. I will also create a column `lastSyncBlockHeight` on the `Network` table so that we don't redundantly resync transactions that were already synced.

_Originally posted by @chedieck in https://github.com/PayButton/paybutton-server/issues/314#issuecomment-1322642527_

PR #333, opened 2022-11-29
[Issue #332] feat: add redis

chedieck:
Description:
Part of #322. Add redis to our project so we can later add [bullmq](https://github.com/taskforcesh/bullmq).

Test plan:
1. After pulling, first run `make reset-dev`
2. After all containers are up, inside the project directory run `yarn docker c` to enter the redis command line.
3. Run `KEYS *`. It should be empty.

 Now, we are going to try to modify the redis cache from within our Next.JS project:
1. Create a file `test.patch` inside the project directory with the following content:
```
diff --git a/pages/api/dashboard/index.ts b/pages/api/dashboard/index.ts
index 6953f9d..f03287f 100644
--- a/pages/api/dashboard/index.ts
+++ b/pages/api/dashboard/index.ts
@@ -5,6 +5,7 @@ import { XEC_NETWORK_ID, BCH_NETWORK_ID } from 'constants/index'
 import { Prisma } from '@prisma/client'
 import moment, { DurationInputArg2 } from 'moment'
 import { setSession } from 'utils/setSession'
+import { redis } from 'redis/clientInstance'

 interface ChartData {
   labels: string[]
@@ -202,6 +203,8 @@ export default async (req: any, res: any): Promise<void> => {
     await setSession(req, res)
     const userId = req.session.userId

+    redis.set('example', 'works')
+
     res.status(200).json(await getUserDashboardData(userId))
   }
 }
```
2. Apply the patch above by running `patch -p1 < test.patch`
3. Acessing the dashboard (http://localhost:3000) now should set the key 'example' to the value 'works'. Verify that by running again `KEYS *` in the redis command line, which should yield `1) "example"`
4. Likewise, `GET example` should return `works`

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -4,4 +4,5 @@ APP_NAME="PayButton"
 API_BASE_PATH=/api
 API_DOMAIN=http://localhost:3000
 WEBSITE_DOMAIN=http://localhost:3000
+REDIS_URL="redis://paybutton-cache:6379"
 WEBSITE_BASE_PATH=
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,5 @@ yarn-error.log*
 
 # vercel
 .vercel
+
+redis/dump.rdb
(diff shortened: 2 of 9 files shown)


ScottMcDermid (approved):
Looks good! Cache is working for me

PR #335, opened 2022-12-09
Style tweaks for widescreens and button page

johnkuney:
Added some styles to fit the screen better on larger monitors and improve the style on the addresses above the tables on the /button pages

### DIFF
--- a/components/Layout/layout.module.css
+++ b/components/Layout/layout.module.css
@@ -8,6 +8,9 @@
   padding-right: 25px; */
   display: flex;
   position: relative;
+  max-width: 1800px;
+  background-color: var(--primary-bg-color);
+  margin: auto;
 }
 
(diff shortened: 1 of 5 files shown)


Klakurka (changes requested):
I'm running a 4k monitor (not ultrawide, but still big) and I don't really like the changes. IMO what we have in master is better.

In this branch, it makes everything look more like 4:3 and has these sorta blue-ish grey bars on the sides.

The button transactions page changes are fine though.

johnkuney:
Ah for sure. I only have my laptop screen to go off of, and in the dev window it seemed too stretched out at simulated wider sizes.  But if its fine IRL thats good for me, reverted those changes

Klakurka (approved):

Issue #301, opened 2022-11-08
Some other cron process to sync current prices


chedieck:
Maybe let's make a new task to sync on some other cron process or something because this sounds like a lot of requests if we have a bunch of users.

_Originally posted by @Klakurka in https://github.com/PayButton/paybutton-server/issues/300#issuecomment-1306379811_

PR #337, opened 2022-12-11
[Issue #301] feat: using Bullmq for tx sync

chedieck:
**Description:**
Uses Bullmq to run price/transactions sync jobs, instead of doing so upon user login (as before).

**Test plan:**
It should work almost as before, with the exception that now the sync will happen together with `next dev`, independently of anybody logging in or not.

One idea is to run `make reset-dev` to reset the project and then, after the node container is up, before logging in in http://localhost:3000, run

```
$ yarn docker db
> SELECT * FROM Price;
> SELECT * FROM Transaction;
```

to see if the prices & the transactions that belong to the seeded addresses, along with the current prices, are being synced.

Depends on:
- [x] #336


Klakurka:
My build gets stuck in a restart loop on this one.

`2022-12-10 18:11:35 /usr/local/bin/docker-entrypoint.sh: exec: line 11: ./scripts/paybutton-server-start.sh: not found`

chedieck:
> My build gets stuck in a restart loop on this one.
> 
> `2022-12-10 18:11:35 /usr/local/bin/docker-entrypoint.sh: exec: line 11: ./scripts/paybutton-server-start.sh: not found`

Should be fixed now

Klakurka (approved):

PR #339, opened 2022-12-13
fix: add openssl1.1-compat to alpine node container

chedieck:
Fixes the:
```
Error: Unable to require('/home/node/src/node_modules/@prisma/engines/libquery_engine-linux-musl.so.node')
Error loading shared library libssl.so.1.1: No such file or directory (needed by .../node_modules/prisma/libquery_engine-linux-musl.so.node)
```
 error.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 FROM node:lts-alpine
-RUN apk add --no-cache python3 make g++
+RUN apk add --no-cache python3 make g++ openssl1.1-compat
 USER node
 WORKDIR /home/node/src/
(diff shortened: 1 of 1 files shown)


Klakurka (changes requested):
Mine is still stuck in a loop when building on this branch.

Klakurka:
Yep... so maybe somebody else can see if it's working for them.

johnkuney:
Its working for me

ScottMcDermid (approved):
Works for me

Klakurka (approved):

Issue #88, opened 2022-06-01
Add socket connection for payment detection


Klakurka:
The server portion of https://github.com/PayButton/paybutton/issues/9

Klakurka:
We'll want to make sure this works through cloudfront so that we don't open ourselves up to FOS.

PR #345, opened 2022-12-29
[Issue #88] feat: automatic detect payments

chedieck:
**Description:**
Solves #88. Creates #346.
Addresses are now subscribed when they are added, such that if a new transactions appears on any address, it is automatically added to the DB.

**Test plan:**
After building, create a paybutton with of **your** addresses. Then, using [Electrum ABC](https://github.com/Bitcoin-ABC/ElectrumABC) or any other XEC wallet, transfer some small quantity to that address. After the transaction is confirmed, it should appear on  the button detail. Same thing with Bitcoin Cash.

Depends on:
- [x] #342

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ yarn-error.log*
 .vercel
 
 redis/dump.rdb
+jobs/out.log
--- a/constants/index.ts
+++ b/constants/index.ts
@@ -40,6 +40,9 @@ export const RESPONSE_MESSAGES = {
 // When fetching some address transactions, number of transactions to fetch at a time.
 export const FETCH_N = 100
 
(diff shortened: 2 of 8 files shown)


ScottMcDermid (approved):
Code looks good. Transactions are populated as expected.

Klakurka (commented):
It doesn't work for me for some reason (new txs not showing up) but I'll maybe try again later if we don't have any other ideas.

Klakurka:
Ok it seems to not work on particular addresses... maybe just addresses that have been used already?

Gattaca1:
Works for me. 

It would be nice if there was some sort of user feedback to let a user know that its syncing. Initially I thought it didn't work but I was just being impatient

Issue #349, opened 2022-12-31
Handle mempool txs as well (not just ones that have been confirmed)


Klakurka:
- Show if finalized (XEC only)
- Remove if they become invalid later

No need to tell people if they've been confirmed or not -- just drop them if they become invalid later.

chedieck:
- [ ] Subscribe addresses to data stream of mempool txs (#350)
- [ ] Also fetch unconfirmed txs when syncing addresses
- [ ]  Remove invalid transactions & not show if txs are confirmed or not in the button detail view.

Klakurka:
We could make a separate task for this but for BCH, we can have the 'confirmed' column like we have it in #350 but for XEC we'll want to use 'Finalized' instead. We might not be able to do finalized until we switch to Chronik though.

Klakurka:
We're still showing a 'Confirmed' column for XEC but this should be changed to 'Finalized'.

Klakurka:
The rest will be done as part of #240 and #241.

PR #350, opened 2023-01-02
[Issue #349] feat: subscribe addresses to unconfirmed txs

chedieck:
Description:
Part of #349. Subscribes addresses also to unconfirmed transactions, so that they are synced as soon as they get into the mempool. For now, I decided showing if they are confirmed or not; will remove that in the PR that removes from the DB transactions that become invalid.

Test plan:
Same as #345, but now waiting for unconfirmed txs. Also, more detailed output now will be visible by running `yarn docker jobs` or examining the file `jobs/out.log`.

Depends on:
- [x] #345

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -68,7 +68,7 @@ model Transaction {
   id        Int                    @id @default(autoincrement())
   hash      String                 @db.VarChar(255)
   amount    Decimal                @db.Decimal(24, 8)
-  confirmed Boolean                @default(true)
+  confirmed Boolean                @default(false)
   timestamp Int
   addressId Int
   address   Address                @relation(fields: [addressId], references: [id], onDelete: Cascade, onUpdate: Restrict)
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -68,6 +68,7 @@ model Transaction {
   id        Int                    @id @default(autoincrement())
   hash      String                 @db.VarChar(255)
   amount    Decimal                @db.Decimal(24, 8)
+  confirmed Boolean                @default(true)
   timestamp Int
   addressId Int
   address   Address                @relation(fields: [addressId], references: [id], onDelete: Cascade, onUpdate: Restrict)
--- a/services/transactionService.ts
+++ b/services/transactionService.ts
@@ -108,7 +108,7 @@ export async function base64HashToHex (base64Hash: string): Promise<string> {
   )
 }
 
-export async function upsertTransaction (transaction: Transaction.AsObject, address: Address): Promise<TransactionWithAddressAndPrices | undefined> {
+export async function upsertTransaction (transaction: Transaction.AsObject, address: Address, confirmed=true): Promise<TransactionWithAddressAndPrices | undefined> {
   const receivedAmount = await getTransactionAmount(transaction, address.address)
   if (receivedAmount === new Prisma.Decimal(0)) { // out transactions
     return
@@ -118,7 +118,8 @@ export async function upsertTransaction (transaction: Transaction.AsObject, addr
     hash,
     amount: receivedAmount,
     addressId: address.id,
-    timestamp: transaction.timestamp
+    timestamp: transaction.timestamp,
+    confirmed: confirmed
   }
   return await prisma.transaction.upsert({
     where: {
--- a/services/transactionService.ts
+++ b/services/transactionService.ts
@@ -108,7 +108,7 @@ export async function base64HashToHex (base64Hash: string): Promise<string> {
   )
 }
 
-export async function upsertTransaction (transaction: Transaction.AsObject, address: Address, confirmed=true): Promise<TransactionWithAddressAndPrices | undefined> {
+export async function upsertTransaction (transaction: Transaction.AsObject, address: Address, confirmed = false): Promise<TransactionWithAddressAndPrices | undefined> {
   const receivedAmount = await getTransactionAmount(transaction, address.address)
   if (receivedAmount === new Prisma.Decimal(0)) { // out transactions
     return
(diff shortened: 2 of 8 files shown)


prisma/schema.prisma
```diff
@@ -68,6 +68,7 @@ model Transaction {
   id        Int                    @id @default(autoincrement())
   hash      String                 @db.VarChar(255)
   amount    Decimal                @db.Decimal(24, 8)
+  confirmed Boolean                @default(true)
```
Gattaca1, line 71: Whats the reason to have this default to `true` instead of `false`? Seems like it could potentially bite us

services/transactionService.ts
```diff
@@ -108,7 +108,7 @@ export async function base64HashToHex (base64Hash: string): Promise<string> {
   )
 }
 
-export async function upsertTransaction (transaction: Transaction.AsObject, address: Address): Promise<TransactionWithAddressAndPrices | undefined> {
+export async function upsertTransaction (transaction: Transaction.AsObject, address: Address, confirmed=true): Promise<TransactionWithAddressAndPrices | undefined> {
```
Gattaca1, line 111: Same here, defaulting to `confirmed = true `looks potentially dangerous

prisma/schema.prisma
chedieck: First because we were already only doing confirmed txs, but also because I think the idea of saving the transaction is mostly those that actually happened. The 'non confirmed' ones are actually the exception, because they are not really transactions, only potential transactions, so when we save them we make it clear that we are not saving a confirmed transaction and that necessarily appears on the code, since the default is `true`.

ScottMcDermid: I think you're arguing mostly from a point of code clarity, which is important. @Gattaca1 is arguing from a point of security. It's better to assume a transaction is unconfirmed in this case, I think.
Relying on our code to explicitly specify an unconfirmed transaction could result in human-error. There are 0-conf attacks that can reject an unconfirmed transaction from the mempool.
If there's a human-error that results in a false-negative (incorrectly flagging a transaction as unconfirmed), then the side-effects are less severe than a false-positive, I'd expect.

ScottMcDermid (changes requested):
Tested for BCH, and works as expected:
2023-01-13-151103_897x539_scrot

I tried adding a second button, and it didn't sync:
2023-01-13-151131_902x777_scrot

I do get an error on startup in my jobs outfile:

```
job syncBCHAddresses: syncing all addresses for network 2...
grpc-bchrpc-node: failed network integrity check
grpc-bchrpc-node: failed network integrity check
grpc-bchrpc-node: failed network integrity check
initial syncing FAILED for network 2
error for initial syncing of network 2: job syncBCHAddresses failed with error 
```

chedieck:
Refactored to default confirmed to false.
@ScottMcDermid, about your error:
1. Can you post the addresses you tried in text? With images I can't copy & paste them to try it out myself.
2. This error refer to the initial sync, the one that is made first when the containers are run. If you added new addresses later, you should look either for another error or for the message `synced [...]` with the addresses that were sync. You can post the full `jobs/out.log` if the error persists.

ScottMcDermid:
> 1. Can you post the addresses you tried in text? With images I can't copy & paste them to try it out myself.

Yes, sorry about that. I was able to recreate the symptoms again in a fresh env:
First button: `bitcoincash:qzwgnlshh8qdg0kjzklklz6w3qfmzmz6kyce7lgvmt` (synced corectly)
Second button: `bitcoincash:qqyy3mss5vmthgnu0m5sm39pcfq8z799ku2nxernca` (never syncs)
I added each button a few seconds apart.

> 2. This error refer to the initial sync, the one that is made first when the containers are run. If you added new addresses later, you should look either for another error or for the message synced [...] with the addresses that were sync. You can post the full jobs/out.log if the error persists.

Here's my `out.log` for the 2 addresses, above.
[out.log](https://github.com/PayButton/paybutton-server/files/10485048/out.log)

chedieck:
Thanks Scott, I was able to reproduce the error here.


The problem is that there are over 700k txs for the second address, so the current implementation goes on to sync them 100 at a time, and it just never finished.

I though about some some possible improvements to this, which I could address in some subsequent PRs (because this issue is  independent of this PR):

1. Progressively commit the tx syncing DB transactions:
   - Transactions would be added, their prices synced, e.g. 1000 at a time (for what I tested there is a byte limit when fetching this data, and for this address it got violated after ~5000 txs at once) and this would already show up in the frontend. It will still take a long time to sync all of them, but we would have the progress feedback.

2. Use another API to get price data
   - The coin.dance API we are using right now only allows to get the price of ONE day a time, which considerably slows the process of getting all txs prices.

3. Have a limit on how many txs to sync, until more is asked
   - Example: Fetch 1000 txs for some address. Only when the user, through the frontend, goes to e.g. page 50 of the tx list and expects to see older transactions do we then fetch the next 1000, and so on.

Any combination of the above is possible, so I am opening another issue  for we to discuss these solutions and also other ideas you guys might have.

ScottMcDermid (approved):
> The problem is that there are over 700k txs for the second address, so the current implementation goes on to sync them 100 at a time, and it just never finished.

Ah ya.. that's definitely part of the issue. By the way, I checked back a while later and there are transactions populated for this address, so I believe everything is working correctly as far as this task goes.

> I though about some some possible improvements to this, which I could address in some subsequent PRs (because this issue is independent of this PR):

I like the idea of your 3rd option's lazy loading, but maybe @Klakurka has an idea what to do. Either way, I think it's out of scope for the task.

Klakurka:
The lazy loading sounds fine to me. I might have more of an opinion if I see it working live.

PR #352, opened 2023-01-03
 [Issue #349] feat: sync addresses' unconfirmed txs 

chedieck:
Description:
Part of #349. When initially syncing existent addresses, also looks for unconfirmed txs and syncs them.

Test plan:
1. Stop the automatic syncing of addresses with `yarn docker js`
2. Create a paybutton with one of your addresses
3. Send a tx to it
4. Before it gets confirmed, sync it with `curl http://localhost:3000/address/transactions/sync/[youraddress]`. The output should display this transaction identifiable by the `confirmed: false` property.
6. It should also appear on the button detail view, also with the `confirmed` column set to `no`.

Depends on:
- [x] #350


Klakurka:
@chedieck conflicts

Klakurka:
I get this when trying to run `yarn docker js` in `paybutton-dev` (which I assume is the correct container:

image

chedieck:
> I get this when trying to run `yarn docker js` in `paybutton-dev` (which I assume is the correct container:

These `yarn docker` commands are not supposed to be run inside the containers, but outside. Part of what the commands do is to get inside the correct container, so if there is anyway you can get a unix command line your local clone of the `paybutton-server/` repo, you would only have to run `yarn docker js` from there (I don't understand much about windows, but I know that now exists something called WSL which maybe serves as that)

Klakurka:
I'm going to let Scott test this one.

ScottMcDermid (approved):
Tests work for me!
Code looks good

PR #354, opened 2023-01-06
[Issue #154] feat: edit wallet show errors

chedieck:
Description:
Editing wallets on http://localhost:3000/wallets now display errors on the modal, like it was already being done on http://localhost:3000/paybuttons.

Test plan:
Go to http://localhost:3000/wallets and try to modify an wallet in an invalid way, such as leaving it with no paybuttons or setting it as a default for some network which it has no addresses to.

Remarks:
- Fixed problem where was possible (on the backend) to set wallet with no XEC addresses to BCH default and vice e versa;
-  This edit feature still does not allow editing the wallet paybuttons, so that any unexpected behaviors related to it are still expected.


ScottMcDermid (approved):
Looks good to me

PR #356, opened 2023-01-09
[Issue #154] feat: wallet creation front end

chedieck:
Description:
Front end to create wallets.

Test plan:
Create some paybuttons, maybe some with an ecash address, others with a bitcoin cash address, others with both. Then try to create a wallet that include some of these addresses. Note that:
- Errors, such as "name not provided" should appear as a red text in the bottom of the form, such as it is being done on the "paybutton create" form;
- There should only be possible to create a wallet and set it as default for NETWORK if it contains an NETWORK address (where NETWORK is ecash or bitcoin cash);
- Editing the HTML before submitting the form, in order to try to set it as default for NETWORK when there are no NETWORK addresses selected should give the error `Default NETWORK wallet must have some NETWORK address.`

Depends on:
- [x] #354

### DIFF
--- a/components/Wallet/WalletForm.tsx
+++ b/components/Wallet/WalletForm.tsx
@@ -4,6 +4,7 @@ import { WalletPOSTParameters } from 'utils/validators'
 import { PaybuttonWithAddresses } from 'services/paybuttonService'
 import { XEC_NETWORK_ID, BCH_NETWORK_ID } from 'constants/index'
 import Image from 'next/image'
+import Link from 'next/link'
 import style from '../Wallet/wallet.module.css'
 import style_pb from 'components/Paybutton/paybutton.module.css'
 import Plus from 'assets/plus.png'
@@ -118,6 +119,7 @@ export default function WalletForm ({ userPaybuttons, refreshWalletList, userId 
 
                   <h4>Paybuttons</h4>
                   <div className={style.buttonlist_ctn}>
+                    {userPaybuttons.length === 0 && <div className={style.make_btn_p}>No Buttons yet. <Link href='/buttons'>Click here</Link> to make one!</div>}
                     {userPaybuttons.map((pb, index) => (
                       <div className={style.input_field} key={`create-pb-${pb.id}`}>
                         <input {...register('paybuttonIdList')}
--- a/components/Wallet/WalletForm.tsx
+++ b/components/Wallet/WalletForm.tsx
@@ -53,10 +53,10 @@ export default function WalletForm ({ userPaybuttons, refreshWalletList, userId 
     if (selectedPaybuttonIdList === undefined) return false
     for (const selectedPaybuttonId of selectedPaybuttonIdList) {
       let paybutton = userPaybuttons.find((pb) => pb.id === Number(selectedPaybuttonId))
-      if (paybutton === undefined) {
-        continue
-      }
-      if (paybutton.addresses.some((addr) => addr.address.networkId === networkId)) {
+      if (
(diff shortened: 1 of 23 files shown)


components/Wallet/WalletForm.tsx
```diff
@@ -1,54 +1,170 @@
...
+      )
+    }
+  }
+
+  function hasAddressForNetworkId(networkId: number): boolean {
+    let ret = false
+    if (selectedPaybuttonIdList === undefined) return false
+    for (const selectedPaybuttonId of selectedPaybuttonIdList) {
+      let paybutton = userPaybuttons.find((pb) => pb.id === Number(selectedPaybuttonId))
+      if (paybutton === undefined) {
```
ScottMcDermid, line 56: Can we avoid using `continue`? We could add this `if` condition to the next line:
```js
if (paybutton !== undefined && paybutton.addresses.some... )
```

ScottMcDermid (changes requested):
Tested, and validation works as expected. Just one code comment.

components/Wallet/WalletForm.tsx
chedieck: Addressed.

Klakurka (changes requested):
- Let's fade out the text for checkbox items that aren't selectable. It's not obvious without first adding a new button that is selectable and seeing the contrast in the checkbox.
image
- There's a weird dot in the middle of the dialog (see above).
- Remove 'exists for this user' from the duplicate name error message. The errors are for the people using this, not for us (the admins / devs).

chedieck:
>Let's fade out the text for checkbox items that aren't selectable. It's not obvious without first adding a new button that is selectable and seeing the contrast in the checkbox.

Addressed. @johnkuney please [check this commit](https://github.com/PayButton/paybutton-server/pull/356/commits/c7c4a0542d3a429598e1bba6e67b59a3f87a367a) to see if this is the right way to do it.


> There's a weird dot in the middle of the dialog (see above).

Fixed.

>  Remove 'exists for this user' from the duplicate name error message. The errors are for the people using this, not for us (the admins / devs).

Fixed.

johnkuney:
Looks great! I just tweaked the disabled style a tad and added a couple other style tweaks hope thats cool. 
- Added an explainer blurb at the top as its not clear why items may be disabled
- Added a line if there are no buttons yet, with a link to the buttons page
- Moved the "make default" checkboxes down on the edit form so it matches the create form

johnkuney:
hmmm guess I broke a check with those commits...not sure why

chedieck:
Thanks for the changes John. About the building breaking:
 
Let's leave on this PR the changes you made regarding only the **create** wallet logic, that is, exclude the last two commits ([2251315](https://github.com/PayButton/paybutton-server/pull/356/commits/2251315c4e7bd2b9e7d2e7f5abfe5b60a0166b80) and [484a298](https://github.com/PayButton/paybutton-server/pull/356/commits/484a298ee546ee7a32177808b59338320a7831c0)). Then, after this is approved, you can do it on [the next PR](https://github.com/PayButton/paybutton-server/pull/357), regarding editing a wallet's paybuttons, that has this as a requirement.

It's very likely that those changes being made here would lead to conflicts, and on this branch editing some wallets paybuttons does not work, so it's better to use a state where it works (next PR) and then making the changes to better the style.

Klakurka (approved):
Ok I'll approve but I made this task for the UI improvements that should go in soon: https://github.com/PayButton/paybutton-server/issues/371

Issue #366, opened 2023-01-20
Refactor decimal places logic


chedieck:
Refactor logic to display amounts with the expected number of decimal places.
Part of this is #348

PR #367, opened 2023-01-20
[Issue #366] refactor: decimal numbers for each quote

chedieck:
**Description:**
Refactor backend to display amounts with the correct number of decimal places. Also now  the total revenue is displayed with a max of 3 decimal places, in the dashboard 

image

**Test plan:**
All should work as before, except for the possible extra decimal place on the place seen in the screenshot above.


Klakurka (commented):
How about only showing the 3rd decimal if the total value is less than $0.01?

chedieck:
> How about only showing the 3rd decimal if the total value is less than $0.01?

I redid it, now it shows the first significant digit when it's less than $0.01.
Examples:
If the value is $0.0028; shows $0.003
If the value is $0.00001222; shows $0.00001

PR #373, opened 2023-02-01
bigger base image that has the necessary packages to run on M1 also

vorg-code:
Please delete `node_modules` folder and docker images before testing everything with `make dev`.

I used a bigger starting docker image that runs in M1 in the end. Also, I believe it has some of the packages that were manually installed afterwards, I removed them.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
-FROM node:lts-alpine
-RUN apk add --no-cache python3 make g++ tmux openssl1.1-compat
+FROM node:lts
+RUN apt-get update && apt-get -y install tmux
 USER node
 WORKDIR /home/node/src/
 ENV PATH /home/node/src/node_modules/.bin:$PATH


Dockerfile
```diff
@@ -1,5 +1,5 @@
-FROM node:lts-alpine
-RUN apk add --no-cache python3 make g++ tmux openssl1.1-compat
+FROM node:lts
+RUN apt-get update && apt-get -y install tmux
```
ScottMcDermid, line 2: Why leave out the equivalent apt-based `python3` and `openssl1.1-compat` packages? Are they bundled in this image?

Dockerfile
```diff
@@ -1,5 +1,5 @@
-FROM node:lts-alpine
```
ScottMcDermid, line 1: Changing our image away alpine is a bit of a large change if we're just trying to get your env up and running. Did you see somewhere that node-alpine isn't compatible with your machine?

ScottMcDermid (changes requested):
Unit and integration tests work for me, just a few questions

ScottMcDermid (approved):
Ah missed your initial comment. I think it's fine to switch to a non-alpine image. Everything seems to still be working on my linux instance.

Klakurka (approved):
Works for me.

Issue #171, opened 2022-07-17
Payment detection server response not compatible with client (broken)


Klakurka:

Klakurka:
Ok so this is due to the API pulling data from the DB and the DB not getting updated without manually logging in each time.

chedieck:
Is this solved with #313 ?

Klakurka:
I think this is still not working. At least I have questions about it in https://github.com/PayButton/paybutton-server/pull/315.

chedieck:
@Klakurka is this solved now?

Klakurka:
The client polling is still broken so we'll need to get that fixed up first before closing this one.

https://github.com/PayButton/paybutton/issues/195

Klakurka:
This one fails:

> http://localhost:3000/address/transactions/bitcoincash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7q9ayphulp

This one works fine:

> http://localhost:3000/address/balance/bitcoincash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7q9ayphulp

vorg-code:
Before we were answering an error when a new wallet (that wasn't in our database) got sent through this endpoint /address/transactions, so, frontend paybutton just got lots of silent errors in the network tab and no transaction would ever be recognized there.

Now if an address that we don't have hits this endpoint, we add it to our database, fetch the transactions from the node and then return the list of transactions. If this is an address with a boatload of transactions, this endpoint could take 1 minute to answer the first time and some seconds in each request, but it should not fail anymore.

Unit test was also changed to the new situation.

PR #375, opened 2023-02-02
[Issue #171] fix: server accepts new addresses and syncs them

vorg-code:
Before we were answering an error when a new wallet (that wasn't in our database) got sent through this endpoint /address/transactions, so, frontend paybutton just got lots of silent errors in the network tab and no transaction would ever be recognized there.

Now if an address that we don't have hits this endpoint, we add it to our database, fetch the transactions from the node and then return the list of transactions. If this is an address with a boatload of transactions, this endpoint could take 1 minute to answer the first time and some seconds in each request, but it should not fail anymore. 

Unit test was also changed to the new situation.


Klakurka:
@vorg-code looks like one of the integration tests is failing.

chedieck:
@vorg-code try to stick with the:
```
Description:
...
Test plan:
....
```
structure when creating a PR.


About the PR, indeed this endpoint `address/transactions/[addr]` returns an error for non existent addresses; but there is already another endpoint, which is `address/transactions/sync/[addr]` which does exactly that of syncing up and then returning the results. So if there are calls that are failing for non already added addresses, you could replace URL of these calls by adding the `/sync/`.

chedieck (changes requested):
.

vorg-code:
@chedieck what motivated this PR was that the Paybutton client calls this endpoint to get info on the transactions when we click on the donate button:
image
when the qrcode appears, we start firing multiple times against this `address/transactions/[addr]` endpoint:
image
Since @Klakurka said that there is no need anymore for the address to be pre-registered in the server, we should accept non registered addresses coming from paybutton client. Do you think that it is better to change the logic in the client to be aware of this error of non registered address in the server so that the client calls the `address/transactions/sync/[addr]` endpoint, waits and then goes back to calling the other endpoint?

PS: I will use the structure from now on.

chedieck:
@Klakurka, what do you think about the endpoint distinction?

As far as I remember, the last time we discussed, we decided to keep both as a way to possibly avoid unnecessary overhead by not syncing every time the transactions is called. The way @vorg-code  implemented it now will only sync for non existent addresses; so this problem would not exist.

I am fine with deleting the `sync/` endpoint and keeping only this one, and maybe later have it accept some optional boolean `sync` parameter, which would sync even addresses that already exists (in order to get new transactions), but since we are already syncing all addresses through jobs, this is not priority.

pages/api/address/transactions/[address].ts
```diff
@@ -32,18 +33,22 @@ export default async (req: NextApiRequest, res: NextApiResponse): Promise<void>
       if (req.query.address === '' || req.query.address === undefined) {
         throw new Error(ADDRESS_NOT_PROVIDED_400.message)
       }
+      let transactions = []
       const address = parseAddress(req.query.address as string)
-      const transactions = await fetchAddressTransactions(address)
+
+      const addressInDB = (await fetchAddressesInList([address])).length
```
Comment (chedieck) [inline pages/api/address/transactions/[address].ts:39]: Since it's only one address, makes more sense to use `fetchAddressBySubstring` than `fetchAddressesInList`. Also, you should not include the `.length` on the variable declaration so that the name `addressInDB` reflects better what the variable is and also mainly because then you don't have to always call `fetchAddressTransactions` down there, since `addressInDB.transactions` will return you exactly that for the case that the address exists.

tests/integration-tests/api.test.ts
```diff
@@ -1010,7 +1010,7 @@ describe('GET /api/address/transactions/[address]', () => {
       }
     }
     const res = await testEndpoint(baseRequestOptions, transactionsEndpoint)
-    expect(res.statusCode).toBe(404)
+    expect(res.statusCode).toBe(200)
```
chedieck, line 1013: Since it does not return an error anymore, I think is good to also check the integrity of the response; i.e, if it is indeed the address transactions (I think there is another test already doing that, so copying and paste that piece of the code should suffice)

chedieck (changes requested):

pages/api/address/transactions/[address].ts
Comment (vorg-code) [reply pages/api/address/transactions/[address].ts:39]: Will do!

tests/integration-tests/api.test.ts
vorg-code: I got that error back in place, both in the api and in the test

vorg-code:
What ended up being done here is that there is now a flag `?server_only=1` that, when set like this, tells us to only retrieve what we have in the database and we will throw an error if address is not registered (old way things worked). If no flag is passed or is set to '0', and the address is not registered, we will search it in the blockchain, register address and transactions in the database and then respond (new way). 

Since the frontend paybutton uses this endpoint without a flag, it will not give any error for any correct address inputted there, so that is also fixed. If we need a fast call, we can set the flag now.

Also the test was back to the way that it was, covering the possibility of address not registered in our database (with the flag being passed as well).

Klakurka:
I'm going to think about this a bit but my initial reaction is that using a simple query string isn't ideal.

vorg-code:
@Klakurka what in functionality terms do you feel that is lacking in this solution?

Klakurka:
Functionally I think it does everything we want, might just want to adjust it a bit but I have to think.

Klakurka:
What are all of the API URLs somebody could use here just so we're on the same page?

vorg-code:
- address/balance/[addr] - gets balance of the address
- address/transactions/[addr]  - gets transactions of the address, if addr not in our db: add, get trx from node, saves in our db, answers
- address/transactions/[addr]?server_only=1 - gets transactions from our db, if addr not in our db answers with an error
- address/transactions/sync/[addr] - add addr to our db (if not there already) and syncs transactions from node

I think all of them have different use cases

Klakurka:
When we run `address/transactions/sync/[addr]`, how does it update over time? I forgot what we did here. I guess a more specific question would be "would calling any of the other `transactions` endpoints give us old data or will it always be ready/able to send back the latest data we have on that address?

vorg-code:
When we call `address/transactions/sync/[addr]` or `address/transactions/[addr]` (and addr is not on our db) we get the latest transactions from the node and update our database. 

However, after the response is sent from this endpoint back to the frontend that called it, that is it, we don't update anymore if not called again, as far as these endpoints are concerned.

We have another point in the code (`/jobs`) that spawns workers that are responsible by keeping our database in sync with the nodes prices and transactions behind the scenes, so that we can trust that our database is up to date. This seems to not be working so well, I tested adding an address and transferring XEC to that address and it was not automatically picked up and it never appeared in our database. But I think that is a `/jobs` issue and should be treated in a different PR.

ScottMcDermid:
Can you describe a scenario where we might use the `address/transactions/[addr]?server_only=1` endpoint?

With the new functionality of our current endpoints, we may not require it.

vorg-code:
@ScottMcDermid if we want to have just a quick check (getting all transactions for a single address from the node can take up to 2 minutes if it is a very active wallet). 
If we implement, for example, a mouseover popup with transaction details, either we get the info in a few seconds or we don't need it at all, since customer won't wait with the mouse on the same spot for more than a few seconds.

ScottMcDermid:
@vorg-code 
> if we want to have just a quick check (getting all transactions for a single address from the node can take up to 2 minutes if it is a very active wallet).

Ya that makes sense.

> If we implement, for example, a mouseover popup with transaction details, either we get the info in a few seconds or we don't need it at all, since customer won't wait with the mouse on the same spot for more than a few seconds.

This example makes sense, but can you describe a user scenario for this? I don't think we ever list untracked addresses in the client.

To me, the closest thing we have is an enhanced _Add New Address_ scenario:

1. Prompt user with a form that accepts addresses for a button
2. User enters a few ecash addresses
3. **Dynamically check server to see if we already track those addresses?**
4. **Show user information about already scanned addresses?**
5. User can review address data of existing addresses before submitting the form

I don't think this feature is planned, though, and a different endpoint could be created for this situation.

vorg-code:
@ScottMcDermid here in this client page, bottom right:
image
you can copy and paste any address, if you do paste an untracked one, it seamlessly works (with this PR, before it it gives an error and the person would need to open the server, add the address there, then come back here for it to work). I asked @Klakurka about this in the group and he said it should work even if the address was not previously added, so I thought this was a good solution (maybe I misinterpreted him, if this was not his intention)...

ScottMcDermid:
@vorg-code 
> he said it should work even if the address was not previously added, so I thought this was a good solution (maybe I misinterpreted him, if this was not his intention)...

Wouldn't specifying the `server_only=1` flag for an address that was not previously added throw an error?

ScottMcDermid:
@vorg-code 

Sorry, misread. That scenario sounds good to me! Plus the functionality is already in place, so not a huge deal if we just leave it unused in the end.
We can always change the name of the endpoint in the future if we end up using it in some other way.

vorg-code:
@ScottMcDermid 

> Wouldn't specifying the `server_only=1` flag for an address that was not previously added throw an error?

Yes it will. To call this endpoint with this flag, the caller must be ready to handle this possible error.

> We can always change the name of the endpoint in the future if we end up using it in some other way.

Yah 👍

pages/api/address/transactions/[address].ts
```diff
@@ -32,22 +33,34 @@ export default async (req: NextApiRequest, res: NextApiResponse): Promise<void>
       if (req.query.address === '' || req.query.address === undefined) {
         throw new Error(ADDRESS_NOT_PROVIDED_400.message)
       }
+      let transactions = []
       const address = parseAddress(req.query.address as string)
-      const transactions = await fetchAddressTransactions(address)
+
+      // this flag ?server_only=1 tells us to only retrieve what we have in the database
+      const serverOnly = req.query.server_only === '1'
```
Comment (chedieck) [inline pages/api/address/transactions/[address].ts:40]: Small thing, but we have been using camelCase for the request parameters (e.g. in `pages/api/transaction/[transactionId].ts`), so best to user `serverOnly` here to mantain consistency.

Comment (chedieck) [reply pages/api/address/transactions/[address].ts:40]: Or if you think snake_case is better for request parameters, you can argue for it, keep like that, and open an issue to fix it another places

chedieck (changes requested):
Works well, just left some comments about small changes.

pages/api/address/transactions/[address].ts
```diff
@@ -32,22 +33,34 @@ export default async (req: NextApiRequest, res: NextApiResponse): Promise<void>
       if (req.query.address === '' || req.query.address === undefined) {
         throw new Error(ADDRESS_NOT_PROVIDED_400.message)
       }
+      let transactions = []
```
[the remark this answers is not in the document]
Comment (vorg-code) [reply pages/api/address/transactions/[address].ts:36]: oh indeed, this was needed before when I had an if...else, thanks for pointing out

pages/api/address/transactions/[address].ts
Comment (vorg-code) [reply pages/api/address/transactions/[address].ts]: done

chedieck (approved):

Klakurka:
@vorg-code conflicts

vorg-code:
@Klakurka conflict solved

pages/api/address/transactions/[address].ts
```diff
@@ -32,22 +33,34 @@ export default async (req: NextApiRequest, res: NextApiResponse): Promise<void>
...
-        case NO_ADDRESS_FOUND_404.message: {
-          res.status(NO_ADDRESS_FOUND_404.statusCode).json(NO_ADDRESS_FOUND_404)
-          break
-        }
         case INVALID_ADDRESS_400.message: {
           res.status(INVALID_ADDRESS_400.statusCode).json(INVALID_ADDRESS_400)
           break
         }
+        case NO_ADDRESS_FOUND_404.message: {
+          res.status(NO_ADDRESS_FOUND_404.statusCode).send(NO_ADDRESS_FOUND_404)
```
Comment (chedieck) [inline pages/api/address/transactions/[address].ts:61]: Better stick to `.json` which is being used in the other cases

chedieck (changes requested):

pages/api/address/transactions/[address].ts
Comment (vorg-code) [reply pages/api/address/transactions/[address].ts]: done

chedieck (approved):

Klakurka (approved):

PR #387, opened 2023-02-08
[Issue #154] feat: allow for wallet creation to override paybutton wallet belongingness

chedieck:
**Description**:
Before, upon wallet creation, we would not allow a user to select paybuttons that already belonged to other wallets. Now this is possible, such that the selected paybuttons will be updated so that they belong to the new wallet.

Furthermore, users cannot "consume" other wallets by this process, i.e, create a new wallet that leaves some already existent wallet with no paybuttons.

**Test plan:**
Try to create a wallet, assure that is possible to override other wallets paybuttons', as long as they are left with at least one paybutton.

**Depends on:**
- [x] #386 
- [x] #389


Klakurka (approved):
Looks good; just want to double check that we do want to actually go ahead with merging this and the other related branch.

ScottMcDermid (approved):
Code looks good. Works as expected

PR #388, opened 2023-02-08
[Issue #154] feat: allow for wallet edit to override paybutton wallet belongingness

chedieck:
Description:
Before, upon wallet edit, we would not allow a user to select paybuttons that already belonged to other wallets. Now this is possible, such that the selected paybuttons will be updated so that they belong to the new wallet.

Furthermore, users cannot "consume" other wallets by this process, i.e, create a new wallet that leaves some already existent wallet with no paybuttons. Neither can the user leave the edit wallet with no paybuttons: at least one paybutton must be selected

Test plan:
Try to edit a wallet, assure that is possible to override other wallets paybuttons', as long as they & the wallet being edited are left with at least one paybutton.

Depends on:

- [x] #386 
- [x] #387
- [x] #389


ScottMcDermid (approved):
Looks good, and tests work for me

PR #389, opened 2023-02-09
[Issue #390] test: avoid .only statements on tests from being committed

chedieck:
**Description:**
Solves #390.
There was a `describe.only` on the code, that was mistakenly commited by me on January. This would make so that most of the `walletService` tests didn't run.

I've removed it & made so that the unittests git action also checks for `describe.only`s and `it.only`s across the tests folder, failing the action in case they're found.

**Test plan:**
Can only be tested by pushing a branch which contains such `.only` statements to github, [here](https://github.com/PayButton/paybutton-server/actions/runs/4133291387/jobs/7143017250) is an example of me testing it before.

### DIFF
--- a/.github/workflows/pull-request-lint-and-test.yml
+++ b/.github/workflows/pull-request-lint-and-test.yml
@@ -14,6 +14,8 @@ jobs:
           fetch-depth: 0
       - name: Install dependencies
         run: yarn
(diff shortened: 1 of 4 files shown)


ScottMcDermid (approved):
That is pretty dangerous - Good catch.
Works for me.

Issue #395, opened 2023-02-10
Remodel wallets to include only addresses and not paybuttons


chedieck:
Right now both paybuttons & addresses objects have a `walletId` field. This remodeling will ultimately remove the field from paybutton objects, such that all logic that uses `paybutton.walletId` should use `address.walletId` instead.

PR #396, opened 2023-02-10
[Issue #395] feat: remodel address to wallet on wallet creation

chedieck:
**Description:**
Creating a wallet now uses the logic of selecting addresses, not paybuttons.

**Test plan:**
Go to the wallet list view, try to create a wallet. The checkboxes disabling should work as before: not allowing e.g to create a wallet with only XEC addresses to be the BCH default; not allowing to leave a wallet with 0 addresses by selecting all of their addresses when creating a new one.

**Remarks:**
- The UI is rudimentary, it only shows the address options. I will later submit another PR which makes so that the user can see the addresses & the paybuttons that these belong to.
- After having created the wallet, it won't set the `paybutton.walletId` anymore, so it will appear in the wallet list view as if it didn't have any paybutton associated to it. This also will be addressed in a different PR.

**Depends on:**
- [ ] #388

### DIFF
--- a/components/Wallet/WalletForm.tsx
+++ b/components/Wallet/WalletForm.tsx
@@ -22,10 +22,12 @@ export default function WalletForm ({ userPaybuttons, refreshWalletList, userId 
   const [isXECDefaultDisabled, setIsXECDefaultDisabled] = useState(true)
   const [isBCHDefaultDisabled, setIsBCHDefaultDisabled] = useState(true)
   const [error, setError] = useState('')
-  const [selectedPaybuttonIdList, setSelectedPaybuttonIdList] = useState([] as string[])
+  const [selectedPaybuttonIdList, setSelectedPaybuttonIdList] = useState([] as number[])
+  const [disabledPaybuttonList, setDisabledPaybuttonList] = useState([] as PaybuttonWithAddresses[])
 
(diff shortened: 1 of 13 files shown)


Klakurka (approved):

components/Wallet/WalletForm.tsx
```diff
@@ -37,58 +37,58 @@ export default function WalletForm ({ userPaybuttons, refreshWalletList, userId
...
-      disabledPaybuttons
+    setDisabledAddressList(
+      disabledAddresses
     )
   }
 
-  function handleSelectedPaybuttonsChange(checked: boolean, paybuttonId: number): void {
-    const paybuttonIsSelected = selectedPaybuttonIdList.includes(paybuttonId)
+  function handleSelectedAddressesChange(checked: boolean, addressId: number): void {
+    const paybuttonIsSelected = selectedAddressIdList.includes(addressId)
```
ScottMcDermid, line 71: Rename to `addressIsSelected`

ScottMcDermid (changes requested):
Works for me. Just one comment.

components/Wallet/WalletForm.tsx
chedieck: Good catch, addressed.

ScottMcDermid (approved):

PR #397, opened 2023-02-10
[Issue #395] feat: remodel address to wallet on wallet creation dependencies

chedieck:
Description:
Some changes required by #396. Basically:
- API endpoint to fetch all of some user addresses & tests for it,
- A error message constant to say that the user has not selected any address

Test plan:
`curl http://localhost:3000/addresses?userId='dev-uid'`  should return all addresses of the main dev user.


Klakurka (changes requested):
I just get an empty array:

`[]`

Klakurka (approved):
Nevermind -- browser I tested in just didn't like the `'` but clearly the system is fine with it.

ScottMcDermid (approved):
Tests work, and code looks good

PR #398, opened 2023-02-10
[Issue #395] feat: show addresses paybuttons on wallet creation

chedieck:
Description:
Also displays the buttons to which the addresses are related, upon creating a wallet.

Test plan:
Try to create a wallet, there should appears the addresses with their related buttons.

Depends on:
- [ ] #396

PR #399, opened 2023-02-10
[Issue #395] feat: remodel address to wallet on wallet card

chedieck:
Description:
Shows the paybuttons that belongs to a wallet by inferring them from the wallet addresses.

Test plan:
Go on to create a new wallet, select some address to it. See if all wallets now are correctly inferring their paybuttons.

Remarks:
- Now, editing a wallet will produce unexpected results. This will be later addressed in another PR.

Depends on:
- [x] #398

### DIFF
--- a/components/Wallet/EditWalletForm.tsx
+++ b/components/Wallet/EditWalletForm.tsx
@@ -7,12 +7,12 @@ import Image from 'next/image'
 import style from '../Wallet/wallet.module.css'
 import style_pb from '../Paybutton/paybutton.module.css'
 import EditIcon from 'assets/edit-icon.png'
-import { WalletWithAddressesAndPaybuttons } from 'services/walletService'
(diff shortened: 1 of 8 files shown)


Klakurka:
I'm not sure how to test this as they're all greyed out.

chedieck:
> I'm not sure how to test this as they're all greyed out.
You can add new ones or reset the database, which should bring it to a state where not all will be greyed out

Klakurka:
Maybe we should indicate that an address is in use in XYZ wallet?

ScottMcDermid (approved):
Looks good

Klakurka:
@chedieck Actually can you explain again why all of the addresses there are unavailable to be used in the new wallet I was trying to make?

Klakurka (approved):
We're going to allow empty wallets.

PR #410, opened 2023-02-28
Chore: update prisma to 4.10

chedieck:
Description:
Updates prisma to version 4.10, that brings cool new functionalities (as being able of making queries by comparing fields).

Test plan:
N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "mysql2": "^2.3.3",
     "next": "latest",
     "node-mocks-http": "^1.11.0",
-    "prisma": "^3.14.0",
+    "prisma": "^4.10.1",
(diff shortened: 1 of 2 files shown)


vorg-code:
there is a library not found issue in the integration test run

chedieck:
After updating the '@prisma/client' module & fixing the necessary changes, tests would hang. I tried to create a global teardown on jest to close the prisma connection to the DB, but didn't succeed, be it with a javascript or a typescript teardown file. Therefore, I am closing this to conclude the address -> wallet remodel that is more important right now; I'll just don't do the queries that I could OR use raw queries; and create an issue to better it in the future

PR #414, opened 2023-03-01
[Issue #395] feat: remodel address to wallet on wallet update

chedieck:
Description:
Part of the #395 remodel, brings the changes to the edit wallet view.

Test plan:
Updating a wallet from the front end should work normally.


Klakurka (changes requested):
- On the 'My Wallet' wallet, it allows you to uncheck the 'Make default wallet' checkbox but it doesn't apply since it's not allowed. Maybe best to just grey it out so that they can't uncheck it?
- I think we can remove the word 'Make ' from 'Make Default BCH/XEC Wallet'.

chedieck:
> * On the 'My Wallet' wallet, it allows you to uncheck the 'Make default wallet' checkbox but it doesn't apply since it's not allowed. Maybe best to just grey it out so that they can't uncheck it?
> 
>     * I think we can remove the word 'Make ' from 'Make Default BCH/XEC Wallet'.
> 
>     * When creating a new wallet, the 'Make Default BCH/XEC Wallet' checkboxes don't work (same with editing a wallet). This is at least the case when there's no paybuttons added to the account.

I will implement all of these as part of #415 , but I think they are out of scope on this PR

Klakurka (approved):

PR #416, opened 2023-03-03
Feat/411 1 chronik get balance

vorg-code:
Steps to test:

1. run (changes in package.json version of libraries): make reset-dev
2. open in you browser and see balance: http://localhost:3000/api/address/balance/ecash:prezk74rzl4uw22wm6agcteuwztdvp5a9umns6k0d3
3. Change in constants/index.ts the ecash choice to chronik:
`export const BLOCKCHAIN_CLIENTS_CHOSEN: KeyValueT<number> = {
  ecash: BLOCKCHAIN_CLIENTS_OPTIONS.chronik,
  bitcoincash: BLOCKCHAIN_CLIENTS_OPTIONS.grpc
}`
4. open in you browser and see balance (now from chronik client): http://localhost:3000/api/address/balance/ecash:prezk74rzl4uw22wm6agcteuwztdvp5a9umns6k0d3

Other API endpoints will not work with chronik as they are not implemented yet, this will be addressed in future PRs.


Klakurka:
Integration test failing.

Klakurka (changes requested):
- The site nav UI is a bit weird. It has a scrollbar for me and also a div border and some other differences. Maybe @johnkuney can look at that? This was almost certainly due to updating the packages
- On the Networks page, eCash isn't connected for me. I might be doing something wrong.
- The Dashboard page doesn't load for me (likely related to the eCash network connection issue).
- Logout doesn't work for me.
- This error shows up in the console: 
image

constants/index.ts
```diff
@@ -98,3 +100,9 @@ export const NETWORK_TICKERS: KeyValueT<string> = {
 
 export const NETWORK_IDS: KeyValueT<number> = { XEC: 1, BCH: 2 }
 export const QUOTE_IDS: KeyValueT<number> = { USD: 1, CAD: 2 }
+
+export enum BLOCKCHAIN_CLIENTS_OPTIONS {'grpc', 'chronik'}
```
chedieck, line 104: It's best to avoid enums, not a really good TS feature
[Here some reasons why](https://fettblog.eu/tidy-typescript-avoid-enums/)

vorg-code: nice read, addressed

vorg-code:
@chedieck done.

@Klakurka 

1. on the network page XEC shows as not connected because this line fails:

`return (await client.getBlockInfo({ index: height })).toObject()`
image

which is weird because the previous method `getBlockchainInfo` executes normally  and also the `utxos` to get the balance. Also the method `getBlockInfo` runs properly for BCH, that is why the green is there, so maybe it is an issue (on this `getBlockInfo` method only) with the API implementation present in the XEC url that we are using.

2. Dashboard opens up normally here:
image

3. Logout fixed

4. Console error fixed

Klakurka (changes requested):
Hmmm I'm still not able to get the dashboard to load. In the console there are these errors:

image

vorg-code:
I am closing this PR since it was split between PRs: #417 and #418.

@Klakurka this last message of yours, let's see if it happens in #418 , maybe it is some version of a package that your environment didn't get correctly, you can try to delete the node_modules folder and then run `npm i` to install all packages again and reply in #418 if that doesn't help.

Issue #370, opened 2023-01-24
Improve sycing of address with many txs


chedieck:
Thanks Scott, I was able to reproduce the error here.

The problem is that there are over 700k txs for the second address, so the current implementation goes on to sync them 100 at a time, and it just never finished.

I though about some some possible improvements to this, which I could address in some subsequent PRs (because this issue is  independent of this PR):

1. Progressively commit the tx syncing DB transactions:
   - Transactions would be added, their prices synced, e.g. 1000 at a time (for what I tested there is a byte limit when fetching this data, and for this address it got violated after ~5000 txs at once) and this would already show up in the frontend. It will still take a long time to sync all of them, but we would have the progress feedback.

2. Use another API to get price data
   - The coin.dance API we are using right now only allows to get the price of ONE day a time, which considerably slows the process of getting all txs prices.

3. Have a limit on how many txs to sync, until more is asked
   - Example: Fetch 1000 txs for some address. Only when the user, through the frontend, goes to e.g. page 50 of the tx list and expects to see older transactions do we then fetch the next 1000, and so on.

Any combination of the above is possible, so I am opening another issue  for we to discuss these solutions and also other ideas you guys might have.

_Originally posted by @chedieck in https://github.com/PayButton/paybutton-server/issues/350#issuecomment-1402423529_

vorg-code:
For the second point, we could simply download just once the last 4 years of prices, it is a small dataset to have in a file to initialize the database upon first build, since it will be 4 years * 365 days * 2 coins = less than 3000, it takes a long time now because it is one request per day that we don't have in the database, if it is all in the database, no HTTTP requests will be necessary (unless it is earlier than the 4 years, which must be super rare, we can also increase to 7 years anyway and almost eliminate this api request, leave it just as a safe fallback).

To keep this solution top notch performance-wise, I think it would be nice to update this file from time to time with recent data and also create a job that pre-requests all days missing until today on startup, so that we don't need to do this when the transactions are coming in.

@Klakurka @chedieck what do you think? If you like this solution, I can start working in a PR specifically for this point 2 now and later we discuss points 1 and 3.

chedieck:
> @Klakurka @chedieck what do you think? If you like this solution, I can start working in a PR specifically for this point 2 now and later we discuss points 1 and 3.

[...]

PR #417, opened 2023-03-04
Feat/411-1 updated versions and new packages

vorg-code:
Next version fixed in the latest 13, react upgraded to 18, new chronik packages added. This PR shouldn't change anything in the functionality of the whole system. 

To test, run (changes in package.json version of libraries): `make reset-dev` and navigate the pages to see if everything works as before.


Klakurka (changes requested):
The UI (nav) changes with this one, as expected.

How about letting @johnkuney fix it up before we merge this?

vorg-code:
@Klakurka sure thing! @johnkuney feel free to hop in. 

Also, @chedieck , as you can see, here the coin dance api is unreachable as well. Any solution besides git unignoring the CSV file?

chedieck:
> Also, @chedieck , as you can see, here the coin dance api is unreachable as well. Any solution besides git unignoring the CSV file?

Git unignoring the CSV file is not a real solution, but a workaround — you are not fixing the problem of github actions not being able to fetch the prices, but making it not having to do so  — and as far as I can see it does not even fixes that: if the commited csv file is one day behind, the same error would happen when trying to sync this day. Also commiting this file brings considerable issues that I already overextended myself in explaining at #370.

You can turn this into a draft PR and remove stuff/ add console.logs to try to debug better what is the change that made it stop working. It's not a problem in the Github end, because this tests are still working fine on master, as shown by [these tests](https://github.com/PayButton/paybutton-server/actions/runs/4336180132).

chedieck:
> You can turn this into a draft PR and remove stuff/ add console.logs to try to debug better what is the change that made it stop working. It's not a problem in the Github end, because this tests are still working fine on master, as shown by [these tests](https://github.com/PayButton/paybutton-server/actions/runs/4336180132).

I was working at another branch and noticed that the actual current behavior is that the seed will fail as it's failing here, but the tests for some reason I don't understand go on and finish successfully. AFAIK, we don't really need these prices on the integration tests, so I think we could add an if clause to the `seed.ts` file to ignore this part if `env === TEST`, what do you think?

vorg-code:
@chedieck  Yeah I have just noticed that github can't get the prices in your linked tests but they aren't marked as failed tests:
image

I will try to understand why that happens, maybe we can get the same behavior going in these new branches. If this investigation yields nothing useful, then I think your `if` suggestion is a nice one.

PS: just for clarification, if the file was not git ignored, we wouldn't fetch anything from coindance even if many days elapsed since the file was created, because those missing days are fetched by the worker and not at build time, the file once created is never modified again.

vorg-code:
There is a way of ignoring errors in the first part of the yarn command of integration tests but that would mask any errors together with this api being unreachable issue, so it would be a bad solution. I will add an alternate flow that provides some random prices (like we had before #370 ) if `env === TEST`.

vorg-code:
Done.

prisma/seeds/prices.ts
```diff
@@ -14,6 +14,45 @@ interface PriceFileData extends KeyValueT<string> {
   priceInUSD: string
 }
 
+export const mockPrices: Price[] = [
+  { // XECUSD
+    id: 0,
```
chedieck, line 19: All of them having the same if shouldn't result in an error? I see in the test that it passed fine, just trying to understand.

prisma/seeds/prices.ts
```diff
@@ -14,6 +14,45 @@ interface PriceFileData extends KeyValueT<string> {
   priceInUSD: string
 }
 
+export const mockPrices: Price[] = [
```
chedieck, line 17: This looks like the objects in `tests/mockedObjects.ts`, so I think it would better go there and then you can import it here

chedieck:
Left some minor comments, other than that LGTM

> PS: just for clarification, if the file was not git ignored, we wouldn't fetch anything from coindance even if many days elapsed since the file was created, because those missing days are fetched by the worker and not at build time, the file once created is never modified again.

I see, makes sense.

johnkuney:
Okay added some css tweaks to get the menu back to where it was. Let me know if any other display issues

prisma/seeds/prices.ts
vorg-code: done

prisma/seeds/prices.ts
vorg-code: since the ID is defined as:
`id           Int                    @id @default(autoincrement())`

I believe it uses autoincrement instead of the value (zero) passed in there, effectively ignoring it (that is an educated guess, this doubt also hit me earlier). Since it is a required field, I had to pass a number, zero seemed the best one for clarity. I also input zero to all of them when reading the file and inserting thousands into the database.

vorg-code:
Everything mentioned so far was addressed.

chedieck (approved):

Klakurka (approved):

PR #421, opened 2023-03-05
[Issue #395] refactor: remove deprecated WalletWithAddressesAndPaybuttons type

chedieck:
Description:
Part of #395. Changing the type that includes a wallet associated paybuttons (since we are deleting this association) for the type that gets the paybuttons from the wallet addresses.

Test plan:
Only a refactor, nothing should change.


Klakurka (approved):
I ran into several different issues and errors but eventually I got it working after re-checking it out coming from master.

Issue #264, opened 2022-10-07
Feature: UUID Primary Keys


ScottMcDermid:
Replace incrementing ID's with UUID's.

Pros:
- Prevents attackers from predicting incrementing ID's
- Allows services to generate ID's without verifying with the db
- Opens the door for distributed architecture
- Auto-Incrementing ID's can cause deadlocks between 2 transactions attempting to insert into the same table. UUID's avoid this.

Cons:
- Slower insertion, unless we store them in an ordered format. I'm not sure if Prisma easily allows this: https://www.percona.com/blog/2014/12/19/store-uuid-optimized-way/

PR #424, opened 2023-03-06
chore: create seed to dev2 user default wallet

chedieck:
**Description:**
We already have logic to create a default wallet for every user, upon signing up. The `dev2@paybutton.org` test user, however, did not had any wallets, which represents what should be an invalid state.
Adds a default wallet to `dev2`.

**Test plan:**
After resetting the database (with `make reset-dev` or simply `yarn docker mr`), this is what should appear after logging into `dev2@paybutton.org` (password `dev`):

image

**Depends on:**
- [ ] #423


Klakurka (approved):

Klakurka (approved):

Klakurka:
Scott was talking about using UUIDs which it seems like we should be doing as right now we're using incrementing ids and those are being passed around in the client network traffic.

https://github.com/PayButton/paybutton-server/issues/264

PR #428, opened 2023-03-07
Feat/411 3 get blockchain info get block info

vorg-code:
This one should be reviewed after #418 

Description:

This PR intends to test the new chronik blockchain client implementation of methods `getBlockchainInfo` and `getBlockInfo`.

Steps to test:

1. open in you browser the networks page: http://localhost:3000/networks (ecash will not show connected because grpc getBlockInfo method seems to be broken for ecash network)
2. Change in constants/index.ts the ecash choice to chronik: `export const NETWORK_BLOCKCHAIN_CLIENTS: KeyValueT<BLOCKCHAIN_CLIENT_OPTIONS> = {   ecash: 'chronik',   bitcoincash: 'grpc' }`
3. Reload the networks page and see ecash status connected


vorg-code:
I will close this PR and try to open another one without commits that already are on master.

vorg-code:
Reopening, as I thought the new chronik changes (#418) were already on master branch, but they are not, that is why they are also part of this PR and also why it is so big. The actual code changes done here after #418 are actually small.

Klakurka:
Conflicts

vorg-code:
@Klakurka conflicts solved

Klakurka:
I tried the steps but couldn't get the network page to show eCash as connected. I'll try on another PC later.

vorg-code:
Tried here again and it is working, also, I noted that grpc's block info is working again also, so both clients are showing as connected to me now.

Klakurka:
My const is different than that. I have: `NETWORK_BLOCKCHAIN_CLIENTS`, not `BLOCKCHAIN_CLIENTS_CHOSEN`.

services/blockchainService.ts
```diff
@@ -20,12 +18,21 @@ export interface GetAddressParameters {
   reversedHashOrder?: boolean
 }
 
+export interface BlockchainInfoData {
```
chedieck, line 21: I think we could remove the `Data` from the name here and on the `BlockInfoData`, I think it is redundant with the `Info` part and it also goes together better with the method names: `getBlockInfo`, `getBlockchainInfo`

services/chronikService.ts
```diff
@@ -1,23 +1,33 @@
...
-    throw new Error('Method not implemented.')
+  async getBlockchainInfo (networkSlug: string): Promise<BlockchainInfoData> {
+    this.validateNetwork(networkSlug)
+    const blockchainInfo = await this.chronik.blockchainInfo()
+    return { height: blockchainInfo.tipHeight, hash: blockchainInfo.tipHash }
+  }
+
+  async getBlockInfo (networkSlug: string, height: number): Promise<BlockInfoData> {
+    this.validateNetwork(networkSlug)
+    const blockInfo = await (await this.chronik.block(height)).blockInfo
```
chedieck, line 29: this first `await` is useless

chedieck (changes requested):
Left some remarks on the code.

Also, `'chronik'` as the network blockchain client for XEC is part of the steps to test, and not part of the diff, because this PR does not yet solves it for anywhere else besides the `/networks` view, right?

It would be better if you included a "Description:" section before the test plan so we know better what the PR is about and not have to guess it from the changes

services/blockchainService.ts
vorg-code: true, addressed

services/chronikService.ts
vorg-code: nice catch

vorg-code:
@Klakurka you have the correct and most updated variable name `NETWORK_BLOCKCHAIN_CLIENTS`, I altered in the PR previous to this one, merged master to this one and forgot to update this PR text description, but now I have done it.

@chedieck indeed, `Description:` added in this PR (and I will also add it to the next ones as well). Code changes addressed.

Klakurka (dismissed):

services/blockchainService.ts
chedieck: `BlockInfoData` still has the `Data` ending

vorg-code: fixed

chedieck (approved):

Issue #426, opened 2023-03-06
Two users can't assing the same address two different wallets


chedieck:
Now that #395 is close to completion, I've noticed that if two different users were to create one paybutton each, but using the same address, then this address would only be able to belong to a wallet of one of the users.

Plan of solving:
- Create a new table relating `Address` to `UserProfiles`,
- Have a column `walletId` on this table,
- Now `Wallet` has a related field `userAddresses`
- Use this field instead of the field `addresses` everywhere across the code base (#430 solves up to here)
- Delete field `addresses`

PR #430, opened 2023-03-07
2  [Issue #395] feat: use user addresses on wallets

chedieck:
**Description:**
Using the new field `wallet.userAddresses` everywhere `wallet.addresses` was previously used.
Still does not solves  #426. Users are able to create add buttons with a shared address to their wallets; but they are gonna see both buttons (theirs and the one of the other user).

**Test plan:**
- Try to create a paybutton assigning it to wallet;
- Try to edit some wallet paybuttons;
- Try to create a new wallet with paybuttons;

All should work, with the aforementioned caveat.

Depends on:
- [x] #429 
- [x] #434


Klakurka (changes requested):
When I click on a button to view/edit, I get this:

image

chedieck:
> When I click on a button to view/edit, I get this

This is happening on master too. I investigated and looks like it has to do with the react update, I'll fix it in another PR

Klakurka (approved):

Issue #411, opened 2023-02-28
Switch XEC from BCHD to Chronik


Klakurka:
We can build against https://chronik.fabien.cash and then run our own instance eventually.

Docs: https://chronik.be.cash/docs/

vorg-code:
Let's implement this migration in parts, the first PR will create the Chronik service, install dependencies and implement the getBalance method for a given address replacing the use of grpc by this new one across the project.

Klakurka:
Some useful links:

- Example getAllTxHistory function: https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/web/alias-server/chronik.js#L120
- Example getAllTxHistory function: https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/web/cashtab/src/utils/chronik.js#L104
- Example parseChronikTx function: https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/web/cashtab/src/utils/chronik.js#L839
- Example outputScriptToAddress function: https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/web/cashtab/src/utils/cashMethods.js#L1248
- Example Chronik output: https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/web/alias-server/test/mocks/aliasMocks.js#L3

Klakurka:
Still more to do on this one.

Issue #403, opened 2023-02-14
improve transaction sync


vorg-code:
Remaining to do points of original's issue #370 message:

> The problem is that there are over 700k txs for the second address, so the current implementation goes on to sync them 100 at a time, and it just never finished. Some possible improvements:

> 1. Progressively commit the tx syncing DB transactions:
>    - Transactions would be added, their prices synced, e.g. 1000 at a time (for what I tested there is a byte limit when fetching this data, and for this address it got violated after ~5000 txs at once) and this would already show up in the frontend. It will still take a long time to sync all of them, but we would have the progress feedback.

> 3. Have a limit on how many txs to sync, until more is asked
>    - Example: Fetch 1000 txs for some address. Only when the user, through the frontend, goes to e.g. page 50 of the tx list and expects to see older transactions do we then fetch the next 1000, and so on.

vorg-code:
Improvement idea to implement on this one:
 - if it has already synced all transactions at least one time, then the first confirmed transaction it tries to insert that is already there, we abort the sync and return, because the past is immutable

The rest of the improvements I had in mind were already implemented in #451

vorg-code:
I think it would be a good idea to have a flag to indicate that we are downloading that address and a job that cleans that flag for every address where no new transactions were added in the last 15 seconds or something like this, so that we know this flag can be trusted no matter if some process got stuck or finished throwing an error that wasn't treated and so the flag was left on and, if that flag is on, we don't sync on new requests

vorg-code:
Another thing to remember when implementing this is that the subscribe transaction job can bring new transactions before the full transaction sync runs, then we would find a transaction already in the database but the past transactions would not be there, so we need another way to know the date of the transaction synced by the full sync and not by the subscribe transactions.

PR #431, opened 2023-03-11
Feat/411 4 get address transactions

vorg-code:
Testing steps:

1. run against the DB: `select * from paybutton.Transaction` (maybe also `delete from paybutton.Transaction` to clean this table)
2. open this url to sync transactions (with any bch or ecash address): http://localhost:3000/api/address/transactions/sync/ecash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7quss2vxek
3. Select again on Step 1 above and see the transactions imported there

This PR already changes in master the choice of the blockchain client from grpc to chronik for both BCH and XEC since most of chronik is already implemented and transactions had a big impact on code, so I had to remove grpc code on transactions, it no longer works.

Also I had to change some tests since the return type of some functions related to transactions changed, so I went on fixing whatever was broken by these changes, midway I realized this PR would be huge if I went on that path and I just commented out all tests that used chronik features, this will be addressed on the final #411 PR.


Klakurka (commented):
It works but I get the following response in the browser on the URL:

`{"statusCode":500,"message":"14 UNAVAILABLE: DNS resolution failed"}`

services/blockchainService.ts
chedieck, line 1: I think the changes in this file misses the point of having a generic `blockchainService` file. The way it is right now only works with chronik, so it makes the whole parsing of client and generalization useless

Take former `getAddress` now `getAddressTransactions` (which is a better name for sure), for example. To actually take advantage of the generalization we should do something like:

```
getAddressTransactions(params: GetAddressTransactionsInput): GetAddressTransactionsOutput {
  if client_for_address_is_chronik:
    chronikGetAddressTransactionsInput =  parseChronikGetAddressTransactionsInput(params)
    chronikResponse = chronik.getAddressTransactions({...chronikGetAddressTransactionsInput})
    return parseTxHistoryPage(chronikResponse)

  else if client_for_address_is_grpc:
    grpcGetAddressInput =  parseGrpcGetAddressInput(params)
    grpcResponse = oldImplementation(...)
    return parseGetAddressTransactionsResponse(grpcResponse)

  else:
    throw new Error("Not implemented for client...")
}
```

Ofc this assumes the creation of the parse functions and the input & output interfaces.

chedieck (changes requested):

Klakurka:
Failed test.

vorg-code:
@chedieck @Klakurka now blockchain implementations are much more decoupled from the rest of the code, making interactions between both sides simpler and more objective. Also I fixed chronic for ecash and grpc for bch, this should stay like this and both clients valid and fully functional until all the migration PRs or #411 are finished.

tests/mockBlockchainClient.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  public set checkNetworkIntegrity (value) {
+    this._checkNetworkIntegrity = value
+  }
+
+  getAddressTransactions (_: any): Tx[] {
+    return mockedBlockchainTransactions
+  }
+
+  // getAddressUtxos (_: object): GetAddressUnspentOutputsResponse {
```
Klakurka, line 19: Commented out tests on this branch?

Klakurka (changes requested):

Klakurka:
Also I get this when trying to view a button:

image

tests/mockBlockchainClient.ts
vorg-code: Yeah, all blockchain tests have been commented out and they will be refactored in the end with one tests only PR.

vorg-code:
@Klakurka the new PRs by @chedieck weren't in this branch yet, merged master into here, it should be working now (tested locally here and could open it).

tests/mockBlockchainClient.ts
chedieck: Font't use multiple `//`, use one simple `/*  */` block and make sure to have some `// WIP` commented in the top of all of these blocks so we don't forget about fixing it

chedieck: Also, you don't have to comment tests, you can add `.skip` to them so they don't run (but put the `// WIP` anyways ofc)

services/blockchainService.ts
```diff
@@ -27,12 +18,23 @@ export interface BlockInfo extends BlockchainInfo {
   timestamp: number
 }
 
+export interface TransfersResponse {
```
chedieck, line 21: Why call those `Transfer` where everywhere else this is called `Transaction`?

Ofc `Transaction` already exists as a `Prisma`  generated type, but this shouldn't be `Transaction` either since it wouldn't be clear that it refers to the `blockchainService` expected inputs or outputs, you could just name the single one `TransactionResponse` and this one `TransactionsResponse`

chedieck: In general, I just think it's a bad practice to take two words that are kind of synonyms and make such a convention that the one means X, the other one means Y, since then the name would not help anybody who reads the code and tries to understand what the structures are.

chedieck (changes requested):

services/blockchainService.ts
vorg-code: Maybe we can find a better name, however a blockchain transaction is a concept where we have possibly many inputs and many outputs, that is not the data structure we have here, here we have just how much money is going in or out of one address and when, similar to one `Transfer`, came to mind. Maybe `Credit`? Anyway, I feel that calling this simple data structure a `Transaction` is conceptually wrong.

Note that this is an interpretation of the blockchain data done after some data crunching and transforming,  we can (and did) generate a new structure and data out of it.

tests/mockBlockchainClient.ts
vorg-code: I have to comment anyway because I changed almost all return types, so the code doesn't even compile, that being said, changed single line comment to multiple and added WIP.

services/blockchainService.ts
chedieck: Yes, you have a good point. Then the DB table should also be called `Transfer` instead of `Transaction`, right?
Actually, now I am asking why don't we use the prisma generated type here? In the end the idea is put those in that table anyways. Then what you called `txid` would be `hash` and `receivedAmount` would be just `amount`.

I can see the difference now and agree that `Transfer` is indeed a better name (it even helps separate `prisma.transaction` from `prisma.$transaction`), as `txid` and `receivedAmount` also are better, clearer than `hash` and `amount`. So I actually think it would be better to make a refactor in the future, change the name of the prisma table & fields.

But a more important question for now, regarding using the prisma generated types here (now or after the refactor): shouldn't we add an `addressId` and a `confirmed` fields to this type? Then it would be fully compatible with the prisma types.

chedieck (changes requested):

tests/mockBlockchainClient.ts
chedieck: Got it, but this is still missing here and on `tests/integration-tests/api.test.ts`

vorg-code: oh indeed, fixed

Klakurka (approved):

services/blockchainService.ts
chedieck: can you add the other two field tho?

vorg-code: Done. I added the Address instead of number addressId since in PR 411-6 I already need an Address in there.

vorg-code:
@chedieck done

tests/mockBlockchainClient.ts
```diff
@@ -16,16 +12,11 @@ export default class MockGrpcClient {
...
-      res.setConfirmedTransactionsList([])
-    } else {
-      res.setConfirmedTransactionsList([mockedGrpc.transaction1, mockedGrpc.transaction2])
-    }
-    return res
+  getAddressTransactions (_: any): Tx[] {
+    return mockedBlockchainTransactions
   }
 
+  /*
```
chedieck, line 19: Missing WIP

services/transactionService.ts
```diff
@@ -165,43 +115,23 @@ export async function upsertManyTransactionsForAddress (transactions: Transactio
...
-
-    let newInsertedTransactions = await upsertManyTransactionsForAddress(nextConfirmedTransactions, address)
-    newInsertedTransactions = [
-      ...newInsertedTransactions,
-      ...(await upsertManyTransactionsForAddress(nextUnconfirmedTransactions, address, false))
-    ]
-    insertedTransactions = [...insertedTransactions, ...newInsertedTransactions]
-    await new Promise(resolve => setTimeout(resolve, FETCH_DELAY))
-  }
+  const addressTransfers = await getAddressTransfers(addressString)
```
chedieck, line 119: I don't like this refactor at all:
1. You are defining now this `getAddressTransfers` in two different places:  `grpcService` and `chronikService`, such that most of it is duplicated code;
2. But the bigger problem is that: notice that before the function would go into upserting transactions as they were fetched from the thrid party; for each fetched page we would insert the transactions into the database before getting the next `FETCH_N` transactions. Now it's just going to await until it has fetched ALL the transactions, page by page, and then it will upsert into the database. There are addresses with 20k+ transactions which will therefore never complete or sync anything. This also makes impossible to implement #403 point 3.

.eslintrc.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
   ignorePatterns: ["node_modules/"],
   rules: {
     '@typescript-eslint/no-non-null-assertion': 'off',
-    '@typescript-eslint/triple-slash-reference': 'off'
+    '@typescript-eslint/triple-slash-reference': 'off',
+		"no-tabs": 0,
```
chedieck, line 10: 1. Bad indentation
2. Why is this in this PR? Why allow tabs?

services/blockchainService.ts
chedieck: I actually still can't see the answer for:
> Actually, now I am asking why don't we use the prisma generated type here?

The reasons I see is that the name `Transfer` is better as are the field names, but since all the codebase already uses the prisma `Transaction` table, it doesn't seem worth to me to create a virtually identical type. In some future refactor we can and should change the name of the Transaction table & its fields, but for now you wouldn't even have to define an extra interface here; simply use the one already provided by prisma

tests/setupMocks.ts
```diff
@@ -1,7 +1,7 @@
-import MockGrpcClient from './mockGrpcClient'
+import MockChronikClient from './mockBlockchainClient'
 
-jest.mock('grpc-bchrpc-node', () => ({
-  ...jest.requireActual('grpc-bchrpc-node'),
+jest.mock('chronik-client', () => ({
```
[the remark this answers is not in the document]
vorg-code: Indeed, this removal was part of the beginning of this PR where I thought grpc would be removed, then I didn't think again about it since tests were to be addressed in the last PR of its own, but I addressed this by creating a new file for each client to be tested (just skeleton)

tests/mockBlockchainClient.ts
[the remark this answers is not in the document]
vorg-code: addressed in the comment above

tests/integration-tests/api.test.ts
```diff
@@ -1188,6 +1188,7 @@ describe('GET /api/address/transactions/[address]', () => {
   })
 })
 
+/*
```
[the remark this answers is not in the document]
vorg-code: done

.eslintrc.js
vorg-code: this was a temporary lint-get-off-of-my-back thing that I forgot to remove, done

services/transactionService.ts
vorg-code: I will consider that when creating the smaller incremental PRs

vorg-code:
I am closing this PR in favor of future smaller PRs.

PR #435, opened 2023-03-13
Feat/411 5 get transaction details and get utxos

vorg-code:
Requires firstly that #431 be merged.

Description:

Implements chronik blockchain client methods: `getTransactionDetails` and `getUtxos`

Test plan:

1. To test `getUtxos`, you can test the `getBalance` endpoint, since it uses the `getUtxos` method. Open, for example: http://localhost:3000/api/address/balance/ecash:qp0kfg0uk64k4pa7tw226djx4v30attzzslrec0ypu
2. To test `getTransactionDetails`, you need any transaction id (you can get it on the block explorer) and then open, for example: http://localhost:3000/api/transaction/bd56e6aaa8aa71e5239ec89fef8ddb7b91e9cf5a321102232967046f94ee0be9?networkSlug=ecash


Klakurka:
Probably going to need to dig into the client to see what we need here.

vorg-code:
Closing this one as it will be addressed after the stream of new smaller PRs originated from #431. Maybe after those, this one will also be broken into smaller ones.

Issue #401, opened 2023-02-11
Allow deleting wallets


Klakurka:
- Don't allow deleting any default wallet.
- If you delete a wallet, move all of the addresses/buttons to the default wallet(s).

chedieck:
I'm closing this because it seems to have already been implemented with the mentioned requirements

PR #440, opened 2023-03-15
[Issue #401] feat: add button & modal to delete wallets

chedieck:
**Description:**
Solves #401.

**Test plan:**
Go to http://localhost:3000/wallets, create a wallet, assign some addresses to it. Then go on and delete it. The assigned addresses should be reassigned to the default wallets (XEC addresses to the XEC default and BCH addresses to the BCH default).

The delete button should not appear for default wallets (even if it did, the deletion would be blocked on the back end).

### DIFF
--- a/components/Wallet/EditWalletForm.tsx
+++ b/components/Wallet/EditWalletForm.tsx
@@ -5,6 +5,7 @@ import Image from 'next/image'
 import style from '../Wallet/wallet.module.css'
 import style_pb from '../Paybutton/paybutton.module.css'
 import EditIcon from 'assets/edit-icon.png'
+import TrashIcon from 'assets/trash-icon.png'
 import { WalletWithAddressesWithPaybuttons } from 'services/walletService'
 import { AddressWithPaybuttons } from 'services/addressService'
 import axios from 'axios'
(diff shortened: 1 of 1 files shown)


components/Wallet/EditWalletForm.tsx
```diff
@@ -128,16 +137,41 @@ export default function EditWalletForm ({ wallet, userAddresses, refreshWalletLi
                     </div>
                   </div>
 
-                  <div className={style_pb.btn_row}>
+                  <div className={style_pb.btn_row2}>
                     {error !== '' && <div className={style_pb.error_message}>{error}</div>}
-                    <button type='submit'>Submit</button>
-                    <button onClick={() => { setModal(false); reset() }} className={style_pb.cancel_btn}>Cancel</button>
+                    {wallet.userProfile === null || (wallet.userProfile.isXECDefault === true ?? wallet.userProfile.isBCHDefault === true)
```
johnkuney, line 142: Think another || here instead of ?? will fix the delete button showing for default bch wallets

johnkuney (changes requested):
Looks good and worked as expected for me. Only thing I noticed was the delete button was still there for the default BCH wallet. Should be fixable with that comment though

Klakurka (changes requested):
If you have a wallet with an XEC address and it's set to be just the default BCH wallet, the delete wallet button still shows up and then an error is thrown when trying to use it. The button should just not be there.

image
image

components/Wallet/EditWalletForm.tsx
chedieck: good catch, solved

chedieck:
@Klakurka should be fixed now

johnkuney (approved):
Cool looks good to me

Klakurka (approved):

PR #446, opened 2023-03-19
Feat/411 6 subscribe transactions

vorg-code:
Description:

Subscription of blockchain transactions, they should be added to our database at the same time as they happen in the blockchain. 

I have tested it for XEC, however I didn't for BCH as I don't have any BCH addresses yet.

How to test:
 
1. add an address you control to the application
2. run `yarn docker c`, then `flushall`, then `exit`
3. run `yarn docker jr`
4. transfer from or to the address in step 1
5. run database `select * from paybutton.Transaction`, you will see your transaction added as unconfirmed fist, then as confirmed

PS: I commented out the sync part (highlighted below) and left only the subscribe part below to test in `workers.ts` (do this if the sync part throws an error, it should be fixed in another PR -> #431 ):

const syncAndSubscribeAddresses = async (addresses: Address[]): Promise<void> => {
`  await Promise.all(
    addresses.map(async (addr) => {
      await transactionService.syncTransactionsAndPricesForAddress(addr.address)
    })
  )`
  await blockchainService.subscribeAddressesAddTransactions(addresses)
}


vorg-code:
Closing this one as it will be addressed after the stream of new smaller PRs originated from #431. Maybe after those, this one will also be broken into smaller ones.

PR #449, opened 2023-03-26
[Issue #264] feat: use  UUID on paybuttons

chedieck:
Description:
Part of #264. Paybuttons already had an `uuid` field, this makes so that it is now the `id` field.

Test plan:
All related to paybuttons should work normally: deleting, creating, modifying, the dashboard...

Nothing should change for the end user except for the fact that the paybutton detail URL shows some UUID instead of some integer.


ScottMcDermid (approved):
Tests work for me, and code looks good!

PR #454, opened 2023-04-03
[Issue #264] fix: create transaction input

chedieck:
Description
---
Some functions that deal with the creation of `Transaction` objects were being expected to receive the type of a `Transaction`, as a row inserted in the database as input. This is not accurate because that type contains an `id` field, that we don't know yet.

Besides that, the `addressId` was set to be a hardcoded 0, which I don't understand why but don't think is correct, since we are going to connect that transaction with some address upon its creation.

Test plan
---
For what I observed, the initial sync of transactions in master is broken again, only for XEC though. Makes this hard to test for improvements right now.

Remarks
---
Associated to # 264 because this is necessary for both the Transaction & Address remodeling of increment number ids to UUIDs.

### DIFF
--- a/jobs/workers.ts
+++ b/jobs/workers.ts
@@ -7,7 +7,7 @@ import * as transactionService from 'services/transactionService'
 import * as priceService from 'services/priceService'
 import * as addressService from 'services/addressService'
 import { GrpcBlockchainClient } from 'services/grpcService'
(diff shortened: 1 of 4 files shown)


Klakurka (approved):

Klakurka:
Sync works for me with this branch although I didn't test master before merging it.

PR #458, opened 2023-04-04
fix: ecashaddrjs type

chedieck:
Description
---
While working on other PR, I was having trouble with the following error:
image

After trying to solve it by many different ways, I decided just to ignore the typing of the `ecashaddrjs` package for now, since the package (suggested by the error) `@types/ecashadddrjs` does not exist

Remarks
---

I also delete a very obsolete file in the `types/` folder and added a github template for the creation of PRs.


chedieck:
Can't do this in a separate PR, or will get another TS lint error

PR #459, opened 2023-04-04
feat: update past newer prices only one request

chedieck:
Description
---
When fetching missing prices in the DB, we were doing several requests, one for each missing day. This changes so that we do only two requests, one for the XEC and other for the BCH prices, then we fill the DB with the missing prices using those lists.

Also, `Promise.all` expects an array of promises, but it was receiving an array of arrays of promises, which mean that it was completing immediatly. So the job was marked as completed before it was actually completed, and the promises would only resolve later. This solves that.

Test plan
---
1. Remove the env variables from  `PRICE_API_TOKEN` from `.env.development.local`, which is sufficient to make the past price sync give errors.
2. Re run the tests with `yd jr` and watch with `yd jw`. In this branch, after the errors one should see:

```
syncing of past prices FAILED
```

... while on master,  after OR before the errors one should see:

```
syncing of past prices finished
```

### DIFF
--- a/services/priceService.ts
+++ b/services/priceService.ts
@@ -139,14 +139,16 @@ export async function syncPastDaysNewerPrices (): Promise<void> {
 
   await Promise.all(
     Object.values(NETWORK_TICKERS).map(async (networkTicker) =>
-      datesToRetrieve.map(async date => {
-        const price = await getPriceForDayAndNetworkTicker(date, networkTicker)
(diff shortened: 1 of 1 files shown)


Klakurka (approved):
Everything seems to work for me.

vorg-code (approved):

vorg-code:
test plan ran as instructed

Issue #248, opened 2022-09-28
Dashboard loads slowly when IFP addresses are loaded into the buttons


Klakurka:
We'll probably need some kind of data summarizing as part of the transaction processing. We don't want the dashboard to have to pull in every single transaction each time.

Klakurka:
In fact for me it doesn't even load at all. Maybe it's just when the IFP address is still syncing?...

Klakurka:
Definitely a good enough experience for v1.

PR #461, opened 2023-04-04
[Issue #248] feat: dedicated redis db to bullmq

chedieck:
Description
---
Create a dedicated redis DB to bullMQ. This way, when caching the dashboard data, they don't get mixed up.

Test plan
---
1. Reset the containers with `make reset-dev`.
2. See if the jobs are being normally executed with `yd jw` or peeking into the `jobs/out.log` file
3. Reset the jobs with `yd jr` and make sure they all run again.

Also, you can enter the redis command-line with `yd c`, then:
1. Running `KEYS *` should return nothing,
2. Running `SELECT 1` and `KEYS *` should return the keys related to bullmq.

Depends on
---
- [x] #459

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ yarn-error.log*
 
 jobs/out.log
 /prisma/seeds/prices.csv
+redis/dump.rdb
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -55,5 +55,5 @@ services:
       - 6379:6379
     volumes:
       - ./redis:/data/redis:z
-    # dump the dataset each 20 seconds if there was at least 1 change
-    command: sh -c "umask 000 && redis-server /data/redis/redis.conf --loglevel warning"
+      - ./scripts:/data/scripts
+    command: sh ./scripts/redis-start.sh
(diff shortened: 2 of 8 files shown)


vorg-code:
Jobs are running and `yd c` has the outputs you described, however I see quite some error messages on `yd jw` output:
image

chedieck:
> Jobs are running and `yd c` has the outputs you described, however I see quite some error messages on `yd jw` output:

I have been getting that too, both here and on master, with not much consistency with regards to if I am gonna get the error or not. Can you check if this is only happening on this branch, for you?

vorg-code (approved):
Problematic messages in master are the same as here, so I am approving this PR.

vorg-code:
@chedieck I just realized that the following file appeared here, not sure if after the run on master or on your branch, should it be git ignored?
image

chedieck:
> @chedieck I just realized that the following file appeared here, not sure if after the run on master or on your branch, should it be git ignored? image

Oh yeah, this is being generated & gitignored in this branch. So when you switched to master, it wasn't gitignored anymore. But you can delete it without worries

Issue #422, opened 2023-03-06
Hide information on networks that are not in use


Klakurka:
There are lots of places where network info is displayed but for now I'm ok limiting this set of changes to the following:

- Hide "Default XYZ Wallet" tag on "Wallets" page when that particular network isn't in use.
- Hide "Default XYZ Wallet" checkboxes in the "Create New Wallet" and "Edit Wallet" dialogs for each network that's not in use.
- Add new "Show Other Networks" button on the Individual network cards on "Networks" page that shows networks that you're not using. Have these networks be hidden unless you click that button after visiting the page.

So if for example you only have eCash in use on your account, you will not see anything related to BCH but you will still see stuff like "Default XEC Wallet".

PR #463, opened 2023-04-04
[Issue #422] feat: get user networks

chedieck:
Description
---
Part of #422. Add a method to see what are the available networks for any user.

Test plan
---
The method is actually not being used anywhere, but it's covered by the tests.

### DIFF
--- a/prisma/seeds/networks.ts
+++ b/prisma/seeds/networks.ts
@@ -1,7 +1,8 @@
-import { NETWORK_SLUGS } from '../../constants/index'
+import { NETWORK_SLUGS, NETWORK_IDS } from '../../constants/index'
 
 export const networks = [
   {
+    id: NETWORK_IDS.XEC,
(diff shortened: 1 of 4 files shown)


prisma/seeds/networks.ts
```diff
@@ -1,35 +1,40 @@
...
   {
+    id: NETWORK_IDS.BCH,
     slug: NETWORK_SLUGS.bitcoincash,
     ticker: 'bch',
     title: 'Bitcoin Cash',
     createdAt: new Date(),
     updatedAt: new Date()
   },
   {
+    id: 3,
```
Klakurka, line 21: Any point in making these UUIDs or better to just leave them?

chedieck: I don't see why, since those `Network` objects are about the project structure, not about user data. They are only inserted /deleted through migrations (actually only inserted)

Klakurka (approved):

PR #464, opened 2023-04-04
[Issue #248] feat: cache logic for dashboard data

chedieck:
Description
---
Part of #248. Implements cache of dashboard data.

Test plan
---
(If not done in #462, need to run `make reset-dev` to reload new redis config)
1. Log into the dev user, see the dashboard. It should take a little while to load. Maybe add some address with a few transactions to make it take a bit longer.
2. After loading for the first time, the payments will be cached and now loading should be faster
3. To erase the cache and try it again, you can go `yd c` and then `FLUSHDB`

Remarks
---
- This only implements loading and creating the cache, not updating it. I will address in a next PR logic to update the payments cache for new transactions.

Depends on
---
- [x] #462


Klakurka (commented):
Adding a button with an address like this one below causes the Dashboard and Payments pages to not load properly for a really long time.

`ecash:qpz9crr5qsvn2lws8jf6x5wxnmklgva7gcx4qsn5sw`

The cache definitely works though. I tried on dev2 and it worked great.

Klakurka (approved):

Issue #432, opened 2023-03-13
Add visual indicator that one or more addresses are still syncing


Klakurka:
After a new button is added, we want to make it clear to people that they're not seeing the latest data yet if it hasn't managed to get processed quickly.

Klakurka:
We have some basic text in place but it's certainly not what we want as a final solution.

Ideas:

1. Add an actual animation / spinner everywhere loading takes place.
2. Differentiate between "hang on while we grab your stuff" and "we've checked and there's nothing to show".

PR #467, opened 2023-04-05
[Issue #432] feat: dedicated component to loading message

chedieck:
Description
---
Adds a simple loading message to the dashboard & button detail views.

Test plan
---
Go to the dashboard and button detail view. Before the data shows up, it should display "Loading...". Maybe add an address with many transactions and delete the cache to make it take longer.

### DIFF
--- a/components/Loading.tsx
+++ b/components/Loading.tsx
@@ -1,0 +1,9 @@
+import React, { FunctionComponent } from 'react'
+interface IProps { text?: string }
+export default ({ text }: IProps): FunctionComponent<IProps> => {
+  return (
+    <div>
+      { text ?? 'Loading...' }
+    </div>
+  )
+}
--- a/pages/button/[id].tsx
+++ b/pages/button/[id].tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import ThirdPartyEmailPassword from 'supertokens-auth-react/recipe/thirdpartyemailpassword'
-import Page from 'components/Page'
+import Loading from 'components/Loading'
 import Router from 'next/router'
 import { PaybuttonDetail } from 'components/Paybutton'
 import { PaybuttonWithAddresses } from 'services/paybuttonService'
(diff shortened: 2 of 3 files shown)


Klakurka (changes requested):
I think maybe we're doing this at the wrong level...

Ideally there should be no page-wide loading screen here.

Right now when I add a button with the IFP address, what happens is very confusing. The button is added initially with 0 txs (the Payments page even says that there are no payments) then after about a minute it loads in a small block of txs. After that, it's another minute before the next batch comes in. This happens repeatedly until all of the txs are loaded in.

The dashboard doesn't detect the new txs yet but we talked about how that's coming in an upcoming branch so that's fine.

So really what we're needing here is:

1. Instant loading of the base pages (really I think it's just the Dashboard which is a problem).
2. An indicator somewhere which communicates that XYZ address(es) are still syncing. Bonus points if we could somehow show a % complete, which might be possible if we can find out the tx count ahead of time on each address.

chedieck:
> I think maybe we're doing this at the wrong level...
> 
> Ideally there should be no page-wide loading screen here.
> 
> Right now when I add a button with the IFP address, what happens is very confusing. The button is added initially with 0 txs (the Payments page even says that there are no payments) then after about a minute it loads in a small block of txs. After that, it's another minute before the next batch comes in. This happens repeatedly until all of the txs are loaded in.
> 
> The dashboard doesn't detect the new txs yet but we talked about how that's coming in an upcoming branch so that's fine.
> 
> So really what we're needing here is:
> 
>     1. Instant loading of the base pages (really I think it's just the Dashboard which is a problem).
> 
>     2. An indicator somewhere which communicates that XYZ address(es) are still syncing. Bonus points if we could somehow show a % complete, which might be possible if we can find out the tx count ahead of time on each address.

Cool, I see what you mean now. I'm closing this, since it doesn't address it.

PR #469, opened 2023-04-07
411_4_2 chronik sync transactions

vorg-code:
Description

Implements blockchain client chronik: sync transactions 

Test plan

1. Create new button with address (http://localhost:3000/buttons): ecash:qp0kfg0uk64k4pa7tw226djx4v30attzzslrec0ypu
2. Change ’grpc’ to ‘chronik’ for XEC in `constants/index.ts`, like this:
  ecash: `'chronik'`,
  bitcoincash: 'grpc'
}
3. Visit: http://localhost:3000/api/address/transactions/sync/ecash:qp0kfg0uk64k4pa7tw226djx4v30attzzslrec0ypu

This will use chronik to sync this address transactions and you will see 5 new transactions if you query the database or open the button you created in step 1.

### DIFF
--- a/services/chronikService.ts
+++ b/services/chronikService.ts
@@ -1,10 +1,15 @@
-import { ChronikClient, ScriptType } from 'chronik-client'
+import { ChronikClient, ScriptType, Tx } from 'chronik-client'
 import { encode, decode } from 'ecashaddrjs'
 import bs58 from 'bs58'
 import { BlockchainClient, BlockchainInfo, BlockInfo, GetAddressTransactionsParameters } from './blockchainService'
 import { GetAddressUnspentOutputsResponse, GetTransactionResponse, Transaction } from 'grpc-bchrpc-node'
-import { NETWORK_SLUGS, RESPONSE_MESSAGES, CHRONIK_CLIENT_URL } from 'constants/index'
-import { TransactionWithAddressAndPrices } from './transactionService'
+import { NETWORK_SLUGS, RESPONSE_MESSAGES, CHRONIK_CLIENT_URL, XEC_TIMESTAMP_THRESHOLD, XEC_NETWORK_ID, BCH_NETWORK_ID, BCH_TIMESTAMP_THRESHOLD, FETCH_DELAY, FETCH_N } from 'constants/index'
+import { TransactionWithAddressAndPrices, upsertManyTransactionsForAddress } from './transactionService'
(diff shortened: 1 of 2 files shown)


services/chronikService.ts
```diff
@@ -59,73 +143,73 @@ export class ChronikBlockchainClient implements BlockchainClient {
...
-    const buffer = Buffer.from(hash160, 'hex')
+  }
 
-    // Because ecashaddrjs only accepts Uint8Array as input type, convert
-    const hash160ArrayBuffer = new ArrayBuffer(buffer.length)
-    const hash160Uint8Array = new Uint8Array(hash160ArrayBuffer)
-    for (let i = 0; i < hash160Uint8Array.length; i += 1) {
-      hash160Uint8Array[i] = buffer[i]
-    }
+  // Test hash160 for correct length
```
Klakurka, line 193: No need for comments that are just English versions of the code. Better to explain why we're bothering to check it.

Klakurka (changes requested):
Test plan works for me but I'm not sure how to sanity check that it's actually using Chronik when I change it over in the constants file.

vorg-code:
@Klakurka you can place a `console.log()` inside the chronik method

services/chronikService.ts
vorg-code: obvious comments removed

Klakurka:
I'm still getting `grpc-bchrpc-node: failed network integrity check` when I switch it to Chronik and my console logs aren't showing up so maybe I need to do a more hard reset of the env?...

vorg-code:
@Klakurka that integrity check happens since forever, we will have to investigate it properly... where did you place the `console.log`?

Klakurka:
All thoughout `chronikservice.ts`.

Klakurka:
I'll leave this up to @chedieck to test properly.

vorg-code:
@Klakurka maybe you forgot to do step 2. 
Also, to see the `console.log` output, after putting it in place in the code, you have to run docker and go into the console of that container, to do so, in the terminal you can type: `docker logs -f paybutton-dev`, then, you can open the sync url in step 3.

Here are both lines you have to alter in the code highlighted and also the output down below of that log in the terminal:
image

Klakurka (approved):
Needed to recompile.

chedieck (approved):
Code looks good, test worked.

PR #471, opened 2023-04-07
[Issue #248] fix: redis flush and jobs repeating

chedieck:
Description
---
Part of #248.
-  Make so that some fail jobs are retried,
-  Removes unnecessary parameter for creating Payments,
- Fixes flushing jobs cache when starting the redis container

Test plan
---
The dashboard should still work normally.

Besides that, to test that the jobs are being rerun when failed:

1. Save the following content into a filed named `temp.patch` on the repo root:
```
diff --git a/jobs/workers.ts b/jobs/workers.ts
index b08a0b4..5acf1c7 100644
--- a/jobs/workers.ts
+++ b/jobs/workers.ts
@@ -40,6 +40,7 @@ const syncAndSubscribeAddressList = async (addressList: Address[]): Promise<void
 
 const syncAllAddressTransactionsForNetworkJob = async (job: Job): Promise<void> => {
   console.log(`job ${job.id as string}: syncing all addresses for network ${job.data.networkId as string}...`)
+  throw new Error("Test")
   try {
     const addresses = await addressService.fetchAllAddressesForNetworkId(job.data.networkId)
     await syncAndSubscribeAddressList(addresses)

```
2. Run `patch -p1 < temp.patch` (this is to make the jobs give an error)
3. Rerun the jobs with `yarn docker jr`
3. Inspect the `jobs/out.log` file. It should give errors when trying to sync the network 1, then another error when syncing network 2. Three times.


Klakurka (changes requested):
Doesn't build for me:

image

chedieck:
> Doesn't build for me
That's weird indeed. I updated with a possible fix. Can you try it again?

Klakurka (changes requested):
It compiles now but I get this when I try to access any of the main pages (that don't seem to load):

image

Klakurka (changes requested):
Working for me now but test is failing.

Klakurka (approved):

PR #472, opened 2023-04-08
Feat/411 5 1 get utxos and get balance

vorg-code:
First review and merge #469, then merge master into this one, code to be reviewed will get a lot smaller.

--- Description:

- Implements chronik: `getUtxos` and `getBalance`
- Improves grpc: `getBalance`
- Makes private: `getUtxos`

The last point takes out `getUtxos` from the `blockchainService` interface which makes sense, normal people would never even want to know that this exists, it is used only to get the address' balance. For that same reason, testing of this method was removed, testing `getBalance` is sufficient to indirectly test this method.

--- Test plan

  ecash: `'chronik'`,
  bitcoincash: 'grpc'
}
2. get balance for a XEC address: http://localhost:3000/api/address/balance/ecash:qp0kfg0uk64k4pa7tw226djx4v30attzzslrec0ypu
3. get balance for a BCH address: http://localhost:3000/api/address/balance/bitcoincash:qzqh7rwaq9zm4zcv40lh9c9u50gy07gcesdmja8426

### DIFF
--- a/services/grpcService.ts
+++ b/services/grpcService.ts
@@ -157,21 +157,15 @@ export class GrpcBlockchainClient implements BlockchainClient {
     return insertedTransactions
   };
 
-  public async getUtxos (address: string): Promise<GetAddressUnspentOutputsResponse.AsObject> {
+  private async getUtxos (address: string): Promise<GetAddressUnspentOutputsResponse.AsObject> {
     const client = this.getClientForAddress(address)
     const res = (await client.getAddressUtxos({ address })).toObject()
     return res
   };
 
   public async getBalance (address: string): Promise<number> {
-    const { outputsList } = await this.getUtxos(address)
-
-    let satoshis: number = 0
-    outputsList.forEach((x) => {
-      satoshis += x.value
-    })
-
-    return satoshis
+    const utxos = (await this.getUtxos(address)).outputsList
+    return utxos.reduce((acc, utxo) => acc + utxo.value, 0)
   };
 
   public async getTransactionDetails (hash: string, networkSlug: string): Promise<GetTransactionResponse.AsObject> {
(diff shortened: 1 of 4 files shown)


services/grpcService.ts
```diff
@@ -129,7 +129,7 @@ export class GrpcBlockchainClient implements BlockchainClient {
 
       if (transactions.confirmedTransactionsList.length === 0 && transactions.unconfirmedTransactionsList.length === 0) break
 
-      // remove transactions older than the networks
+      // remove transactions older than the network
```
Klakurka, line 132: Maybe good opportunity to improve or remove this comment as I have 0 clue what it means.

I'm going to approve the PR anyway since this really isn't making it any less confusing than before.

Klakurka (approved):

services/grpcService.ts
vorg-code: @Klakurka  there is a date in the past where we don't look past it when fetching transactions in the past, as if it were the date where that token/network was born. We just ignore any transaction older than that date.

vorg-code: I can replace that comment with the longer text of my previous comment if that makes it clearer

Klakurka (approved):

services/grpcService.ts
```diff
@@ -157,21 +159,15 @@ export class GrpcBlockchainClient implements BlockchainClient {
...
   public async getBalance (address: string): Promise<number> {
-    const { outputsList } = await this.getUtxos(address)
-
-    let satoshis: number = 0
-    outputsList.forEach((x) => {
-      satoshis += x.value
-    })
-
-    return satoshis
+    const utxos = (await this.getUtxos(address)).outputsList
```
chedieck, line 169: The variable name here should be called `outputsList`

chedieck (changes requested):
LGTM, just a small thing remarked

chedieck (approved):

PR #474, opened 2023-04-12
Feat/411 5 2 get transaction details

vorg-code:
First review and merge #472 , then merge master into this one, clearer to review the code changed here.

--- Description:

Creates a new interface for common data to be answered by blockchain clients implementation of `getTransactionDetails`, changes grpc's implementation to conform (and its test) and implements it in chronik. 

--- Test plan

ecash: `'chronik'`,
bitcoincash: 'grpc'
}
2. get transaction details for a XEC address: http://localhost:3000/api/transaction/4b86d8dbe408fcc095eb2a5e5b7cb72175b46c9de9cac0e113d4897d4c60c1eb?networkSlug=ecash
3. get transaction details for a BCH address: http://localhost:3000/api/transaction/4ea2126520dea06ab6969c65c0d0783ac5216a6b293707d2c99af002f8486fc9?networkSlug=bitcoincash


Klakurka (dismissed):

services/blockchainService.ts
```diff
@@ -24,13 +22,30 @@ export interface GetAddressTransactionsParameters {
   maxTransactionsToReturn: number
 }
 
+interface InputOutput {
+  value: number
```
chedieck, line 26: is this value an integer in 'satoshis'? If not so, I think this would be better stored as a Decimal

services/blockchainService.ts
```diff
@@ -24,13 +22,30 @@ export interface GetAddressTransactionsParameters {
...
+  outputScript?: string
```
chedieck, line 27: another question: I see that rn chronik is using the outputScript and `grpc` the `address`. Isn't there a way to convert the `outputScript` to `address` so we only use one field and make it required?

services/blockchainService.ts
vorg-code: it is indeed, done

services/blockchainService.ts
vorg-code: yes indeed! Damn, much better, done

services/chronikService.ts
```diff
@@ -159,11 +179,12 @@ export function toHash160 (address: string): {type: ScriptType, hash160: string}
...
 
-export function outputScriptToAddress (outputScript: String): string | boolean {
-  // returns P2SH or P2PKH address
+export function outputScriptToAddress (outputScript: String | undefined): string {
+  // returns P2SH or P2PKH address, empty string otherwise
   // P2PKH addresses are in outputScript of type 76a914...88ac
   // P2SH addresses are in outputScript of type a914...87
-  // Return false if cannot determine P2PKH or P2SH address
+
+  if (outputScript == null) return ''
```
chedieck, line 187: some things here:
- If it can be `undefined` why check for `null`?
- what's the reasoning for expecting undefined values and returning the empty string when they exist?

chedieck: In better words: why do we expect inputs or outputs to have no address associated with them? Are these 'mine transactions'? If that's the case, seems better to me to let `address` be optional and explicitly setting it as null instead of as an empty string. If that's only possible for outputs and not for inputs (or vice versa), then I think it would be better to have both interfaces instead of a unified one.

chedieck (changes requested):

services/chronikService.ts
vorg-code: 1. changed verify  `undefined` instead of `null`
2. changed to return `undefined` instead of `''`, indeed the inputs can have no address, I assume being the coins transferred from Coinbase (mined), however I chose to allow undefined in the output's address as well because the `outputScriptToAddress` function can return undefined if address is not a P2PKH or P2SH format. I imagine this could happen, so better be prepared.

services/chronikService.ts
```diff
@@ -154,11 +180,12 @@ export function toHash160 (address: string): {type: ScriptType, hash160: string}
   }
 }
 
-export function outputScriptToAddress (outputScript: String): string | boolean {
-  // returns P2SH or P2PKH address
+export function outputScriptToAddress (outputScript: string | undefined): string | undefined {
+  // returns P2SH or P2PKH address, empty string otherwise
```
chedieck, line 184: comment obsolete

chedieck (changes requested):
LGMT, just a small thing

services/chronikService.ts
vorg-code: fixed

Klakurka (dismissed):

Klakurka:
Conflicts @vorg-code

vorg-code:
@Klakurka fixed.

Klakurka (approved):

chedieck (approved):

PR #476, opened 2023-04-12
[Issue #264] feat: dont sync prices with transactions

chedieck:
Description
---
Removes the syncing of prices together with transactions. The reasoning for this is that the syncing was causing significant overload when syncing the transactions, and it was unnecessary, since our code is already supposed to have all prices synced upon build.

Test plan
---
Rebuild the codebase, it should function just as before (tho now a bit faster)

Remarks
---
- Now we only sync the codebase transactions after ALL prices have been syncing,
- Left `console.time` on the code to time how long is it taking to upsert the transacitions into the DB. I've noticed it takes significantly longer in the BullMQ jobs then when using the API. When we get reasonable upsert prices, I'll take it off.

Depends on
---
- [x] #471

### DIFF
--- a/jobs/initJobs.ts
+++ b/jobs/initJobs.ts
@@ -40,16 +40,6 @@ const main = async (): Promise<void> => {
     },
     children: [
       {
-        name: 'syncXECAddresses',
-        data: { networkId: XEC_NETWORK_ID },
-        opts: {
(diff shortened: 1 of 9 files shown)


Klakurka:
Checks failing.

Klakurka (approved):

PR #477, opened 2023-04-14
Feat/411 5 3 hide get block info get blockchain info

vorg-code:
Depends on:
---
#474 

Description
---
`getBlockInfo` and `getBlockchainInfo` are no longer available in blockchainService file to be used by the rest of the application, they are too blockchain specific. `getLastBlockTimestamp` was created in their place since it is the only use of those methods to get the time of the last block.

Test plan
---
These functions are used only in the network page, so, to test that everything is still working as usual:
ecash: `'chronik'`,
bitcoincash: 'grpc'
}
2. open: http://localhost:3000/networks

### DIFF
--- a/services/blockchainService.ts
+++ b/services/blockchainService.ts
@@ -82,8 +82,8 @@ export async function syncTransactionsForAddress (parameters: GetAddressTransact
 
 export async function getLastBlockTimestamp (networkSlug: string): Promise<number> {
   const client = getObjectValueForNetworkSlug(networkSlug, BLOCKCHAIN_CLIENTS)
-  const getBlockchainInfo = await client.getBlockchainInfo(networkSlug)
-  const lastBlockInfo = await client.getBlockInfo(networkSlug, getBlockchainInfo.height)
+  const blockchainInfo = await client.getBlockchainInfo(networkSlug)
+  const lastBlockInfo = await client.getBlockInfo(networkSlug, blockchainInfo.height)
   return lastBlockInfo.timestamp
 }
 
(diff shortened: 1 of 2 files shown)


Klakurka (dismissed):

Klakurka:
Conflicts.

vorg-code:
@Klakurka fixed

Klakurka (dismissed):

services/blockchainService.ts
```diff
@@ -80,12 +80,11 @@ export async function syncTransactionsForAddress (parameters: GetAddressTransact
...
 
-export async function getBlockchainInfo (networkSlug: string): Promise<BlockchainInfo> {
-  return await getObjectValueForNetworkSlug(networkSlug, BLOCKCHAIN_CLIENTS).getBlockchainInfo(networkSlug)
-}
-
-export async function getBlockInfo (networkSlug: string, height: number): Promise<BlockInfo> {
-  return await getObjectValueForNetworkSlug(networkSlug, BLOCKCHAIN_CLIENTS).getBlockInfo(networkSlug, height)
+export async function getLastBlockTimestamp (networkSlug: string): Promise<number> {
+  const client = getObjectValueForNetworkSlug(networkSlug, BLOCKCHAIN_CLIENTS)
+  const getBlockchainInfo = await client.getBlockchainInfo(networkSlug)
```
chedieck, line 85: I think this variable could be called `blockchainInfo`, without the `get`, for better clarity

chedieck (changes requested):
LGTM, just a minor thing

services/blockchainService.ts
vorg-code: indeed, name improved

chedieck (approved):

PR #478, opened 2023-04-18
Feat/411 6 1 subscribe transactions

vorg-code:
Description
---
Subscribes to receive broadcast of new transactions in the blockchain (grpc client only). They should be added to our database at the same time as they happen in the blockchain, first as an unconfirmed transaction, then as a confirmed transaction when it confirms.

Test plan
---

- add an address you control to the application (one that will receive the transfer)
- run: yarn docker c, then flushall, then exit
- run: yarn docker jr
- run: yarn docker jw
- transfer to the address in step 1
- run database select * from paybutton.Transaction, you will see your transaction added as unconfirmed fist, then as confirmed

### DIFF
--- a/constants/index.ts
+++ b/constants/index.ts
@@ -109,13 +109,13 @@ export const QUOTE_IDS: KeyValueT<number> = { USD: 1, CAD: 2 }
 
 export type BLOCKCHAIN_CLIENT_OPTIONS = 'grpc' | 'chronik'
 export const NETWORK_BLOCKCHAIN_CLIENTS: KeyValueT<BLOCKCHAIN_CLIENT_OPTIONS> = {
-  ecash: 'grpc',
+  ecash: 'chronik',
   bitcoincash: 'grpc'
 }
 
 export const CHRONIK_CLIENT_URL = 'https://chronik.be.cash/xec'
 
-export const UPSERT_TRANSACTION_PRICES_ON_DB_TIMEOUT = 15000
+export const UPSERT_TRANSACTION_PRICES_ON_DB_TIMEOUT = 45000
 export const NUMBER_OF_TRANSACTIONS_TO_SYNC_INITIALLY = 200
 
 export const PAYMENT_WEEK_KEY_FORMAT = 'YYYY:WW'
--- a/constants/index.ts
+++ b/constants/index.ts
@@ -43,8 +43,7 @@ export const RESPONSE_MESSAGES = {
   DEFAULT_WALLET_CANNOT_BE_DELETED_400: { statusCode: 400, message: 'A default wallet cannot be deleted.' },
   USER_PROFILE_NOT_FOUND_400: { statusCode: 400, message: 'User profile not found.' },
   CACHED_PAYMENT_NOT_FOUND_404: { statusCode: 404, message: 'Cached payment not found.' },
-  NO_ADDRESS_FOUND_FOR_TRANSACTION_404: { statusCode: 404, message: 'No address found for transaction.' },
-  ADDRESSES_ALREADY_SUBSCRIBED_400: { statusCode: 400, message: 'Addresses were already subscribed.' }
+  NO_ADDRESS_FOUND_FOR_TRANSACTION_404: { statusCode: 404, message: 'No address found for transaction.' }
 }
 
 export interface KeyValueT<T> {
@@ -95,9 +94,10 @@ export const SYNC_TXS_JOBS_RETRY_DELAY = 2000
 export const BCH_TIMESTAMP_THRESHOLD = 1501588800 // 2017 Aug 1, 12PM
 export const XEC_TIMESTAMP_THRESHOLD = 1605398400 // 2020 Nov 15, 12AM
 
+export const DEFAULT_WORKER_LOCK_DURATION = 120000
 // Wait time (in ms) between sync of current prices
 export const CURRENT_PRICE_SYNC_DELAY = 60000
-export const DEFAULT_WORKER_LOCK_DURATION = 120000
+export const SUBSCRIBE_ADDRESSES_RETRY_DELAY = 60000
 
 export const NETWORK_TICKERS: KeyValueT<string> = {
   ecash: 'XEC',
--- a/constants/index.ts
+++ b/constants/index.ts
@@ -43,7 +43,8 @@ export const RESPONSE_MESSAGES = {
   DEFAULT_WALLET_CANNOT_BE_DELETED_400: { statusCode: 400, message: 'A default wallet cannot be deleted.' },
   USER_PROFILE_NOT_FOUND_400: { statusCode: 400, message: 'User profile not found.' },
   CACHED_PAYMENT_NOT_FOUND_404: { statusCode: 404, message: 'Cached payment not found.' },
-  NO_ADDRESS_FOUND_FOR_TRANSACTION_404: { statusCode: 404, message: 'No address found for transaction.' }
+  NO_ADDRESS_FOUND_FOR_TRANSACTION_404: { statusCode: 404, message: 'No address found for transaction.' },
+  ADDRESSES_ALREADY_SUBSCRIBED_400: { statusCode: 400, message: 'Addresses were already subscribed.' }
 }
 
 export interface KeyValueT<T> {
--- a/constants/index.ts
+++ b/constants/index.ts
@@ -96,8 +96,7 @@ export const XEC_TIMESTAMP_THRESHOLD = 1605398400 // 2020 Nov 15, 12AM
 
 export const DEFAULT_WORKER_LOCK_DURATION = 120000
 // Wait time (in ms) between sync of current prices
-export const CURRENT_PRICE_SYNC_DELAY = 60000
-export const SUBSCRIBE_ADDRESSES_RETRY_DELAY = 60000
+export const CURRENT_PRICE_REPEAT_DELAY = 60000
 
 export const NETWORK_TICKERS: KeyValueT<string> = {
   ecash: 'XEC',
--- a/constants/index.ts
+++ b/constants/index.ts
@@ -109,7 +109,7 @@ export const QUOTE_IDS: KeyValueT<number> = { USD: 1, CAD: 2 }
 
 export type BLOCKCHAIN_CLIENT_OPTIONS = 'grpc' | 'chronik'
 export const NETWORK_BLOCKCHAIN_CLIENTS: KeyValueT<BLOCKCHAIN_CLIENT_OPTIONS> = {
-  ecash: 'chronik',
+  ecash: 'grpc',
   bitcoincash: 'grpc'
 }
 
--- a/constants/index.ts
+++ b/constants/index.ts
@@ -10,6 +10,7 @@ export const RESPONSE_MESSAGES = {
   NAME_NOT_PROVIDED_400: { statusCode: 400, message: "'name' not provided." },
   PAYBUTTON_NAME_ALREADY_EXISTS_400: { statusCode: 400, message: 'Button name already exists.' },
   WALLET_NAME_ALREADY_EXISTS_400: { statusCode: 400, message: 'Wallet name already exists.' },
+  TRANSACTION_ALREADY_EXISTS_FOR_ADDRESS_400: { statusCode: 400, message: 'Transaction already exists for address.' },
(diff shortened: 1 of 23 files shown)


Klakurka:
I can't get the subscribing to work -- need some help.

chedieck (changes requested):
Doesn't work for newly added addresses, only for addresses that already were in the database the moment the jobs started

jobs/workers.ts
```diff
@@ -107,7 +113,7 @@ export const syncUnsyncedAddressesWorker = async (queue: Queue): Promise<void> =
     async (job) => {
       const newAddresses = await addressService.fetchUnsyncedAddresses()
       if (newAddresses.length !== 0) {
-        await syncAndSubscribeAddressList(newAddresses)
+        await syncAddressList(newAddresses)
```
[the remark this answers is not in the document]
vorg-code: when it runs, the code gets all database's addresses, so new addresses would be subscribed in the next run of this task:

`const addresses = await addressService.fetchAllAddresses()
    await subscribeAddressesAddTransactions(addresses)`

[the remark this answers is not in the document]
vorg-code: the lockout is 2 minutes, so this job should run every 2 min I guess?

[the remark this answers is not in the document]
vorg-code: Ah, but the retry is missing, I will adjust that and run every 1 minute this job, it is very light as it just gets the address string list and register subscription methods (for addresses that aren't subscribed already).

jobs/initJobs.ts
```diff
@@ -67,7 +70,19 @@ const main = async (): Promise<void> => {
                   jobId: 'syncPastPrices',
                   ...RETRY_OPTS
                 },
-                queueName: pricesSync.name
+                queueName: pricesSync.name,
```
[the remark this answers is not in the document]
vorg-code: done, subscribe now runs after all other jobs

jobs/workers.ts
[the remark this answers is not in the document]
vorg-code: image

jobs/initJobs.ts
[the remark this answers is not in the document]
vorg-code: image

Klakurka:
Conflicts

vorg-code:
@Klakurka solved.

chedieck:
@vorg-code tested again but still does not work.

Steps to reproduce:
1. Recreate the containers with `make reset-dev`
2. Go to the Buttons view and create a paybutton with one of your addresses
3. Transfer XEC to this address
4. Wait for it to appear in the button detail view (refresh needed). It won't appear.

jobs/initJobs.ts
```diff
@@ -16,58 +17,77 @@ const RETRY_OPTS = {
...
+    opts: {
+      removeOnComplete: false,
+      removeOnFail: { count: 3 },
+      jobId: 'subscribeTransactions',
+      ...RETRY_OPTS
+    }
+  }
+
+  const flowProducer = new FlowProducer({ connection: redisBullMQ })
+  await flowProducer.add({
```
chedieck, line 79: This is great, much more readable

services/grpcService.ts
```diff
@@ -203,27 +212,41 @@ export class GrpcBlockchainClient implements BlockchainClient {
...
+    if (addresses.length === 0) return
+
+    const addressesAlreadySubscribed = addresses.filter(address => Object.keys(this.subscribedAddresses).includes(address.address))
+    if (addressesAlreadySubscribed.length === addresses.length) return
+    addressesAlreadySubscribed.forEach(address => {
+      console.log(`This address was already subscribed: ${address.address}`)
+    })
+    addresses = addresses.filter(address => !addressesAlreadySubscribed.includes(address))
+
+    const addressesInClients: KeyValueT<Address[]> = {}
```
chedieck, line 225: I think some comments in this method summarizing code blocks would be good, e.g: here could be "intiialize one empty array for each client"

services/grpcService.ts
```diff
@@ -203,27 +212,41 @@ export class GrpcBlockchainClient implements BlockchainClient {
...
+    this.availableNetworks.forEach(networkSlug => {
+      addressesInClients[networkSlug] = []
+    })
+    addresses.forEach(address => {
```
chedieck, line 229: Actually I think this block from 225 to 233 could be in a separate auxiliary method (maybe in `utils/index` or `services/addressService`) like `groupAddressesByNetwork`

services/grpcService.ts
```diff
@@ -203,27 +212,41 @@ export class GrpcBlockchainClient implements BlockchainClient {
...
+      const prefix = getAddressPrefix(address.address)
+      if (!Object.keys(addressesInClients).includes(prefix)) { throw new Error(RESPONSE_MESSAGES.INVALID_ADDRESS_400.message) }
+      addressesInClients[prefix].push(address)
+    })
+
+    for (const [key, value] of Object.entries(addressesInClients)) {
```
chedieck, line 235: this loop could use better names than `key` and `value` such as `network` and `networkAddresses`

services/grpcService.ts
```diff
@@ -203,27 +212,41 @@ export class GrpcBlockchainClient implements BlockchainClient {
...
         includeBlockAcceptance: false,
-        addresses
+        addresses: addressesToSubscribe
       })
 
       const nowDateString = (new Date()).toISOString()
-
       // output for end, error or close of stream
       void confirmedStream.on('end', () => {
         console.log(`${nowDateString}: addresses ${addresses.join(', ')} confirmed stream ended`)
```
chedieck, line 252: I feel that here and below it shouldn't be `addresses.join` but `addressesToSubscribe.join`.

services/grpcService.ts
```diff
@@ -246,19 +269,54 @@ export class GrpcBlockchainClient implements BlockchainClient {
 
       // output for data stream
       void confirmedStream.on('data', (data: TransactionNotification) => {
-        const objectTxn = data.getConfirmedTransaction()!.toObject()
-        console.log(`${nowDateString}: got confirmed txn`, objectTxn)
-        onTransactionNotification(objectTxn)
+        void this.processSubscribedNotification(data)
```
chedieck, line 272: Why not for the unconfirmedStream also?

services/grpcService.ts
```diff
@@ -246,19 +269,54 @@ export class GrpcBlockchainClient implements BlockchainClient {
...
+
+      // remove unconfirmed transactions that have now been confirmed
+      const transactionsToDelete = await fetchUnconfirmedTransactions(confirmedTransaction.hash as string)
+      await deleteTransactions(transactionsToDelete)
     }
-    await createTxnStream()
-  };
+
+    // get new unconfirmed transactions
+    const unconfirmedTransaction = data.getUnconfirmedTransaction()?.toObject()
```
chedieck, line 310: AFAIK this won't ever get called as `processSubscribedNotification` is only being called on the `confirmedTransaction` stream. Maybe refactor this to accept a `confirmed` parameter?

chedieck (changes requested):

Klakurka:
Conflicts

jobs/initJobs.ts
```diff
@@ -16,58 +17,77 @@ const RETRY_OPTS = {
 }
 
 const main = async (): Promise<void> => {
-  // sync all db addresses transactions
-  const initTransactionsSync = new Queue('initTransactionsSync', { connection: redisBullMQ })
+  const pricesQueue = new Queue('pricesSync', { connection: redisBullMQ })
+  const initTransactionsQueue = new Queue('initTransactionsSync', { connection: redisBullMQ })
+  const newAddressesQueue = new Queue('newAddressesSync', { connection: redisBullMQ })
+  const subscribeTransactionsQueue = new Queue('subscribeTransactionsSync', { connection: redisBullMQ })
```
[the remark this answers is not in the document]
vorg-code: nice catch, fixed

services/grpcService.ts
vorg-code: done

services/grpcService.ts
vorg-code: indeed, much better, done

services/grpcService.ts
vorg-code: indeed, done

services/grpcService.ts
vorg-code: indeed! done

services/grpcService.ts
vorg-code: ahá! forgot it, nice catch, fixed

services/grpcService.ts
vorg-code: fix in the comment above solves this

vorg-code:
@Klakurka conflicts solved

@chedieck I changed to chronik because initially I was implementing both changes in this PR, then I forgot to change back, that was reverted all your remarks were implemented.

chedieck:
> @Klakurka conflicts solved
> 
> @chedieck I changed to chronik because initially I was implementing both changes in this PR, then I forgot to change back, that was reverted all your remarks were implemented.

Alright, but there still conflicts now

constants/index.ts
```diff
@@ -94,9 +94,10 @@ export const SYNC_TXS_JOBS_RETRY_DELAY = 2000
 export const BCH_TIMESTAMP_THRESHOLD = 1501588800 // 2017 Aug 1, 12PM
 export const XEC_TIMESTAMP_THRESHOLD = 1605398400 // 2020 Nov 15, 12AM
 
+export const DEFAULT_WORKER_LOCK_DURATION = 120000
 // Wait time (in ms) between sync of current prices
 export const CURRENT_PRICE_SYNC_DELAY = 60000
-export const DEFAULT_WORKER_LOCK_DURATION = 120000
+export const SUBSCRIBE_ADDRESSES_RETRY_DELAY = 60000
```
chedieck, line 100: better rename "retry" to "repeat", if I understood correctly, because "retry" conveys it failed and it will retry, whereas this is being used to determine the interval in which the job will repeat regardless of it's success.

jobs/initJobs.ts
```diff
@@ -16,58 +17,77 @@ const RETRY_OPTS = {
...
       {
-        name: 'syncBCHAddresses',
-        data: { networkId: BCH_NETWORK_ID },
-        opts: {
-          removeOnFail: false,
-          jobId: 'syncBCHAddresses',
-          ...RETRY_OPTS
-        },
-        queueName: initTransactionsSync.name,
+        ...flowJobSyncUnsyncedAddresses,
```
chedieck, line 83: Testing this, I get the following behavior:
image

The relevant part here is that there is a huge gap between syncing new addresses and subscribing them.

When a new address is added, we have to sync it's previous transactions and subscribe it so we get new transactions too. A gap between this two operations means we potentially lose a transaction that happens in between, so we want to make the gap as small as possible. Better yet, actually, we would first subscribe it and then, right away, we would sync its previous txs because then what we risk is commiting the same tx to the database, which is not a problem if we just parse this error and ignore it.

So I think the order should be inverted in such a way that we only sync the address txs after it is already subscribed.

chedieck (changes requested):
After simply resetting the containers I get the following behavior:
image

I haven't interacted with the front end nor added any other address:
image

jobs/initJobs.ts
chedieck: A more basic question: what is the rationale behind separating syncing new addresses and subscribing them into different jobs? As far as I know, those things are very much intertwined so I see no reason for it, while I see the disadvantage of creating the aforementioned gap.

Klakurka:
Conflicts.

constants/index.ts
vorg-code: indeed, done

jobs/initJobs.ts
vorg-code: Great point, sir. I removed the separate flow for subscribing and now we subscribe the address immediately before syncing it, so there is no gap anymore.

vorg-code:
@chedieck remarks addressed, much better subscribing before syncing, with 0 gaps. Now I feel this is truly beautiful. Worked flawlessly on my test (I only tested before merging master into it, but just a few conflicts, shouldn't have changed anything).

@Klakurka conflicts gone.

No need to remake all the environment, a fast way to test is to just delete all transactions and then restart the jobs:
delete from paybutton.PricesOnTransactions;
delete from paybutton.Transaction;

chedieck (approved):
Works well for me!

Issue #479, opened 2023-04-19
Upserting txs taking too long


chedieck:
Ideally we should _create_ them, and not upserts. We shouldn't be trying to put the same transaction into the DB more than once.

PR #480, opened 2023-04-19
chore: update prisma

chedieck:
Description
---
Updates the prisma module.

Test plan
---
Recreate the containers, run the tests. Should work with no issues.

Remarks
---
This actually solves #479. Initial sync of txs should be way faster now.


Klakurka (changes requested):
Dashboard doesn't load for me:

image

chedieck:
> Dashboard doesn't load for me:

Should be fixed now.

Klakurka (approved):

PR #481, opened 2023-04-19
[Issue #479] feat/create transaction not upserting

chedieck:
Depends on
---
- [x] #480 

Description
---
Reformulated the logic responsible to upsert transactions so that transactions are created, not upserted. The difference, essentially, is that if we try to create a transaction that already exists, it will return an error (as it should, because we should not be reinserting transactions).

Test plan
---
Initial sync of txs should happen normally.

### DIFF
--- a/services/transactionService.ts
+++ b/services/transactionService.ts
@@ -77,7 +77,6 @@ export async function createTransaction (transactionData: Prisma.TransactionUnch
   if (transactionData.amount === new Prisma.Decimal(0)) { // out transactions
     return
   }
-  transactionData.hash = await base64HashToHex(transactionData.hash)
   return await prisma.transaction.create({
     data: transactionData,
     include: includeAddressAndPrices
--- a/services/transactionService.ts
+++ b/services/transactionService.ts
@@ -95,7 +95,7 @@ async function addUUIDToTransactions (transactions: Prisma.TransactionUncheckedC
   return uuidList
 }
 
-export async function createManyTransactionsForAddress (transactionsData: Prisma.TransactionUncheckedCreateInput[]): Promise<TransactionWithAddressAndPrices[]> {
+export async function createManyTransactions (transactionsData: Prisma.TransactionUncheckedCreateInput[]): Promise<TransactionWithAddressAndPrices[]> {
   const uuidList = await addUUIDToTransactions(transactionsData)
   await prisma.transaction.createMany({
     data: transactionsData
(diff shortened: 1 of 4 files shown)


vorg-code:
That won't do for chronik and grpc sync transactions method to just change them from `upsert` to `create` calls, on the next sync run they will start throwing exceptions. We should implement #403 logic here to prevent those exceptions and also improve performance.

Klakurka (dismissed):

Klakurka:
Conflicts

chedieck:
> That won't do for chronik and grpc sync transactions method to just change them from `upsert` to `create` calls, on the next sync run they will start throwing exceptions. We should implement #403 logic here to prevent those exceptions and also improve performance.

AFAIK the subscription happens after the sync is completed, so I don't understand how that would be possible. But in any case, I think it's good we throw error if that happen. Since txs do not change after they are completed, we shouldn't spend computing power accessing the database to reinsert txs which are already there.

chedieck:
> Conflicts

Rebased.

services/transactionService.ts
```diff
@@ -72,45 +73,43 @@ export async function base64HashToHex (base64Hash: string): Promise<string> {
...
     return
   }
-  const hash = await base64HashToHex(transaction.hash)
-  const transactionParams = {
-    hash,
-    amount: transaction.amount,
-    addressId: address.id,
-    timestamp: transaction.timestamp,
-    confirmed: transaction.confirmed
+  transactionData.hash = await base64HashToHex(transactionData.hash)
```
chedieck, line 80: Actually, I think this shouldn't be here. @vorg-code do you know why this is needed? With this refactor this would have to be done seperately also in the `createManyTransactions` method, but I think this should be done only when it's necessary, in the place that calls it. I think it's chronik, so I put it there, but I am not so sure

vorg-code:
> AFAIK the subscription happens after the sync is completed, so I don't understand how that would be possible. But in any case, I think it's good we throw error if that happen. Since txs do not change after they are completed, we shouldn't spend computing power accessing the database to reinsert txs which are already there.

@chedieck This can happen with new addresses as the repeating subscribe job runs independent of the sync new address job as the PR is now. Probably a good idea to change that. 
But there is another possibility, previous address sync didn't finish because of some error either on the blockchain client or logic, then when it tries to run again, it will eventually try to sync a transaction it has already synced. Very simple example: address has 2 transactions, syncs 1 then errors, when server goes down and up again, lets say another transaction has happened, it will try to sync the 3 transactions, 1st will sync, 2nd will error because it will try to create when it already exists.

Klakurka (approved):

Issue #466, opened 2023-04-05
Edit paybutton address should be more user friendly


chedieck:
Suppose a user wants to add a BCH address to the following button:

![image](https://user-images.githubusercontent.com/21281174/230227501-8ee039d6-a124-45b8-80de-02c6b6b9d233.png)

the placeholder shows the currently present ecash address. So he would have to somehow type that all again, break a line, and add the bitcoincash address, which would be very annoying

Klakurka:
Perhaps we can just put the address(es) in there as actual values rather than as a placeholder. I get that having it as a placeholder means that we can easily tell if it was actually changed or not when processing the submit but probably not a big deal.

Klakurka:
Ya I think just putting the actual values in there is best. We can probably keep the placeholder in there just for their reference if they delete the value in there while editing.

PR #483, opened 2023-04-20
[Issue #466] Improve edit button form ui

johnkuney:
Description
The current values of the button as a placeholder in the form was not ideal as it would disappear if new text was entered. 
Changed it to be the actual value for the inputs so a user can edit it more easily

Test plan
Try editing a button name and addresses and see if it works better and is easy to use


Klakurka (commented):
I was thinking we could keep the placeholder as it was rather than replacing it. What do you think?

Klakurka (approved):

PR #485, opened 2023-04-24
[Issue #264] feat: connect prices when inserting transactions

chedieck:
Depends on
---
- [x] #487 

Description
---
When creating transactions, automatically connect them to the existent price.

Test plan
---
1. Spin up the containers
2. After all is up, add a new address that has some transactions on it. Make sure you see the transactions in the button detail view.
3. Enter the database with `yarn docker db`
5. The number returned by `SELECT COUNT(*) FROM PricesOnTransactions;` should be twice the number returned by `SELECT COUNT(*) FROM Transaction;`

Remarks
---
The rationale for this is so that the cached payments, that appear on the dashboard, can be updated as soon as a new transaction is inserted. This PR still does not acomplishes that yet, but makes so that the `createManyTransactions` and `createTransaction` functions return transactions already with it's connected prices.

I also removed all the code related to the previous logic of price syncing (on demand: if a price didn't exist, we would fetch it. Now we expect all prices to exist and just connect the txs to them.)

### DIFF
--- a/constants/index.ts
+++ b/constants/index.ts
@@ -38,6 +38,7 @@ export const RESPONSE_MESSAGES = {
   NETWORK_SLUG_NOT_PROVIDED_400: { statusCode: 400, message: "'networkSlug' not provided." },
   QUOTE_SLUG_NOT_PROVIDED_400: { statusCode: 400, message: "'quoteSlug' not provided." },
   NO_CURRENT_PRICES_FOUND_404: { statusCode: 404, message: 'Current prices not found.' },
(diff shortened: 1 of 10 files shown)


Klakurka (changes requested):
The transactions within the button showed up immediately but not on the dashboard.

`SELECT COUNT(*) FROM PricesOnTransactions;` returned 40 and `SELECT COUNT(*) FROM Transaction;` returned 72, so possibly not what was expected?

PR #489, opened 2023-04-26
[Issue #264] feat: update payments cache

chedieck:
Depends on
---
- [x] #488

Description
---
Now the cache updates when a new address is added or when a new transactions comes. This is updating the transaction even when it's confirmed, I'll open an issue to deal with this (so it is only added after confirmation) and address is after #478, since it changes this logic.

Test plan
---
Add an address, preferably one of yours. See if the payments in the dashboard correctly reflect this addition. Send some XEC to that address, see if that also is correctly updated. Finally, delete the address, see if it went away.

Remarks
---
Removed redis cache persistence. The db is reset with the containers, so no point in persisting the cache if we don't do it with the DB.


Klakurka:
Checks failing.

chedieck:
> Checks failing.

Should be fixed now

Klakurka:
Conflicts.

Klakurka (changes requested):
Both the dashboard and the transactions page don't show any txs for the address that I add. I get this in the logs:

image

chedieck:
> Both the dashboard and the transactions page don't show any txs for the address that I add. I get this in the logs:
> 
> image

When do you get that, exactly? after adding the address?

Klakurka:
Yeah... Might be at other times too. I'll have to check.

Klakurka:
image

I get this after adding a new button with txs, then accessing the dashboard.

chedieck:
> I get this after adding a new button with txs, then accessing the dashboard.

@Klakurka should be fixed now. It's possible that you have to delete `redis/dump.rdb` (delete it if it exists) because this PR changes the redis behavior to not persist anything.

Klakurka (changes requested):
I don't get any errors now but adding a new button doesn't seem to pull in the txs for it.

image

Klakurka (approved):

Issue #491, opened 2023-04-26
Deleting/creating a paybutton does not update the association between its address and the user


chedieck:
the table `AddressesOnUserProfile` is not being updated by the operations of deleting, creating and updating paybuttons.

PR #497, opened 2023-05-01
[Issue #491] refactor: user profile same id as supertokens

chedieck:
Description
---
Changes the `UserProfile` table so its primary key is the same as the table in the `supertokens` table.

Before, if I had the `id` for a `userProfile`, I would have to fetch that info from the DB to get its `userId` just to e.g. see the wallets or addresses associated with it. Now `id` and `userId` are the same thing, which is what before was called `userId`.

Test plan
---
Nothing should change.


Klakurka:
Was there a reason why we made them different in the first place?

chedieck:
> Was there a reason why we made them different in the first place?

I think it was an overlook on my part.

Klakurka (approved):

Issue #498, opened 2023-05-01
Wallets networks not updating when paybuttons change


chedieck:
On going to the wallets view, the listed networks are not correct after having deleted or created a new paybutton.

PR #499, opened 2023-05-01
[#Issue 498, #491] feat: creating and deleting paybuttons updated address to user relation

### LINKED ISSUE #491: Deleting/creating a paybutton does not update the association between its address and the user

chedieck:
Depends on
---
- [x] #497

Description
---
Almost fixes #491 and #498: Updates the table `AddressesOnUserProfiles` when paybuttons are removed or created, but not when they are updated 

Paybutton update will be addressed in the next PR

Test plan
---
Create paybuttons, delete paybuttons, and check in the wallets view if the shown networks are correctly reflecting the addresses associated do the wallets through its paybuttons.


Klakurka (changes requested):
Checks failing.

Klakurka (approved):

PR #501, opened 2023-05-04
411 6 2 chronik subscribe transactions implemented

vorg-code:
Description
---
Subscribes to receive broadcast of new transactions in the blockchain (chronik client only). They should be added to our database at the same time as they happen in the blockchain, first as an unconfirmed transaction, then as a confirmed transaction when it confirms.

Also, this PR sets chronik as the blockchain client to be used for all XEC addresses.

Test plan
---
- add an address you control to the application (one that will receive the transfer)
- run: yarn docker c, then flushall, then exit
- run: yarn docker jr
- run: yarn docker jw
- transfer to the address in step 1
- run database select * from paybutton.Transaction, you will see your transaction added as unconfirmed fist, then as confirmed

### DIFF
--- a/services/chronikService.ts
+++ b/services/chronikService.ts
@@ -1,21 +1,26 @@
-import { ChronikClient, ScriptType, Tx, Utxo } from 'chronik-client'
+import { ChronikClient, ScriptType, SubscribeMsg, Tx, Utxo, WsConfig, WsEndpoint } from 'chronik-client'
 import { encode, decode } from 'ecashaddrjs'
 import bs58 from 'bs58'
-import { BlockchainClient, BlockchainInfo, BlockInfo, GetAddressTransactionsParameters, TransactionDetails } from './blockchainService'
-import { NETWORK_SLUGS, RESPONSE_MESSAGES, CHRONIK_CLIENT_URL, XEC_TIMESTAMP_THRESHOLD, XEC_NETWORK_ID, BCH_NETWORK_ID, BCH_TIMESTAMP_THRESHOLD, FETCH_DELAY, FETCH_N } from 'constants/index'
-import { TransactionWithAddressAndPrices, createManyTransactions, base64HashToHex } from './transactionService'
+import { AddressWithTransaction, BlockchainClient, BlockchainInfo, BlockInfo, GetAddressTransactionsParameters, TransactionDetails } from './blockchainService'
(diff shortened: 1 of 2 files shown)


services/chronikService.ts
```diff
@@ -157,9 +162,67 @@ export class ChronikBlockchainClient implements BlockchainClient {
...
+
+  private async processWsMessage (msg: SubscribeMsg): Promise<void> {
+    // delete unconfirmed transaction from our database
+    if (msg.type === 'RemovedFromMempool' || msg.type === 'Confirmed') {
+      const transactionsToDelete = await fetchUnconfirmedTransactions(msg.txid)
+      await deleteTransactions(transactionsToDelete)
+    }
+    // create unconfirmed or confirmed transaction
+    if (msg.type === 'AddedToMempool' || msg.type === 'Confirmed') {
+      const transaction = await await this.chronik.tx(msg.txid)
```
chedieck, line 183: why two awaits?

services/chronikService.ts
```diff
@@ -157,9 +162,67 @@ export class ChronikBlockchainClient implements BlockchainClient {
...
+      const addressWithTransactions = await this.getPrismaTransactionsForSubscribedAddresses(transaction)
```
chedieck, line 184: I am confused about this method which is being called. It returns a dictionary where the keys are all addresses subscribed, why? This transaction will probably have to do with only 1 or 2 of the subscribed addresses AFAIK, so why go through all subscribed addresses and then create the transaction for all of them? Looks wrong to me unless I am not understanding something correctly

chedieck: I am probably not understanding something correctly for I have tested and seems to be working.

chedieck (changes requested):
Tested and worked well, just some minor things and clarification

services/chronikService.ts
vorg-code: type error, fixed

services/chronikService.ts
vorg-code: A transaction can have multiple inputs and outputs, we don't know which of the subscribed addresses are in the transaction, so we get the amount transacted to each one of them in this transaction and then filter the result by only those addresses that have any amount transacted before returning:

`return addressWithTransactions.filter(
   addressWithTransaction => addressWithTransaction.transaction.amount > new Prisma.Decimal(0)
 )`

chedieck: got it, very cool

chedieck (approved):

Issue #334, opened 2022-12-04
Update pages automatically without the need to refresh


Klakurka:

Klakurka:
Needs to be retested.

Klakurka:
I found one issue. If you go into the /button/xxx page and use the pencil icon to update the addresses in a button, you'll need to refresh the page once before it will remove/hide any addresses that had been removed/replaced during the update.

PR #506, opened 2023-05-17
[Issue #334] feat: Updating paybuttons txs automatically

chedieck:
Depends on
---
- [x] #505 

Description
---
Solves #334 for paybuttons. Live update transactions on the paybutton detail view.

Test plan
---
1. Rerun the containers
2. Go to a paybutton with one of your addresses (create one if needed)
3. Send some XEC to it. You should very quickly see the unconfirmed transaction there, without the needing to refresh the page.
4. When the transaction is confirmed, this also should be updated immediately.

Technical Remarks
---
- I changed the behavior introduced in #505 so that we don't fetch all txs again when a new one arrives, but simply use the data from the new one to update all clients. This means the updating the client requires no DB access.
- I refactored the `PaybuttonDetail` component to be a functional component instead of a class component. The reasoning is that not only this is encouraged in the current [react docs](https://react.dev/reference/react/Component), but also was easier to setup the cleanup for the SSE clients when using `useEffect`.


Klakurka (changes requested):
I can't get this to work (it never updates automatically for me; no socket traffic) so I'll need some help.

Klakurka:
Also after pulling in 1 tx, it seems to stop completely even after refreshing.

Klakurka (approved):
OK I tested it with a less busy address and it does work.

One thing though is that it would be nice to show previous transactions too without a refresh. I'll make a separate task for this.

Issue #513, opened 2023-05-26
Big addresses smart handling


chedieck:
1. **Big addresses syncing is non blocking** — When a big address (one with many txs) is added, its sync does not make so that addresses added afterwards have to wait for its sync to finish so they themselves start syncing.
2. **Mid-sync tx arrival is well handled** — if a big address is in the middle of the sync and a tx arrives, it's gonna be synced from the subscription job. Nonetheless, when the sync is almost over, this tx will appear again and we will try to sync it again. Therefore, it's necessary to make sure this will lead in an error and this attempt to save in the DB a tx that is already there is simply ignored.

Klakurka:
We can push this to the v2 milestone if we can pre-sync the address before allowing public access and it's all working as expected for people.

chedieck:
Cool, then it's as simple as adding the desired addresses to be pre-synced in a seeds file that is only seeded when ENV = production

PR #517, opened 2023-05-29
[Issue #513] feat: Production seeding for ifp addresses

chedieck:
Depends on
---
- [x] #516 
- [x] #522

Description
---
Logic to save the synced txs for some address in a CSV file and seed it directly next time the containers are built.

Test plan
---
1. Modifly `.env.local`, add a line: `ENVIRONMENT=production`
2. Rerun the containers. The sync job should start syncing the IFP address `ecash:prfhcnyqnl5cgrnmlfmms675w93ld7mvvqd0y8lz07`. This can take a while. The progress can be estimated by running `SELECT count(*) FROM Transaction;` on the DB container, should go up to ~54k
3.  This will take some time to run. After is done, a CSV file with all the txs should appear on `prisma/seeds/productionTxs.csv
`.
4. Resetting the containers now will make so that these txs are inserting in the seeding and ignored on the sync job.

Known issues
---
Those will be addressed in a next PR:
- The cached Txs are not being connected to their prices,
- Caching the txs in the CSV file and reestoring it later makes so that some txs may be missing, if they arrived after the last cached tx.


Klakurka (changes requested):
I tried it on a different address with only about 1000 txs. It didn't automatically sync before I added a button for it.

Any ideas? Address is: `ecash:qrmynmlsw668klgg8ee3z77evjadds72kyfstea2xm`

Klakurka (changes requested):
I changed the address to one with fewer txs but still quite a few. It didn't try syncing it until I added a button for it.

Klakurka (approved):

PR #520, opened 2023-05-30
[Issue #519] feat: Production set-up: pm2 

chedieck:
Description
---
Improved startup scripts by using `pm2` process manager & added logic to run a production build.

Test plan
---
Currently, running the production environment still does not works, due to many TS errors that exist right now in the code base. I'll address that in a next PR

For now, the development env should work as before, except that everything now is being managed by `pm2`, so logs won't appear in the container log but can be accessed in the `logs/next-*.log` files, or read directly by using `yarn docker nextlogs`/`yarn docker nl` (equivalent to running `pm2 logs next` inside the docker container)

Remarks
---
1. Upon changing to this branch, there wil be some unstaged files: jobs/out.log and sse-service/out.log. This files have been deprecated in favor of now a `logs/` directory containing all the logs; they can simply be deleted.


ScottMcDermid (dismissed):

Klakurka (changes requested):
image

Still doesn't work for me.

PR #521, opened 2023-05-30
[Issue #422] feat: API endpoint to get user networks.

chedieck:
Description
---
Add endpoint to get the networks for a user.

Test plan
---
This uses the user session to get the userId, so not simple `curl` possible.

Best way to test, add the following lines:
```
      let temp = await fetch('/api/networks/user')
      console.log('teste: ', await temp.json())

```
on `pages/dashboard/index.tsx:65`, that is, right after `      const fetchData = async (): Promise<void> => {`

Then, upon accessing the dashboard, the user networks should appear in the **browser** console.

### DIFF
--- a/pages/api/networks/user.ts
+++ b/pages/api/networks/user.ts
@@ -1,0 +1,18 @@
+import * as networkService from 'services/networkService'
+import { setSession } from 'utils/setSession'
+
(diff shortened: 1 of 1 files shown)


johnkuney (approved):
looks good and worked well for me

PR #524, opened 2023-05-31
UI hide unused networks

johnkuney:
Description
As per #422 hiding info and labels on networks not in use

Test plan
Run the app and remove all buttons. You should not see any network specific "default" labels on any wallet screens, and both networks should be visible in the networks tab
Next try adding one ecash address and you should see XEC labels in those places and the BCH network should be hidden in the networks tab
Next try adding a BCH address and check the same


johnkuney:
Yes no prob, should be good now

Klakurka (approved):

Issue #544, opened 2023-06-17
"Verify your email address" infinite loop


Klakurka:
Steps to reproduce:

1. Access the prod URL: `https://api.paybutton.org`
2. Sign up as a new user
3. Check your email inbox and click the verify link
4. You should be redirected into the admin panel, but then it will try to send you to `https://api.paybutton.org/auth/verify-email?rid=thirdpartyemailpassword` and then back to the admin panel in an infinite loop

![image](https://github.com/PayButton/paybutton-server/assets/4713204/5a094567-30cf-492d-a5af-d75789ccf780)

Klakurka:
Was due to some bad caching. Scott disabled all caching for now but we'll want to re-introduce that eventually.

PR #551, opened 2023-06-27
[Issue #544] fix: infinite verification email loop

chedieck:
Related to #544 

Depends on
---
- [x] #553 

Description
---
Solves #544 in production.

Test plan
---
Only happened in production, so only testable there.

Remarks
---
Taken from: https://supertokens.com/docs/thirdparty/nextjs/setting-up-backend


chedieck:
seems like this is solved already, closing. If reappears we try it again.

Issue #539, opened 2023-06-09
/address/transactions doesn't have new txs which are necessary to show payment feedback


Klakurka:

Klakurka:
This is the URL that the button currently checks every second to see if there's been a new tx.

Klakurka:
Looks like it's a node issue.

Klakurka:
Still nothing even after node has sync'd.

PR #552, opened 2023-06-27
[Issue #539] fix: no new txs on endpoint

chedieck:
Solves #539 

Description
---
Endpoint `address/transactions` now will also subscribe the addresses, so that we monitor them for when new txs arrive.

Test plan
---
This is running now in prod, so it should solve #539


Klakurka (changes requested):
Payment is detected but the success process doesn't seem to run: https://github.com/PayButton/paybutton/issues/218

Up to you want to address that in this branch or make a new one.

Klakurka (approved):

Issue #567, opened 2023-07-01
Poll address-based


chedieck:
Right now the SSE server sends all tx data for all clients; change this so that the connection is made only for N addresses and only the data relevant to this addresses is sent.

PR #568, opened 2023-07-01
[Issue #567] feat: tx broadcast is address based

chedieck:
Related to #567

Depends on
---
- [x] #566

Description
---
Solves #567.
This is also part of the client plan to change from periodic requests to SSE connection to our server. Now the connection can be made based on the address and, therefore, the button can use it.

Test plan
---
All should work normally as before: e.g. receving a tx should result in it appearing without the need to refresh the button detail page.


Klakurka:
Github checks failing + conflicts.

Klakurka (approved):

Issue #565, opened 2023-07-01
Server stops polling new txs after a while


Klakurka:

Klakurka:
Still an issue.

Klakurka:
Looks like it might be working now after that last fix... I'll re-test later tonight before closing this.

Klakurka:
I think we're good!

If this comes up again, we can re-open and investigate.

PR #582, opened 2023-08-02
fix: only delete uncofirmed txs if they exist

chedieck:
Related to #565

Description
---
#565 may be happening for other reasons, but right now I noticed on the logs that sometimes we were trying to delete non-existent txs. This PR fixed that

Test plan
---
1. Add two of your addresses to some paybutton, one of which has spendable coins.
2. Transfer money from one to the other. On ABC's wallet you can do that by right clicking an address and selecting "Spend from this address".
3. When the tx is confirmed, it should in parallel try to delete the unconfirmed tx twice. This makes so that we get the error "Record doesn't exist", but this can also not happen since async is non deterministic (one process can be executed quite further away from the other so that when it tries to find the unconfirmed tx this has already been deleted). This error should happen on master but not in this branch.

Overall, this is tricky to test and long also. The changes only makes so that we ignore the case where we try to delete a transaction that doesn't exist, so one option is just to approve it and see if we actually stop getting this error on prod.


Klakurka:
I'm going to skip testing this one for the review and do a full test in prod to make sure things are working at that point.

We are now only deleting txs that were double spent effectively, which is correct.

Issue #581, opened 2023-07-25
Update readme (replace XECD with Chronik)


Klakurka:
Currently the readme still shows the old XECD config option.

PR #584, opened 2023-08-02
Chore/update readme

chedieck:
Related to #581

Description
---
Updates readme with current repo info

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 git_hook_setup = cp .githooks/pre-commit .git/hooks/pre-commit
 git_diff_to_master = git diff --name-only --diff-filter=ACMRTUXB origin/master > DIFF
-create_test_paybutton_json = echo { \"priceAPIURL\": \"foo\", \"networkBlockchainClients\": { \"ecash\": \"chronik\", \"bitcoincash\": \"grpc\" }, \"chronikClientURL\": \"https://chronik.be.cash/xec\" } > paybutton-config.json
+create_test_paybutton_json = echo { \"priceAPIURL\": \"foo\", \"networkBlockchainClients\": { \"ecash\": \"chronik\", \"bitcoincash\": \"grpc\" }, \"chronikClientURL\": \"https://chronik.be.cash/xec\", \"wsBaseURL\": \"localhost:5000\" } > paybutton-config.json
 
 dev:
 	$(git_hook_setup)
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ https://paybutton.org
 - Create a `.env.local` file with two environment variables:
   + `PRICE_API_TOKEN="<COINDANCE_API_KEY>"`
   + `WS_AUTH_KEY="<RANDOMLY_GENERATED_UUID>"`
-  + You can get the coin.dance API key at [WIP]()
+  + You can get the coin.dance API key at [coin.dance](coin.dance/api) web page.
 
 - Run the following make command* to build/pull the relevant docker images and run the server locally:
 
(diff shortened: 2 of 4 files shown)


config/example-config.json
```diff
@@ -1,8 +1,8 @@
 {
   "apiDomain": "http://localhost:3000/api",
+  "apiBasePath": "/api/auth",
+  "websiteBasePath": "/auth",
   "websiteDomain": "http://localhost:3000",
-  "apiBasePath": undefined,
-  "websiteBasePath": undefined,
   "wsBaseURL": "http://localhost:5000",
   "showTestNetworks": false,
   "grpcBCHNodeURL": "bchd.greyh.at:8335",
```
Klakurka, line 8: Let's remove the actual example URL here.

README.md
```diff
@@ -7,6 +7,12 @@ https://paybutton.org
 ### Installing and running PayButton locally
 - [Install and configure Docker](https://docs.docker.com/get-docker/)
 - [Install docker-compose](https://docs.docker.com/compose/install/)
+- Clone the repo and go to the repo directory
+- Copy `config/example-config.json` to `paybutton-config.json`
+- Create a `.env.local` file with two environment variables:
+  + `PRICE_API_TOKEN="<COINDANCE_API_KEY>"`
+  + `WS_AUTH_KEY="<RANDOMLY_GENERATED_UUID>"`
+  + You can get the coin.dance API key at [coin.dance](coin.dance/api) web page.
```
Klakurka, line 15: For now let's just remove this line.

README.md
```diff
@@ -19,9 +25,12 @@ https://paybutton.org
 - App will be available at http://localhost:3000.
 - Local changes on source code should trigger a reload immediately.
 
+
 ### Optional configuration
+- To run on production mode, set `ENVIRONMENT=production` on `.env.local`. This would optimize the build so that the app is faster, but local changes will not trigger a reload immediately.
```
Klakurka, line 30: Change to:

For production, set `ENVIRONMENT=production` in `.env.local`. This optimizes the build for performance and skips the setup of various dev tools (like LiveReload).

Klakurka (changes requested):

config/example-config.json
chedieck: only for BCH or for both? Maybe change to something like `bchd.example.url`?

Klakurka: How about just localhost:8335?

chedieck: I think is a little misleading because most people wouldn't run their own GRPC servers but use a third party; whereas the other places where `localhost` appears are places where using it as `localhost` requires no additional set up of a third party service (such as a BCH GRPC node)

Makefile
```diff
@@ -1,6 +1,6 @@
 git_hook_setup = cp .githooks/pre-commit .git/hooks/pre-commit
 git_diff_to_master = git diff --name-only --diff-filter=ACMRTUXB origin/master > DIFF
-create_test_paybutton_json = echo { \"priceAPIURL\": \"foo\", \"networkBlockchainClients\": { \"ecash\": \"chronik\", \"bitcoincash\": \"grpc\" }, \"chronikClientURL\": \"https://chronik.be.cash/xec\" } > paybutton-config.json
+create_test_paybutton_json = echo { \"priceAPIURL\": \"foo\", \"networkBlockchainClients\": { \"ecash\": \"chronik\", \"bitcoincash\": \"grpc\" }, \"chronikClientURL\": \"https://chronik.be.cash/xec\", \"wsBaseURL\": \"localhost:5000\" } > paybutton-config.json
```
Klakurka, line 3: Can we replace the Chronik client URL with something generic as well?

PR #586, opened 2023-08-06
fix: no need to delete unconfirmed txs

chedieck:
Description
---
Remove the deleting of unconfirmed txs in GRPC, txs should be simply updated to confirmed.


Klakurka (approved):
Code looks good; tx processing still works.

Issue #156, opened 2022-07-06
v1 Account


Klakurka:
Display email and password, and allow them to change their password.

Klakurka:
TODO:

- Allow changing email
- First pass of improving the UI by @johnkuney

Klakurka:
I'm going to move changing their email out into a v2 #594.

PR #588, opened 2023-08-10
[Issue #156] feat: accounts page front-end

chedieck:
Related to #156

Description
---
Adds the front-end for the Account page
image

Note: this is **only** the front end, clicking the button shouldn't work.

Test plan
---
Verify that the usability of the front end is satisfactory

 
Remarks
---
I used the same icon of the Dashboard in the sidebar and marked it as WIP. This should be addresses on a different PR related to #587


Klakurka (approved):
Good enough for a v1.

PR #589, opened 2023-08-10
 [Issue #156] feat: accounts page back end

chedieck:
Related to #156 

Depends on
---
- [x] #588 

Description
---
Functional accounts page that allows for password changing.

Test plan
---
Try changing the password, also try inputting a wrong password, a new password that is equal to the old one, etc.


Klakurka (changes requested):
I tried changing the password but it didn't actually work. When I log out and log back in, the old password is still set.

image

chedieck:
> I tried changing the password but it didn't actually work. When I log out and log back in, the old password is still set.
> 

In this screenshot the button is greyed out, all fields are filled and it shows no error messages, which shouldn't happen.
That means you didn't actually got to press the button and receive the "successfully changed password" green text confirmation, right?

Could you see if there are error messages on your browser console?

Klakurka:
No errors and no network traffic when pressing the button just like I'd expect would be the case on the previous branch.

Klakurka (approved):

Issue #493, opened 2023-04-29
v1 Landing page


Klakurka:
Must have:
- PayButton logo / title
- "The easiest way to accept eCash online." subtitle
- Separate links Sign In and Sign Up
- Create Your PayButton section #6

Should have:
- Dark / light theme just like existing pages

Could have:
- Link to docs site

Won't have:
- Supporter section (v2 item) #5

PR #590, opened 2023-08-10
[Issue #493] feat: landing page

johnkuney:
Related to #493 

Description
Adding a landing page to the app 

Test plan
Run the app and navigate to localhost:3000
Check out the page on different screens and browsers
Try dark mode toggle
Try logging in 
Rest of the app should function as normal 

Remarks
I know we want to ideally have separate pages for login vs sign up, but I am still looking into how to achieve this, so for now both buttons go to the same place


Klakurka (changes requested):
- Obviously as you said, the Sign In & Sign Up buttons should point to the different dialogs if we can.
- Hide the Supporters section.

johnkuney:
Okay for sure, updated

components/LandingPage/GetStarted.tsx
```diff
@@ -0,0 +1,31 @@
...
+import style from '/styles/landing.module.css';
+
+export default function GetStarted(): JSX.Element {
+  return (
+    <div className={style.get_started_ctn}>
+      <div className={style.container}>
+        <h2 className={style.heading2}>Get Started</h2>
+        <p className={style.centerp}>
+          Simply add the following to your website’s HTML, replacing
+          <span className={style.your_address}>YOUR_ADDRESS_HERE</span> with
```
Klakurka, line 11: There's a space missing to the left of the YOUR_ADDRESS_HERE text now.

Klakurka (changes requested):
- Let's add a GitHub icon link in the header and footer. https://github.com/PayButton/ is a fine link I think since we have 2 main repos. Have this open in a new tab.
- Let's also add a link to the Telegram channel: https://t.me/paybutton. Have this open in a new tab.
- Just like I had mentioned before, the PayButton logo is for some reason a different colour as the rest of the text on the page (same off-dark-grey text colour as in the admin nav when logged in). Anything you think we should do about that?
image

johnkuney:
hmm thats still so weird on the logo being washed out. I cant recreate it. What about other images on the page? Like the background curves or dashboard image?

Klakurka:
- Landing page logos (in header and footer)
- Auth page logo (at top)
- Admin nav (all content)

chedieck:
> the Sign In & Sign Up buttons should point to the different dialogs if we can.

Do we have a task for this? Just created a PR #595

Klakurka:
> > the Sign In & Sign Up buttons should point to the different dialogs if we can.
> 
> Do we have a task for this? Just created a PR #595

No task open, just the one for visually differentiating the two pages.

Klakurka:
Note that the new signin/signup pages will have the following URLs `/signin` and `/signup` respectively so that will need to be changed too.

Klakurka (changes requested):
The landing page redirects to /signin immediately which is an issue.

johnkuney:
ah for sure will take a look. Im actually still trying to get it to run with this new branch, so havent seen it yet

johnkuney:
okay should be good now

Klakurka (changes requested):
- Can we add a copy button into the corner of the "Get Started" HTML section? I remember we had one w/ the button maker but I guess that was removed. It would be nice to have it if we could.
- Nit: Can we can a better blue 'pay' button where the corners are properly rounded? Could just take a screenshot from a zoomed in PayButton w/ that exact colours 
image
- Nit: How about getting a screenshot with slightly "better" data? It's weird/wrong to see "year total: 312" but then the chart only shows 0s and 1s for payment totals for each month. 
image

johnkuney:
Yep for sure, do you have a good address that has some transaction data that would make for a good image? I didnt have a good one off hand

Klakurka (changes requested):
- Github should be GitHub
- Signing into the admin panel has the landing page flash up for a second; probably out of scope for this specific branch but wanted to mention it.
- I can't seem to find a good mix of addresses for the chart screenshots so maybe you could just make them up if that's what you did for these? Just need to make sure that the math lines up otherwise somebody will think it's got a bug. The 'Revenue' side looks good but the 'Payments' side could use some variety.

johnkuney:
Okay for sure, any ideal values there for everything? Revenue amount is good? what about amount of buttons and payments?

Klakurka:
Maybe keep it under $30k for the year, something like that. Slow/gradual trend upwards over the year is also probably good.

johnkuney:
Okay added those and also went ahead and cleaned up the login pages here

Klakurka (changes requested):
- /reset-password UI needs to be updated
- /verify UI needs to be updated (likely need to actually going through the sign up + verify process to actually see the page properly)
- If you're logged in, navigating to the landing page now redirects you back to the dashboard. We should allow people to visit the dashboard if they want to even if they're logged in. Bonus points would be to show 'Dashboard' and 'Sign Out' instead of 'Sign Up' and 'Sign In' in the nav but we can v2 this if it's too much trouble.

johnkuney:
hows that looking now? For the changing the nav if logged in I think will save for another PR

Klakurka (changes requested):
- Let's stick with 'email' instead of 'e-mail' (/verify page). 
image
- Let's change 'Forgot Password' to 'Reset Password' to match the /reset-password URL. 
image

Klakurka (approved):

Klakurka:
Still needs to get into `master`.

Issue #596, opened 2023-08-12
Custom login UI


chedieck:
Enable custom login UI with
1. Email verification
2. Password trials limit
3. Forget password logic

PR #595, opened 2023-08-11
feat: custom UI login

chedieck:
Related to #

Description
---
Creates custom `/signin` and `/signup` pages, and show better errors when logging in.

**Forgot password** will be implemented in a next PR.

Test plan
---
**Need to run `yarn docker y` or enter the dev container and run `yarn`.**

1. Check that the sign in, sign up and log out all work.

2. All pages should work normally, but if not signed in, going to any URL besides `/signin` and `/signup` should redirect to `/signin`

3.  Inserting invalid email or password should give a clear error.

4. When logged in, going to `/signin` and `/signup` should redirect to the dashboard.

PR #597, opened 2023-08-12
[Issue #596] feat: email verification

chedieck:
Related to #596 

Depends on
---
- [x] #595 

Description
---
Adds email verification to the sign up process.

Test plan
---
1. Create an account, check that all steps up to verifying the email work well.
2. Check that unverified users get redirected to `/verify` when trying to access other pages.
3. Check that after verification `/verify` redirects to the dashboard.


Klakurka (approved):
Functions as expected.

Issue #549, opened 2023-06-22
Allow Remote Database Configuration


ScottMcDermid:
Allow option to connect to a remote database.
This might require an additional environment variables to be passed into containers.

Klakurka:
We'll need some extra config options:

- Host
- Username
- Password

ScottMcDermid:
We will eventually want an option to run a production environment, that doesn't spin up a docker db.

I'm envisioning something like

`make prod`

Which might read a `docker-compose.prod.yml` file, which omits the `db` container

PR #599, opened 2023-08-13
[Issue #549] feat: configurable db credentials

chedieck:
Related to #549 

Description
---
Database credentials are fully configurable using env files.

Test plan
---
Check the defaults in `.env`. These can be changed or overridden on `.env.local`. Resetting the containers should always work regardless, and the changed settings can be tested by entering the DB container (`yarn docker dbs`)  and running e.g.:

```
mariadb -u <main_db_user> -p<main_db_password> -D <main_db_name>
```
or for supertokens credentials:
```
mariadb -u <supertokens_db_user> -p<supertokens_db_password> -D supertokens
```
...substituing the `<variables>` above for the same value set on the `.env` or `.env.local` files.


ScottMcDermid:
Attempting to run `yarn docker dbs` gives me an error:

```
❯ yarn docker dbs
yarn run v1.22.19
$ ./scripts/docker-exec-shortcuts.sh dbs
./scripts/docker-exec-shortcuts.sh: line 11: .: .env: file not found
```

ScottMcDermid (changes requested):
`yarn docker dbs` gives me an error:
```
❯ yarn docker dbs
yarn run v1.22.19
$ ./scripts/docker-exec-shortcuts.sh dbs
./scripts/docker-exec-shortcuts.sh: line 11: .: .env: file not found
```

chedieck:
> 

Do you have a `.env` file? Because that's a commited file, so it should be present in the repo. `yarn docker dbs` works normally for me

ScottMcDermid:
I have an `.env` and `.env.local` file. Using `yarn docker dbs` on `master` works for me, but not this branch.

ScottMcDermid:
I can work around the convenient script, and call the command directly.
`doc exec db mariadb -u paybutton-user -p paybutton`

This allows me to connect with the updated username/password

ScottMcDermid:
I found something odd when I tried running parts of the `docker-exec-shortcuts.sh`

```
$ . .env
.: no such file or directory: .env
```

versus:

```
. ./.env
```

which works.

Also, David tried running both of these, and encountered no errors.

ScottMcDermid:
Ok.. Learned something new:

https://superuser.com/a/1499030

Seems to be a specific bash feature. I use zsh, which requires a path or else it searches `$PATH`.

Can we add the relative path in order to make this compatible with any shell?

```
. ./.env
. ./.env.local
```

chedieck:
> Ok.. Learned something new:
> 
> https://superuser.com/a/1499030

Very cool, didn't know that either. Fixed.

ScottMcDermid (approved):

PR #605, opened 2023-08-17
fix: account change password

chedieck:
Description
---
Fixes the non functional Account Change Password and displays supertokens default requirement to user on front end.

Test plan
---
Changing password should work, and require a password that is at least 8 chars long and has a letter and a number.


johnkuney (approved):
worked well for me for all cases

Issue #538, opened 2023-06-09
v1 Admin


Klakurka:
Admin page(s) that are only visible to accounts where isAdmin = 1.

- [ ] Admin dashboard / summary page
- [ ] Accounts detail page w/ option to make an account an admin (toggle)

There's a lot we can do here so maybe @johnkuney can come up with a concept and then we can build on it as we need to.

The target audience is really just us until somebody else out there hosts their own server and has some feedback on it.

For dev purposes, we can make the main `dev` account an admin.

Klakurka:
The rest of this can go into phase 3.

PR #607, opened 2023-08-18
[#538] feat: add admin dashboard view

chedieck:
Related to #538

Description
---
Enables the default user dashboard of supertokens and create an admin account for it. 

Test plan
---
Authenticate as the admin user at http://localhost:3000/api/auth/dashboard and try it out.


Klakurka (approved):
I wasn't aware they had a built-in user dashboard so we'll want to think about options to incorporate this into the PB-specific admin page(s) that we'll want.

Issue #613, opened 2023-08-22
Add payment triggers / reactions (Webhooks) section


Klakurka:
Allow attaching a set of "payment triggers" to a button, managed through the /button/xxx page.

When a payment is received on an address within the button, do what they've set it to do.

Options:
- Send an email (eventually we can expand this once #543 is done) to send emails to different email addresses within their organization. For now we'll limit it to just their own.
- Send a POST request to a URL, with a custom JSON object that they can set (maybe in a provided textarea) with placeholders for the various values.

Data:
- Button name
- Payment address
- Currency
- Amount
- Datetime
- Payload / BIP 70 data (eg. tx id or other context; think Coin Dance supporters section)

Could have:
 - Display warning on the page when the user tries to add a payment trigger to a button that already has one attached to it that's defined in another button that shares one or more addresses (as we allow addresses to be associated with multiple buttons). If they choose to ignore the warning, they may find that they have multiple triggers firing for a single payment.
 - Send a text message (same idea as email).

Klakurka:
For the UI, at the top of the section we'd have it say: "When a payment is received..." then we'd have maybe a checkbox toggle on each option to activate them.

chedieck:
- send txId as data too,
- only allow one trigger per address for now

Klakurka:
Let's also add a message signature that they can optionally check.

This will also require us to give everyone a private code / key which we can display on the same page used to generate the post requests (can put it below the form).

Klakurka:
BIP 70 payload, email + text notifications will be done later.

Klakurka:
Super cool.

PR #615, opened 2023-08-23
[#613] feat: payment triggers frontend skeleton

chedieck:
Related to #613

Description
---
Adds a skeleton for the frontend that sets up the triggers.

Test plan
---
For now, the submitted info will be logged in the backend, nothing else.
The following PRs will validate, commit the trigger to the the database and do the actual trigger activation upon payment

### DIFF
--- a/components/Paybutton/PaybuttonDetail.tsx
+++ b/components/Paybutton/PaybuttonDetail.tsx
@@ -1,4 +1,4 @@
-import React, { FunctionComponent } from 'react'
+import React from 'react'
 import { PaybuttonWithAddresses } from 'services/paybuttonService'
 import style from './paybutton.module.css'
 import EditButtonForm from './EditButtonForm'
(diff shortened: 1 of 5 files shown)


Klakurka (changes requested):
I'm not going to nitpick too much as John can revise it later but here's a few things we can do to start:

- Move whole section below the Transactions divs as it's primarily for advanced users.
- Change "Set triggers" to "When a Payment is Received..."
- Fix "Receive email" input + label (currently broken somehow)
- Change "Possible placeholders" to "Options".

The placeholder label in the textarea is nice.

chedieck:
>     * Move whole section below the Transactions divs as it's primarily for advanced users.

Done

>     * Change "Set triggers" to "When a Payment is Received..."

Done
 
>     * Fix "Receive email" input + label (currently broken somehow)

This is weird right now because in dark mode it doesn't show checked and the checkbox spans the whole width of the block, I tried to fix quickly but unsuccessfully, so I decided just to leave as it is, since John will rebuild the design and knows how to fix it, and the way it is right now is enough for testing
 
>     * Change "Possible placeholders" to "Options".

Done. Also, we are planning on letting them use variables, right? Like, writing down: {"buttonName": <buttonName>} would be parsed in our server to substitue <buttonName> with the actual button name



Finally, I have rebased this branch not to master but to `feat/paybutton-triggers`. We can go on approving until is ready, and then we merge that branch to master.

Klakurka (approved):
Yeah we'll allow custom values in the post request payload, just might need to somehow limit the total size so that's not easily abused.

PR #616, opened 2023-08-23
Feat/button generator

johnkuney:
Description
---
Adding a button generator module to the landing page.  Users can input a handful of fields and the code to render that button will be displayed below. 

Test plan
---
Run the app and go the button generator on the landing page. Make a custom button and try copying the different code blocks to a page to see if it renders properly

Remarks
---
I used the paybutton react module for the button preview, but its not working with eCash addresses as far as I can tell. Should it be, or will it be?

### DIFF
--- a/components/ButtonGenerator/CodeBlock.tsx
+++ b/components/ButtonGenerator/CodeBlock.tsx
@@ -1,0 +1,135 @@
+import { useEffect, useState } from 'react';
+import Image from 'next/image';
+import style from './button-generator.module.css';
+import CopyIcon from '/assets/copy.png';
+
+export default function CodeBlock({ button }): JSX.Element {
+  const [isCopied, setIsCopied] = useState(false);
+  const [codeType, setCodeType] = useState('HTML');
+  const [code, setCode] = useState({
+    html: '',
+    javascript: '',
+    react: '',
+  });
+
+  const codetypes = ['HTML', 'JavaScript', 'React'];
+
+  const handleCopyClick = async (): Promise<void> => {
+    const codeToCopy =
+      codeType === 'HTML'
+        ? code.html
+        : codeType === 'JavaScript'
+        ? code.javascript
+        : code.react;
+
+    try {
+      await navigator.clipboard.writeText(codeToCopy);
+      setIsCopied(true);
+      setTimeout(() => {
+        setIsCopied(false);
+      }, 1000); // Reset the "Copied" state after 2 seconds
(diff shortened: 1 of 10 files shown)


Klakurka (changes requested):
- As you said, eCash addresses don't seem to be working... but that's because the button generator isn't using the latest build since we haven't published it yet. Once that's published, we can double check before merging this.
- Have the TO field start empty and add a placeholder into the input. Maybe just 'Your address'.
- Hide the preview if the TO field is empty or invalid.
- Remove 'SAT' as an option in the CURRENCY dropdown.
- By default, hide/remove 'BCH' from the CURRENCY dropdown.
- When a BCH address is entered into the TO field and the CURRENCY is set to 'XEC', hide/remove 'XEC' from the CURRENCY dropdown, add 'BCH' back in and have it be selected. Same goes the other way for BCH -> XEC. If a fiat currency is selected when this change happens then just show/hide the appropriate XEC/BCH option in the bg but don't actually select it.
- Limit AMOUNT field to 13 digits (or 9999999999.99) and have it fix the issue of not working when somebody enters a number with one or more commas.
- Since the left side of inputs need more room, let's set that side to ~75% and the other side to ~25%, then set the whole control section to ~75% and the actual button preview to ~25%. This seems to work well on larger screens, but it'll need to also be adjusted to work well on mobile.

johnkuney:
Good edits, got those in. Except for the CSS not loading right away, still looking into that one

Klakurka (changes requested):
- CSS still not loading instantly like you said.
- If you enter an invalid address, it doesn't actually tell you and behaves like you haven't entered anything. If somebody accidentally copies an address but misses the the last 1-2 characters, they'll run into this and be confused if we don't show an error somewhere.
- When a BCH address is entered, change the theme colour to `#4bc846` (and back to the blue when an XEC address is entered -- only do this though if they have one of those 2 default colours chosen because otherwise they probably don't want their colour they chose messed with.

Klakurka (changes requested):
- In dark mode, the text in the colour picker isn't visible: 
image
- Let's go with the primary XEC colour for the blue default in the geneator: `#0074C2`. I have an idea to allow them to select "popular themes" in the generator that we can make up ourselves. I'll make a new task for that; it's pretty low prio.

Klakurka (changes requested):
Dark mode input font is still an issue for me.

Klakurka (approved):

PR #618, opened 2023-08-24
[#613] feat: payment triggers backend

chedieck:
Related to #613

Description
---
Adds backend to triggers.

Test plan
---
Try to create some triggers, making sure that:
- It requires URL to be a valid URL,
- It requires Post Data to be valid JSON,
- It doesn't require neither. But if one is set, the other one must be too.
- You can't create a trigger if one already exists for the button.
- You **can't** create a trigger if one already exists for a button of the **same user** that shares an address.
- You **can** create a trigger if one already exists for a button of **another user** that shares an address.

The triggers for the paybutton are being show as a strigified JSON. Of course, this is temporary and in a next PR the UI will be adjusted to edit the current trigger if one already exists for the button or not allowing the creation if one already exists for some associated address.


Klakurka (changes requested):
Right now the placeholder text isn't valid JSON (missing quotes & "<>").

We will need to eventually sort this out but to start let's at least have quotes in the placeholder text where necessary.

chedieck:
> Right now the placeholder text isn't valid JSON (missing quotes & "<>").
> 
> We will need to eventually sort this out but to start let's at least have quotes in the placeholder text where necessary.

Done

Klakurka (approved):
Ok good enough but obviously '<buttonName>' isn't valid json either.

PR #622, opened 2023-08-29
[#613] feat: payment triggers create or edit

chedieck:
Related to #613

Depends on
---
- [ ] #618

Description
---
Adds logic to update existing triggers instead of creating, if some trigger already exists for the button.

Test plan
---
I still left the WIP debugging line showing all the button triggers in the front end, to facilitate testing.
Be sure that:
1. There is at most 1 trigger.
2. You can change every field for the trigger and update only it.
3. You can set the postData and postURL to blank and this will make them be updated as blank
4. The fields get correctly filled with the database data.
5. The button says "Update" or "Create" depending if there is or there isn't a trigger already, respectively
6. Updating a trigger to a blank state (no receive emails, no postData, no postURL) will actually delete it.

Remarks
---
Deleting triggers right now is only possible through updating it to an innocuous trigger. This can be improved later with a "Delete trigger" button on the other side of the "Update" button.


Klakurka (changes requested):
Can we have the form not disappear and require a refresh to show on submit?

PR #624, opened 2023-08-29
[#613]  feat: payment triggers post data replacement

chedieck:
Related to #613 

Depends on
---
- [x] #623 

Description
---
Implements variable substitution on trigger data.

Test plan
---
Example of postData one could set on the trigger:
```
{
"A": {"name": <buttonName>, "am": <amount>},
"b": {"c": <currency>, "pa": <paymentAddress>, "other": "txId", "f": <timestamp>},
"hash": <txId>
}
```

...after the trigger execution this should get replaced with the appropriate data in each field. Of course, this makes so that setting the trigger data to e.g.`{"a": <txId>}` should be valid JSON, whereas `{"a": <transactionId>}`  should not, since this variable is not expected.


Klakurka (changes requested):
If somebody enters: {"btnName":  "_<buttonName>_"} then it'll tell them that it's invalid JSON. Perhaps we can either go with a different error message or "fix" their JSON for them by removing the quotes around the variable.

chedieck:
> If somebody enters: `{"btnName": "<buttonName>"}` then it'll tell them that it's invalid JSON. Perhaps we can either go with a different error message or "fix" their JSON for them by removing the quotes around the variable.

This is an interesting point. I've updated it to give a different error message.

Overall, since we're doing string replacement there are some weirdness to deal with. In this particular case, I think it should indeed break, it looks cleaner to me if those variables act as if they either are properly used or shouldn't be used at all.

In reality, would be better, in the example given, if ` "<buttonName>"` wasn't even replaced. However, dealing with a string that can be formatted in any manner (since is user input), this would be a bit complex to implement: at least would require checking for quotes starting & ending texts, to see if the variable is outside or inside quotes, then replacing them or not based on that. I could try implementing that if you think it's worthy.

Klakurka (approved):
Agreed.

As people use it (and run into issues), we can refine this.

PR #627, opened 2023-08-31
[#613] feat: payment trigger logs

chedieck:
Related to #613 

Description
---
Logs every trigger execution, regardless of if it failed or not.
Also makes so that the URL defined by the user is the one used.

Test plan
---
1. Add the line `  "triggerPOSTTimeout": 3000` to your `paybutton-config.json`. See the example in `config/example-config.json`. This will make the trigger post wait at most 3 seconds for an answer.

2. Create a trigger, for example, with URL `https://httpbin.org/post` and data `{ "myThing": <amount>, "other": <buttonName> }`.
3. Send a payment to that button
4.  Enter the docker db container  (`yarn docker db`) and run `SELECT * FROM TriggerLog;`, and see if it was all logged alright.

5. Then change that trigger URL, to e.g: `https://sitethatdefinitelydoesnotexist.com`, send another payment to some of the button addresses and see if the error was logged correctly too.

### DIFF
--- a/config/example-config.json
+++ b/config/example-config.json
@@ -16,5 +16,6 @@
   },
   "networksUnderMaintenance": {
     "bitcoincash": true
-  }
+  },
+  "triggerPOSTTimeout": 3000
 }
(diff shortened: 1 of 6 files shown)


Klakurka (changes requested):
- I missed this in the previous review but we want the form to continue showing the form after it's submitted instead: 
image

Everything else looks good.

PR #632, opened 2023-09-01
[#613] feat: payment triggers private key

chedieck:
Related to #613

Depends on
---
- [x] #631

Description
---
1. Adds a master secret key to the app and secret keys to each user.
2. Allows another variable on the trigger, `<postHash>`. This is the hash of the trigerring TX timestamp, using the user's secret key

Test plan
---

1. Add to your `.env.local` a `MASTER_SECRET_KEY=SOMESECRETKEY`. You can generate one with e.g. `openssl rand -base64 32`.
2. Create a trigger, e.g:
   postURL: `https://httpbin.org/post`
   postData:`{"Y": <postHash>, "t": <timestamp>, "id": <txId>}`

4. Then, after making a transaction, check in the db `SELECT * FROM TriggerLog;` and see something like: 
  `\"Y\":\"EXAMPLE_POST_HASH\", \"t\": 1693590786}`
in the postedData info.

5. Now go to your user Account page and get the secret key, e.g. `EXAMPLE_SECRET_KEY`

6. Finally, a user can check on his own that this post request is legit by executing e.g. in a bash command line:
```
echo -n "1693590786" | openssl dgst -sha256 -hmac "EXAMPLE_SECRET_KEY"
```

... and seeing if it's exactly `EXAMPLE_POST_HASH`

PS:  The command above can be executed inside the node container with `openssl1.1` instead of `openssl`


Klakurka (changes requested):
- Have the 'Update Password' button on the Account page be above the secret key section. John can touch it up later.
- We need to have at least a small note somewhere about how to verify the hash / signature. Perhaps we can have that info next to the secret in the Account page with a hyperlink next to `<postHash>`.
- Really what we're trying to do here is signing, not just hashing so do we need to remove the secret out of the hash add an encryption phase w/ the secret (`<signature>`)?
- I think we can go with `<hash>` instead of `<postHash>`.
- I think we can go with `<address>` instead of `<paymentAddress>`. If we need to enter a URL or something, we can call it something else.

Klakurka:
https://crypto.stackexchange.com/questions/5646/what-are-the-differences-between-a-digital-signature-a-mac-and-a-hash

services/triggerService.ts
```diff
@@ -187,16 +189,46 @@ export async function executeAddressTriggers (broadcastTxData: BroadcastTxData):
...
+  timestamp: number
+  txId: string
+  buttonName: string
+  paymentAddress: string
+  postHash: string
+}
+
+async function hashPostData (userId: string, postDataParameters: PostDataParameters): Promise<string> {
+  const hmac = crypto.createHmac('sha256', await getUserSecretKey(userId))
+  hmac.update(`${postDataParameters.timestamp}}`)
```
Klakurka, line 222: We want to hash all of the data we're sending so let's split this out into two steps:

1. Hash all of the data (however you want to package it) using vanilla SHA256.
2. Hash / encrypt / sign the hash with SHA256 again but this time with the user secret.

This way we get the benefit of it being easy to verify as it's still just one string but also are hashing all of the data since there's a chance of duplicates if 2 txs come in at the same time.

chedieck: Good point, I addressed it hashing all ` amount, currency, paymentAddress, timestamp, txId ` but not `buttonName` because that could change and lead to problems

chedieck:
>  Have the 'Update Password' button on the Account page be above the secret key section. John can touch it up later.

Done

>   We need to have at least a small note somewhere about how to verify the hash / signature. Perhaps we can have that info next to the secret in the Account page with a hyperlink next to `<postHash>`.

Done, is not the best thing right now but I think John can touch it up too.


>   I think we can go with `<hash>` instead of `<postHash>`.

Done, though maybe `<messageHash>` or something would be better since `<hash>` is sometimes used to refer to the `txId`

>    I think we can go with `<address>` instead of `<paymentAddress>`. If we need to enter a URL or something, we can call it something else.

Done

Klakurka (changes requested):
We need the `hmac` to be available in the post data for them to verify it. What about calling it `<signature>`?

chedieck:
> We need the `hmac` to be available in the post data for them to verify it. What about calling it `<signature>`?

You mean the simple hash done before hashing with their secret key? They could do that themselves, maybe sending both of them facilitates, I would call one `< hash>` and the other `<hashHMAC>`, the latter being the one that uses their secret key.

Klakurka:
> > We need the `hmac` to be available in the post data for them to verify it. What about calling it `<signature>`?
> 
> You mean the simple hash done before hashing with their secret key? They could do that themselves, maybe sending both of them facilitates, I would call one `< hash>` and the other `<hashHMAC>`, the latter being the one that uses their secret key.

Yeah. My only issue is that "hashHMAC" doesn't make it clear what it's for which is why I was wanting to go with signature or something (even though I get that they're different).

Klakurka:
It looks like we need to fix up the POST data. Here's how it looks in PHP:

```
Array
 (   [{"address":"ecash:xxx","currency":"XEC","amount":100,"tx":"abc123","custom":"blarg"}] =>
 )
```

Notice how the data is in there as the key for an array, with an empty value.

Here's what we're expecting:

```
Array
 (
     [address] => "ecash:xxx"
     [currency] => "XEC"
     [amount] => 100
     [tx] => "abc123"
     [custom] => "blarg"
 )
```

chedieck:
> Yeah. My only issue is that "hashHMAC" doesn't make it clear what it's for which is why I was wanting to go with signature or something (even though I get that they're different).


Renamed to `hmac`



> Here's what we're expecting:

Should be fixed now.

Klakurka (changes requested):
Doesn't build for me.

Klakurka (approved):

PR #633, opened 2023-09-04
fix: subscribe db addresses

chedieck:
Description
---
Fix the problem where resetting the prod app would lose the subscriptions of addresses already in the DB (since the DB is not being reset)

Test plan
---
- Prod is already running this, so receiving live info about txs of any address that was subscribed before (or not) should work now.
- https://api.paybutton.org/status should list  all subscribed addresses, and this should be all `ecash` addresses in the DB.

One other test would be resetting prod and still having all addresses that were subscribed before subscribed now.

 
Remarks
---
This PR brings a variety of technical changes, so I'll list them here:
-  Jobs don't subscribe addresses, only sync their txs
- Fixes `yarn docker` commands to access remote DB
- Add `yarn docker` command to create dump of DB (this was unstaged on the remote server, so I brought and commited it) 
- The chronik instance initialization also subscribe addresses

### DIFF
--- a/jobs/initJobs.ts
+++ b/jobs/initJobs.ts
@@ -2,7 +2,7 @@ import { XEC_NETWORK_ID, BCH_NETWORK_ID, CURRENT_PRICE_REPEAT_DELAY, SYNC_TXS_JO
 import { Queue, FlowProducer, FlowJob } from 'bullmq'
 import { redisBullMQ } from 'redis/clientInstance'
 import {
-  syncAndSubscribeAllAddressTransactionsForNetworkWorker,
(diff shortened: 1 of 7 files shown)


Klakurka (changes requested):
Test worked but failing some checks.

Up to you if you think we should leave those console logs in there.

chedieck:
> Test worked but failing some checks.
> 
> Up to you if you think we should leave those console logs in there.

Fixed tests, I think is good to leave them for now (the ones I added should only appear once in the beginning after starting up the app)

Klakurka (approved):

PR #635, opened 2023-09-04
feat: secure status api

chedieck:
Depends on
---
- [x] #633 

Description
---
Makes `/status` endpoint secure by requesting WS_AUTH_KEY as a request paremeter.

Test plan
---
Accessing localhost:3000/status should return "unauthorised", in order to see the subscribed addresses one would have to send e.g. http://localhost:3000/status?secret=<WS_AUTH_KEY>  where `WS_AUTH_KEY` should be set on `.env.local`.


Klakurka (changes requested):
I actually get this error instead:

image

Klakurka (dismissed):
Just needed to login once to get it to compile.

PR #636, opened 2023-09-05
[#613] feat: payment triggers

chedieck:
Finishes to #613 

Description
---
Proxy branch with all approved changes for #613


Klakurka (changes requested):
Checks are failing.

Everything works as expected though.

PR #644, opened 2023-09-05
UI: Button generator code format

johnkuney:
Description
---
Adding a package to hightlight the code snippets on the button generator, as well as some logic to remove default or empty values from the code snippets

Test plan
---
Run the app and check out the button generator


Klakurka (approved):
Looks great!

PR #645, opened 2023-09-06
fix: bring back secret on account page

chedieck:
Description
---
I lost some changes in a failed rebase (still don't know how exactly), this brings back the changes about the secret keys on the Accounts page.

Test plan
---

### DIFF
--- a/pages/account/account.module.css
+++ b/pages/account/account.module.css
@@ -20,12 +20,25 @@
   display: flex;
   align-self: baseline;
   margin-top: 10px;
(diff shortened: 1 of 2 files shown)


Klakurka (approved):
I'm going to approve this but let's simplify this a bit for people in a follow-up branch.

We can provide them the option to take the <hash> and then they can do the check in 1 step instead of having to build the hash as well.

We can also tell them how the original hash is generated so they can go the extra mile if they want to.

PR #651, opened 2023-09-08
UI: Handle default values on the button generator

johnkuney:
Related to #

Description
---
A few tweaks to handle the default value treatment better. Now all values should not show in the code to copy, if they are default values. Also added a little reset button. Helps with testing and maybe useful for users 

Test plan
---
Run the app and try the button generator. Try adding and removing values


components/ButtonGenerator/CodeBlock.tsx
```diff
@@ -70,6 +70,51 @@ export default function CodeBlock({ button }): JSX.Element {
...
+
+  const checkTheme = (value, type) => {
+    if (
+      (value.palette.primary === primaryXEC &&
+        value.palette.secondary === secondaryXEC &&
+        value.palette.tertiary === tertiaryXEC) ||
+      (value.palette.primary === primaryBCH &&
+        value.palette.secondary === secondaryBCH &&
+        value.palette.tertiary === tertiaryBCH)
+    ) {
```
Klakurka, line 101: How about re-writing this to support N number of currencies?

These colours should be in a const or conf file, ideally.

components/ButtonGenerator/CodeBlock.tsx
```diff
@@ -70,6 +70,51 @@ export default function CodeBlock({ button }): JSX.Element {
     }
   };
 
+  const checkCurrency = (value, type) => {
```
Klakurka, line 73: Feels like we could come up with a better function name here but I don't have any suggestions.

Klakurka (changes requested):
Functionality looks good.

johnkuney:
Good points, made some tweaks

Klakurka (approved):
I think we can improve this over time but it's good enough for now IMO.

Issue #661, opened 2023-09-12
IFP address sync failure


Klakurka:
paybutton-dev  | 2|next  | 2023-09-12T02:03:04: Invalid `prisma.address.findMany()` invocation:
paybutton-dev  | 2|next  | 2023-09-12T02:03:04: 
paybutton-dev  | 2|next  | 2023-09-12T02:03:04: 
paybutton-dev  | 2|next  | 2023-09-12T02:03:04: Timed out fetching a new connection from the connection pool. More info: http://pris.ly/d/connection-pool (Current connection pool timeout: 10, connection limit: 5)
paybutton-dev  | 2|next  | 2023-09-12T02:03:04:     at pn.handleRequestError (/home/node/src/node_modules/@prisma/client/runtime/library.js:176:6477)
paybutton-dev  | 2|next  | 2023-09-12T02:03:04:     at pn.handleAndLogRequestError (/home/node/src/node_modules/@prisma/client/runtime/library.js:176:5907)
paybutton-dev  | 2|next  | 2023-09-12T02:03:04:     at pn.request (/home/node/src/node_modules/@prisma/client/runtime/library.js:176:5786)
paybutton-dev  | 2|next  | 2023-09-12T02:03:04:     at async t._request (/home/node/src/node_modules/@prisma/client/runtime/library.js:179:10484)
paybutton-dev  | 2|next  | 2023-09-12T02:03:04:     at async fetchAllUserAddresses (/home/node/src/.next/server/chunks/4784.js:496:10)
paybutton-dev  | 2|next  | 2023-09-12T02:03:04:     at async getUserUncachedAddresses (/home/node/src/.next/server/chunks/4784.js:187:21)
paybutton-dev  | 2|next  | 2023-09-12T02:03:04:     at async getPaymentList (/home/node/src/.next/server/pages/api/dashboard.js:230:25)
paybutton-dev  | 2|next  | 2023-09-12T02:03:04:     at async getUserDashboardData (/home/node/src/.next/server/pages/api/dashboard.js:239:23)
paybutton-dev  | 2|next  | 2023-09-12T02:03:04:     at async __WEBPACK_DEFAULT_EXPORT__ (/home/node/src/.next/server/pages/api/dashboard.js:276:26)
paybutton-dev  | 2|next  | 2023-09-12T02:03:04:     at async Object.apiResolver (/home/node/src/node_modules/next/dist/server/api-utils/node.js:372:9) {
[...]

PR #663, opened 2023-09-12
Fix: block IFP addresses

ScottMcDermid:
Related to #661

Description
---
Prevents IFP addresses from being added or updated

Test plan
---
1. Try adding an IFP address to a paybutton - you should encounter an error
2. Try updating a paybutton to use an IFP address - you should encounter an error


Klakurka (approved):
We'll remove this once we can handle it.

PR #674, opened 2023-09-14
[#673] fix: update timestamp when tx is confirmed

chedieck:
Related to  #673

Description
---
Tx confirmation now also updates the timestamp.

Test plan
---
Necessary to reset the containers.
Create a button with one of your addresses and send some XEC to it. See the timestamp when it's unconfirmed and check it again when confirmed, it should have updated (unlike in master)

### DIFF
--- a/services/transactionService.ts
+++ b/services/transactionService.ts
@@ -134,7 +134,8 @@ export async function createTransaction (
       }
     },
     update: {
-      confirmed: transactionData.confirmed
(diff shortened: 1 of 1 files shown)


Klakurka (approved):
I'm ok with this as it means that we have a consistent + reliable column to use. `createdAt` still would be effectively the original 'received' time assuming all is operating normally.

Issue #679, opened 2023-09-19
Allow entering a 'Description' and 'Website' for each PayButton


Klakurka:
The URL and description would show up on the button/xxx page. The URL would be a clickable hyperlink (somewhere tasteful; open in new browser tab) near the description.

These fields would be optional and available at both button creation and on edit.

PR #680, opened 2023-09-19
[#679] feat: button url and description

chedieck:
Related to #679

Description
---
Adds fields `url` and `description` to Paybutton.

Test plan
---
First, execute the migrations with `yarn docker m`; or `yarn prisma migrate dev` inside the node container.
Then, make sure you can create & edit paybuttons such that these fields work as expected, which means:
- They are optional,
- URL must be valid,
- When editing a Paybutton, setting those field to blank will save them as blank.
- When adding a URL, `example.com` is valid and saved as `https://example.com/`. `example` would be invalid.

<!-- 
Remarks
---
- I added another paybutton with one of my addresses to the initial seeding, since I often needed to add it manually to test the behavior of a transaction. This should not afect prod in any way, since we use a fixed DB there.

- I also added a `*` next to `Name` and `Addresses` on the button creation form, to indicate they are required.


Klakurka (changes requested):
- Have the `URL` and `Description` form fields below `Addresses` and `Wallet` as they're optional.
- `Wallet` field is missing from button edit.
- On button create, the `Name` field has a label of `Name*` but on edit it's `Button Name`. Let's go with `Name*` for both.
- Change `URL` to `Website` so that people aren't worried about making sure the URL is exact (sometimes it's not possible to know or it may be on multiple pages of the same website)
- Let's add some placeholder text to act as helpers:
  - Name: `The unique name of your button`
  - Addresses: `The address(es) you will be receiving payments to`; only do this for create as edit uses the placeholder to show the previously-entered address(es).
  - Website: `Where will your button be used? (optional)`
  - Description: `More information about your button (optional)`
- I'm getting a bunch of console errors -- maybe check that you're seeing the same thing?

chedieck:
> Have the `URL` and `Description` form fields below `Addresses` and `Wallet` as they're optional.

Done
>   `Wallet` field is missing from button edit.

Out of scope of this PR, created a new issue #681
>   On button create, the `Name` field has a label of `Name*` but on edit it's `Button Name`. Let's go with `Name*` for both.

Done 
>  Change `Addresses` to `Addresses*` to match the `Name*` field label. Maybe do this for `Wallet` too even though it forces you to enter something w/ the dropdown.

 Done
>  Change `URL` to `Website` so that people aren't worried about making sure the URL is exact (sometimes it's not possible to know or it may be on multiple pages of the same website); do this on both the forms and on the main button page. John can improve it later.
 
Done
>   Let's add some placeholder text to act as helpers:
 
Done, added to the button creation form. In the edit form the placeholder is the old values, tell me if I should change.
>   I'm getting a bunch of console errors -- maybe check that you're seeing the same thing?
 
 Is this `Warning: Each child in a list should have a unique "key" prop.`? If so, it should be fixed now.
 
>   I managed to kinda break the URL checking with: `p://abcdefg123.com/asdfasdf` (bad URL prefix). There's probably a good regex out there for this.

This is actually a valid URL. But since I think that the idea is that we always use `http` or `https` protocols on this field, I have added this requirement.

johnkuney:
Looks good to me. Tested and worked well. I just pushed some small tweaks hope that cool
- adjusted the form spacing a tad
- added some conditional rendering for the optional fields
- made the website a link
- added some conditional placeholder text in the edit button form

johnkuney (approved):

Klakurka (approved):

PR #689, opened 2023-09-25
[#687] fix: disallow empty string for post data

### LINKED ISSUE #687: TriggerLog DB error (data is not valid JSON)
Found this error on the DB (`TriggerLog`):

`{"errorName":"Error","errorMessage":"Data is not valid JSON.","errorStack":"Error: Data is not valid JSON.\n    at parseTriggerPostData (/home/node/src/.next/server/chunks/9895.js:431:11)\n    at postDataForTrigger (/home/node/src/.next/server/chunks/1488.js:2828:118)\n    at /home/node/src/.next/server/chunks/1488.js:2801:11\n    at async Promise.all (index 0)\n    at async executeAddressTriggers (/home/node/src/.next/server/chunks/1488.js:2791:3)\n    at async /home/node/src/.next/server/chunks/1488.js:1060:15\n    at async Promise.all (index 0)\n    at async ChronikBlockchainClient.processWsMessage (/home/node/src/.next/server/chunks/1488.js:1038:7)","triggerPostData":"","triggerPostURL":""}`

It shouldn't be possible to send invalid JSON but perhaps a Chronik error caused it?

chedieck:
Related to #687

Description
---
Fixes #687. Don't allow the creation of empty triggers, which were resulting in error logs since there is nothing to execute.

Test plan
---
Try to create a trigger without setting an URL and a postData, should not be possible.

Remarks
---
This should be changed when we implementing receiving emails as a functionality of the trigger, but for now makes no sense allowing to create a trigger that does nothing.

### DIFF
--- a/constants/index.ts
+++ b/constants/index.ts
@@ -68,7 +68,7 @@ export const RESPONSE_MESSAGES = {
   WRONG_PASSWORD_400: { statusCode: 400, message: 'Wrong password.' },
   INVALID_URL_400: { statusCode: 400, message: 'Invalid URL.' },
   INVALID_WEBSITE_URL_400: { statusCode: 400, message: 'Invalid website URL.' },
-  POST_URL_AND_DATA_MUST_BE_SET_TOGETHER_400: { statusCode: 400, message: 'URL and post data must neither or both be set.' },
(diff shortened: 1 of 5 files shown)


Klakurka:
Tests failing.

Klakurka (approved):

Issue #652, opened 2023-09-08
'Advanced' Button Generator options


Klakurka:
Add a new 'Advanced' button next to 'Reset' that expands the form to cover all of the other options available.

Up to you if you want to have the button be a toggle between 'Advanced' and 'Simple' or if it's just one way that requires a refresh to get back to the current 'Simple' form options.

Klakurka:
We really do want `random-satoshis` to be in there (either in simple or advanced) since there will certainly be people who want to turn it off but won't see the option.

Klakurka:
I think 'Widget' (toggle) should be on the 'Simple' side.

Klakurka:
I had a chat

PR #691, opened 2023-09-26
[#652]UI: Button generator advanced options

johnkuney:
Related to #652

Description
---
Adding the advanced options to the button generator, as well as a toggle for the widget

Test plan
---
Go to the button generator and try the new options. Check the code output looks correct and the preview works


Klakurka (changes requested):
Can't seem to build this branch.

image

Klakurka:
Same thing after a git prune.

johnkuney:
hmm maybe was behind some commits, just did a rebase

Klakurka (changes requested):
Still getting the same error.

components/ButtonGenerator/CodeBlock.tsx
```diff
@@ -115,12 +125,67 @@ export default function CodeBlock({ button }): JSX.Element {
     }
   };
 
+  const checkReactValue = (value, attribute) => {
+    if (value !== undefined && value !== '') {
+      if (attribute === 'currency') {
+        if (['USD', 'CAD', 'EUR', 'GBP', 'AUD'].includes(value)) {
```
Klakurka, line 131: I'd rather this come from a central list somewhere. Maybe from the DB. It'll be one of those things we miss when we add/remove currencies. In fact EUR, GPB, and AUD aren't even supported rn.

johnkuney:
Will revisit this with a more incremental approach since we are having issues with the branch

Issue #685, opened 2023-09-20
Add 'Payments leaderboard' chart/table to dashboard


Klakurka:
List all buttons, sorted by payments (total amount; DESC).

PR #692, opened 2023-09-27
[#685] feat: buttons leaderboard

chedieck:
Related to #685

Description
---
Implements #685:

image

Test plan
---
After loading the branch, reset the cache with `yarn docker cmr` and check if the leaderboard appears normally.

### DIFF
--- a/components/Dashboard/Leaderboard.tsx
+++ b/components/Dashboard/Leaderboard.tsx
@@ -1,0 +1,45 @@
+import TableContainer from 'components/TableContainer'
+import { useMemo } from 'react'
+import { ButtonData } from 'redis/dashboardCache'
+import style from '../Transaction/transaction.module.css'
+
+interface IProps {
+  buttons: ButtonData[]
+}
+
+export default ({ buttons }: IProps): JSX.Element => {
+  const columns = useMemo(
+    () => [
+      {
+        Header: 'Name',
+        accessor: 'displayData.name',
+        Cell: (cellProps) => {
+          return <div >{cellProps.cell.value}</div>
+        }
+      },
+      {
+        Header: () => (<div style={{ textAlign: 'right' }}>Total Payments</div>),
+        accessor: 'total.payments',
+        Cell: (cellProps) => {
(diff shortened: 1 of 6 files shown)


Klakurka (changes requested):
- Can we have 'Total Revenue' match the charts in that they're using the user's selected fiat currency to measure it? This would be the daily $ summed up (same as if they had manually looked at the revenue chart and added up the $ from each day). Right now I'm not even sure what number it's using because there's no '$' symbol and the decimals should be limited to 2.
- Set decimal rounding to 2.
- On the dev account, 'Foo XEC' has only 1 transaction but the leaderboard shows 2 payments.

Klakurka:
I forgot to reset the cache so I'll do that when I can test next.

chedieck:
> * Can we have 'Total Revenue' match the charts in that they're using the user's selected fiat currency to measure it? This would be the daily $ summed up (same as if they had manually looked at the revenue chart and added up the $ from each day). Right now I'm not even sure what number it's using because there's no '$' symbol and the decimals should be limited to 2.

> Set decimal rounding to 2.

I changed it so it uses the same formatting function as the other places in the same page: it uses 2 digits of precision unless the quantity is very small, in which case it just shows the first significant digit.
 

>   On the dev account, 'Foo XEC' has only 1 transaction but the leaderboard shows 2 payments.

I couldn't reproduce that: in my leaderboard, as seen in the printscreen, `Foo XEC` shows as having only one payment.

chedieck:
Observation:

image


Here we can see that the total payments according to the top right chart is 39, but adding up the payments from the leaderboard gives us 40.

This happens because two buttons share the same address, so this single payment appears twice in the leaderboard: as the one payment for **"Foo XEC"** and for **"Paybutton XEC & Coindance BCH"**

Klakurka (changes requested):
Other changes look good but I still have this issue:

image

(total payments is wrong somehow)

Klakurka (approved):
Looks like it's probably #695 which we can do next.

PR #693, opened 2023-09-28
feat: only show page selector if > 1 pages

chedieck:
Description
---

Minor change.

Table page selector:
image

... only appears if total pages is greater than 1.

Test plan
---
Check some tables (payments, button transactions, button leaaderboard if #692 is merged) and see if the page selector appears if and only if there is more than only the first page.

### DIFF
--- a/components/TableContainer.tsx
+++ b/components/TableContainer.tsx
@@ -69,42 +69,44 @@ const TableContainer = ({ columns, data }): JSX.Element => {
       </table>
     </div>
 
-<div className='table-navigation-ctn'>
-    <button
-      onClick={() => gotoPage(0)}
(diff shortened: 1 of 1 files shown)


Klakurka (changes requested):
I think if there's more than one address on a button, both address tx sections in /button/xxx should change when you change 'Show X'. Right now only one changes and it's a bit weird.

chedieck:
> I think if there's more than one address on a button, both address tx sections in /button/xxx should change when you change 'Show X'. Right now only one changes and it's a bit weird.

Hmm, I see, it's because they are two different different tables indeed. But in any case, I think this is out of scope for this branch, we could open an issue for it.

Klakurka (approved):
#696

Issue #634, opened 2023-09-04
Move /status info into admin page


Klakurka:
Right now it's available to the public which is not what we want.

Klakurka:
We could also keep `/status` to be available to the public but change what it actually shows.

Example:

PayButton Server - UP
XEC Connection (Chronik) - UP
BCH Connection (BCHD) - DOWN

PR #700, opened 2023-10-03
[#538] feat: admin page

chedieck:
Related to #538, #634

Description
---
Creates a very basic admin page that links to the supertokens admin dashboard & shows chronik info that was previously shown in `/status` :
image

The PR was made considering @johnkuney  

Test plan
---

#### Log into any other non-admin account and verify that:
- "Admin" menu item is not in the sidebar,
- Trying to go to `/admin` manually will actually redirect you to `/dashboard`
- `/chronikStatus` gives `unauthorised` response.
 

Remarks
---
- This is only supposed to be a rudimentary first version of the admin page.
- Would be great if @johnkuney could take over and prettify it generally. In particular, I left two `TODO`s in the code: one for creating a new icon for the admin sidebar (I just reused the account one) and another to change its color (I just hardcoded FF0 yellow). This last change may not be important/desired, I just added because I thought it is good to log into an admin account and have an immediate visual signal that you are an admin.


Klakurka (changes requested):
- The nav is now showing up on the signin page for some reason (same with signup): 
image
- We can keep the text colour the same as the others (in the nav); right now it's yellow.
- Have 'Supertokens Admin Dashboard' open in a new browser tab.
- Maybe we could use a small subtitle on each Subscription section that makes it easier to understand what each list actually represents?

chedieck:
> The nav is now showing up on the signin page for some reason (same with signup):

Fixed.
>  We can keep the text colour the same as the others (in the nav); right now it's yellow.

Removed the yellow. I still think would be good to have a more contrasted visual indicator that the user is an admin, upon login.
>  Have 'Supertokens Admin Dashboard' open in a new browser tab.

Done.
>  Maybe we could use a small subtitle on each Subscription section that makes it easier to understand what each list actually represents?

I actually removed the display of the two lists and only show the second one if they are different, with a warning message.

Klakurka:
The thing is about admins is that it's such a niche feature that I really don't care about highlighting it.

The target audience for the moment is so small and even after others are running it themselves, it's still going to be only people who actually know what they're doing that have this feature available to them.

Klakurka (approved):

PR #701, opened 2023-10-05
fix: landing page should redirect to dashboard if logged in

chedieck:
Description
---
When logged in, manually navigating to `/` would show the landing page with the Sidebar.

The PR fixes this and manually navigating to `/` redirects to `/dashboard`.

Test plan
---
After logging in, go to http://localhost:3000/ and make sure you get redirect to the dashboard.

 
Remarks
---

About the blinking landing page before, this is an issue with all `AUTHORIZED_UNLOGGED_URLS`. Currently, those are:
- `/`,
- `/signin`,
- `/signup`,
- `/reset-password`,
- `/auth/reset-password`

... and this is an issue with the current implementation for this redirect logic, that could be solved in a different PR.

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -16,6 +16,7 @@ if (typeof window !== 'undefined') {
 }
 
 const AUTHORIZED_UNLOGGED_URLS = [
+  '/',
   '/signin',
   '/signup',
   '/reset-password',


Klakurka (changes requested):
We intentionally don't want to do a redirect when logged in so that somebody who's logged in can still work with the button generator or otherwise look at the landing page for whatever reason.

The sidebar wasn't showing up when logged in until this last branch was merged (I'm pretty sure).

chedieck:
> We intentionally don't want to do a redirect when logged in so that somebody who's logged in can still work with the button generator or otherwise look at the 

Alright, makes sense
 
> The sidebar wasn't showing up when logged in until this last branch was merged (I'm pretty sure).
I think it did, actually, because it was when working on the sidebar issue in that branch that I noticed this and checked the behavior on master; but I can change that in another branch (I'll close this one)

PR #702, opened 2023-10-06
chore: don't spin DB container on prod

chedieck:
Description
---
On production, we should not spin up the container, since we use a remote db.

Test plan
---
Hard to test it thoroughly, one could manually create a remote DB and then configure `.env.local` with its credentials and run `make prod`. But running `make prod` without doing that should not create the 4 containers, only 3, and should then show an error about not being able to find the DB to connect.


Klakurka:
Works in prod so going to just put this through.

Issue #706, opened 2023-10-06
Button leaderboard show filter text


chedieck:
Make clear that the values displayed are for e.g. "30 Days" if the 1M filter is selected

PR #707, opened 2023-10-06
[#706] feat: filter info on leaderboard

chedieck:
Related to #706

Description
---
Show filter info in leaderboard and remove redundant info.

Test plan
---
Make sure that the filter info is being shown in the leaderboard (**red**), and that the total revenue info (**yellow**) is not being redundantly shown in the **blue** area. 
image

### DIFF
--- a/components/Dashboard/Leaderboard.tsx
+++ b/components/Dashboard/Leaderboard.tsx
@@ -11,9 +11,10 @@ import { formatQuoteValue } from 'utils'
 
 interface IProps {
   buttons: PaymentDataByButton
+  totalString: string
 }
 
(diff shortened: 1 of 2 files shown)


Klakurka (changes requested):
Can you either remove the "active networks" column (easy) or have it only show it the user has multiple networks in use on their account like we do on the buttons page labels (hard)?

chedieck:
> Can you either remove the "active networks" column (easy) or have it only show it the user has multiple networks in use on their account like we do on the buttons page labels (hard)?

I can just remove it since we're not supporting BCH right now, but I think is out of scope here.

Klakurka (approved):

Issue #601, opened 2023-08-16
Have login status change the landing page nav menu


Klakurka:
If you're logged in, navigating to the landing page now redirects you back to the dashboard. We should allow people to visit the dashboard if they want to even if they're logged in. Bonus points would be to show 'Dashboard' and 'Sign Out' instead of 'Sign Up' and 'Sign In' in the nav but we can v2 this if it's too much trouble.

Klakurka:
We could also have the logo in the upper left take you to the landing page once this is in.

chedieck:
1. Create link to Landing Page on the Sidebar,
2. Removes the Sidebar from the landing page,
3. Change "Sign In" and "Sign Up" for "'Dashboard" and " "Sign out" in the Landing Page, if logged in.

PR #709, opened 2023-10-06
[#601] feat: better logged in landing page navigation

chedieck:
Related to #601

Description
---
Better navigation to the Landing Page when logged in.

Test plan
---
[As commented](https://github.com/PayButton/paybutton-server/issues/601#issuecomment-1751412650)  in #601,  make sure  all work fine.


Klakurka (approved):
Looks good.

It would be nice to maybe highlight the Dashboard nav button somehow but this works. If @johnkuney has any ideas later, we can try that.

Issue #713, opened 2023-10-09
Have PayButton logo be the same size on the landing page and after logged in


Klakurka:
Also same position on mobile.

Issue #655, opened 2023-09-12
Sign Up button is missing from footer on light theme


Klakurka:
![image](https://github.com/PayButton/paybutton-server/assets/4713204/28b48026-75b3-4fd1-afe6-12d0d4957a1c)

PR #719, opened 2023-10-16
[#655, #713]UI: footer button and logo sizing

johnkuney:
Related to #655 #713

Description
---
Fixing the issue of the sign up button not showing in the footer, and also matching the new behavior of the navbar so dashboard and Sign out show in the footer when logged in. Also sniped the logo size to match in the navbar, footer, and dashboard views

Test plan
---
Run the app and check the footer links look good when logged in and not logged in

### DIFF
--- a/components/LandingPage/Footer.tsx
+++ b/components/LandingPage/Footer.tsx
@@ -2,8 +2,9 @@ import Link from 'next/link';
 import Image from 'next/image';
 import style from '/styles/landing.module.css';
 import logoImageSource from 'assets/logo.png';
+import LogoutButton from './LogoutButton'
 
-export default function Footer(): JSX.Element {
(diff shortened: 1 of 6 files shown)


Klakurka (commented):
The logo size is the same now but position is slightly off -- do you think it's better to just leave it like that or adjust the alignment on the landing page to match?

johnkuney:
ah yeah sure, I adjusted so the heights will match. Unfortunately theres not a great way to ensure the horizontal positions are always the same if you are a large screen, since the dashboard has no fixed width, the logo can be farther to the left. Can adjust any of those pieces if needed though of course depending what feels best for larger screens

Klakurka (approved):
Good enough for me.

PR #728, opened 2023-10-31
[#660] feat: big address sync improvement

chedieck:
Related to #660

Depends on
---
- [x] #724
- [x] #725 
- [x] #726 
- [x] #727 

Description
---
Improves the performance when inserting transactions.

This is achieved using for-loops instead of async concurrent promises to insert the txs on the database, while running every batch of the tx insertions inside a DB transaction.

The plots below show how this considerably brings down the CPU usage without diminishing the insertion speed too much. For all of them, the beginning should be seen as the initial sync of addresses + spinning up of the app, then a hiatus of nothing being synced and then, finally,  the addition of a large address (in all examples [qrcuqadqrzp2uztjl9wn5sthepkg22majylrprqkuk](https://ecashexplorer.com/address/ecash:qrcuqadqrzp2uztjl9wn5sthepkg22majylrprqkuk) was used). This last part of the plot is where the tx sync of a large address is happening.

## Master:
performance_plot

## This branch:
performance_plot

Test plan
---

1. Make sure you have python and the python packages `matplotlib` and `dotenv` installed.
2. Spin up the containers and then run `python scripts/performance.py`. On the plots above, I've used `make reset-dev && sleep 20; python scripts/performance.py `
3. Log in the dev user and wait until the activity in the plot is close to 0
4. Add a button with a large address.
5. Check the performance.
6. To stop running the performance script, close the plot and immediately press CTRL+C.
7.  A `plot_performance.jpg` file will be saved in the repository root dir.


Klakurka (changes requested):
- Txs are initially added quickly when a new address is added, then it totally stalls out for some reason.

- On /button/xxx when syncing, it doesn't automatically re-add the pagination controls as the number of results goes above the max for the page -- at least for a while, then it adds the pagination controls but they can't be interacted with. image The cursor shows up as a red circle w/ a red line through it.

I tried testing this with the IFP address (ecash:prfhcnyqnl5cgrnmlfmms675w93ld7mvvqd0y8lz07)

Klakurka (changes requested):
It grabs the first 100 txs quickly but then seems to stall out.

Klakurka (approved):
Next branch will fix retrieving more than 100 results.

PR #729, opened 2023-11-01
Button generator refactor 

johnkuney:
Related to #652

Description
---
Refactoring the button generator component to be more flexible and easy to maintain
The way it was first built was a headache to keep track of all of the inputs. Now all the items are in one place, and easy to update/add to/ move around
Will make adding the other options much quicker

Also adding the other 'standard' options here. Widget, random satoshis, and goal amount 

Test plan
---
Run the app and try out the button generator, it should function as before


Klakurka (changes requested):
- Can we hide or fade out the 'Random Satoshis' controls when 'Amount' = 0?
- Our own site CSS is overriding stuff in the widget... I guess this needs a new client task? I feel like some of the CSS work done on the dialog didn't also apply to the widget which it should have.
- In light mode, the 'Currency' and 'Animation' options are faded out.
- Can we move the 'Reset' and 'Documentation' buttons to be within the container above it so that when widget is toggled on, it doesn't use up as much vertical space?

johnkuney:
Okay added hiding random satoshis when amount = 0, and moved the reset options closer

I couldnt reproduce the faded dropdowns? can you screenshot? And looking into fixes those overridden styles on the client

johnkuney:
how about now?

Klakurka:
Looks good -- last thing being the css overriding issue on the widget.

johnkuney:
For sure, just put a PR up for that

Klakurka (approved):

Issue #730, opened 2023-11-01
Cache balance


chedieck:
When fetching wallets containing large addresses (many txs), it takes very long to calculate the balance, thus making both the wallets page and the button add modal very slow

Create a `balance` field on wallets to cache this balance & a websocket listener to update it after sync completion & tx arrival.

PR #733, opened 2023-11-06
[#660] feat: big address sync improvement > pagination

chedieck:
Related to #660 

Description
---
#### Paginated tx fetch on button detail table

We no longer fetch all txs for some address and then pass this as props to the table component, we actually pass the getter function to the table component and it only fetches from the DB  the txs we are actually going to render.

This means that addresses with a lot of txs don't become unusable, although there is still some considerable lag (~1s) when ordering & changing pages for these types of addresses, since prisma's findMany function lags when skipping & counting entries. 

Test plan
---
Test the `button/[id]` (button detail view) for large and small addresses: the orderings should work, the new TX arrival automatic refresh should work, etc.

This PR also adds new integration tests to test the new query parameters for the `address/transactions/[addressId]` endpoint.


Klakurka (changes requested):
- After loading the first 100 results, it starts grabbing single transactions after but really really slowly. 
image

- When txs are loading in, it should be made clear in the UI. We had this but somehow it's not working here. image
- Sometimes after refreshing, the total number of results shrinks (it went from 102 for me to 100, then to 101).

chedieck:
> * After loading the first 100 results, it starts grabbing single transactions after but really really slowly.

I couldn't reproduce that; with the address in question I only see it grabbing single *new* transactions, since all the old ones are seeded at build time and should be present already.
>     * When txs are loading in, it should be made clear in the UI. We had this but somehow it's not working here. 

This should show "No transactions.", I don't understand why it doesn't show in yours. Maybe there is something in the browser/app console log?
> Sometimes after refreshing, the total number of results shrinks (it went from 102 for me to 100, then to 101).

I'll try to reproduce that, it is indeed weird. 


>     * Wallets page loading time is really bad after adding a big address.
This is consequence of #730, I'll fix in a separate PR

Klakurka (changes requested):
image

Dashboard doesn't seem to be sync'd up with the button txs.

image
image

Klakurka (approved):
Should be fixed up in a follow-up branch to deal with caching issues.

PR #735, opened 2023-11-08
UI/button generator tooltips

johnkuney:
Related to #

Description
---
Adding some tooptips to the button generator fields based on copy from the docs page 

Test plan
---
Run the app and check the button generator. Try the tool tips on the form

### DIFF
--- a/components/ButtonGenerator/button-generator.module.css
+++ b/components/ButtonGenerator/button-generator.module.css
@@ -63,6 +63,7 @@
 .col_xs,
 .col_currency {
   padding: 3px;
+  position: relative;
 }
 
 .col_lg {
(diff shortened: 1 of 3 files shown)


Klakurka (changes requested):
Nice idea.

Let's either keep to using proper complete sentences like we're doing on the docs page or go super brief.

I think the only one that's rubbing me the wrong way is the TO one.

I don't have a specific suggestion but hopefully you get what I mean.

johnkuney:
Sure, guess I lean towards brief for tooltips, how about these?
Also note, I added one on random satoshis, but it was making the label too long, so I abbreviated the label...let me know if any issues/suggestions there

Klakurka (approved):

Issue #734, opened 2023-11-06
Grey out '<<' and '<' paging buttons when they're disabled


Klakurka:
<img width="153" alt="image" src="https://github.com/PayButton/paybutton-server/assets/4713204/103b224f-aa4f-4205-b339-08b4e5bbee15">

PR #739, opened 2023-11-13
[#734] UI: Pagination buttons disabled styles

johnkuney:
Related to #734

Description
---
Greying out pagination buttons when they are disabled

Test plan
---
Preview the site and check out a table with multiple pages


Klakurka (changes requested):
Can we go with `opacity: 0.1` ?

Klakurka (approved):

Issue #741, opened 2023-11-21
Paginate `payments` endpoint


chedieck:
the `api/payments` can return a large array, so this should be paginated just like the endpoint to get transactions (and the frontend should expect paginated responses accordingly)

Issue #316, opened 2022-11-17
Create loading screen 


johnkuney:
Need to add some sort of loading animation/icon/screen for each page

Klakurka:
Some options:

- https://tobiasahlin.com/spinkit/
- https://uiball.com/loaders/
- https://www.framer.com/motion/

Klakurka:
To tweak the scope of this a bit - we do want the page to "load" instantly, then have components individually load in as they can be.

PR #742, opened 2023-11-21
[#730] feat: cache rework

chedieck:
Related to #730

Description
---
1. Add cache for the balance of an address.
2. Better organize cache logic. We currently support 3 different caches (dashboard data, payment list and the address balance), this is now reflect into three files inside the `redis/` directory plus a `types.ts` file with all cache interfaces. Furthermore, there is now two interfaces `CacheSet` and `CacheGet` on `redis/index.ts` which will be used all across the codebase to actually get or set the cache values. The `CacheSet` interfaces export functions that are named after the unique context they will be used: for example: `CacheSet.paybuttonUpdate` is what will run when a paybutton is updated (and never in any other situation).
4. Uses `isSyncing` new address column to show on the wallet card that one of its addresses is syncing
5. Fix out-of-date cache issues when new txs arrive.

Test plan
---
Experiment with having a big address added (I use `ecash:qrcuqadqrzp2uztjl9wn5sthepkg22majylrprqkuk`) and checking if the dashboard values, the balance shown in the wallet and the payment list all are updated live, as the transactions arrive.

Also, after the sync is done, check that these values are not changing anymore and load very fast. This should include the logic for creating a Paybutton, since it gets the data for wallets and this in master was bottlenecking paybutton creation when a wallet would have a lot of transactions.

The whole codebase should work well even with large addresses added to the account, except for the `payments` page which should still await for #741 to have a smooth behavior.


Klakurka (changes requested):
This is much better than what we have in master.

Here's the issues I've been able to find:

- If you add an address with a boatload of txs, the Dashboard doesn't seem to get most of them: 
image
- Would be good to be able to differentiate between when there are 'No transactions.' and when it just needs a moment to grab them the first time. We have a separate task for this I'm pretty sure.
image
- 'Payments' page no longer has a scrollbar for some reason (although scrolling with the mouse or keyboard still works).

chedieck:
> If you add an address with a boatload of txs, the Dashboard doesn't seem to get most of them:

This seems to be true to the `IFP_ADDRESSES`, I created a commit that should fix it.

>   Would be good to be able to differentiate between when there are 'No transactions.' and when it just needs a moment to grab them the first time. We have a separate task for this I'm pretty sure.


Indeed. I found this issue #316 which is not too specific but we can make that clearer there.

>   'Payments' page no longer has a scrollbar for some reason (although scrolling with the mouse or keyboard still works).
This shall be fixed when I address the Payments page in detail in the next PR,

Klakurka (changes requested):
When loading up the IFP address, the dashboard shows the revenue changes (doubles) on Nov 18, but this actually happened on Nov 15... any ideas?

image


It's obviously not a time zone issues as it's 3 days apart.

Klakurka (approved):
You know what -- it actually looks totally fine. Due to the difficulty adjustment that took place until Nov 17, there were fewer than expected blocks then.

We can double check the numbers later but looks good at first glance.

balanceCache.ts
```diff
@@ -0,0 +1,39 @@
+import { Prisma } from '@prisma/client'
+import { AddressPaymentInfo, getAddressPaymentInfo } from 'services/addressService'
+import { redis } from './clientInstance'
+import { Payment } from './types'
+
+// ADDRESS:balance
```
Klakurka, line 6: What's the value of this comment?

constants/index.ts
```diff
@@ -167,9 +167,9 @@ export const PRIMARY_BCH_COLOR = '#4bc846'
 export const SECONDARY_BCH_COLOR = '#FFFFFF'
 export const TERTIARY_BCH_COLOR = '#231f20'
 
-export const IFP_ADDRESSES = [
-  'ecash:pqnqv9lt7e5vjyp0w88zf2af0l92l8rxdg2jj94l5j',
-  'ecash:prfhcnyqnl5cgrnmlfmms675w93ld7mvvqd0y8lz07'
+export const IFP_ADDRESSES: string[] = [
+  // 'ecash:pqnqv9lt7e5vjyp0w88zf2af0l92l8rxdg2jj94l5j',
+  // 'ecash:prfhcnyqnl5cgrnmlfmms675w93ld7mvvqd0y8lz07'
```
Klakurka, line 172: We can just remove them. If we need them again later then we can add them since they're known.

redis/balanceCache.ts
```diff
@@ -0,0 +1,56 @@
+import { Prisma } from '@prisma/client'
+import { AddressPaymentInfo, generateAddressPaymentInfo } from 'services/addressService'
+import { TransactionWithAddressAndPrices } from 'services/transactionService'
+import { redis } from './clientInstance'
+import { Payment } from './types'
+
+// ADDRESS:balance
```
Klakurka, line 7: ?

redis/dashboardCache.ts
```diff
@@ -1,231 +1,171 @@
...
-  hash: string
-  buttonDisplayDataList: ButtonDisplayData[]
-}
-
-export interface ButtonData {
-  displayData: ButtonDisplayData
-  total: {
-    revenue: Prisma.Decimal
-    payments: number
+// USERID:dashboard
```
Klakurka, line 9: ?

redis/paymentCache.ts
```diff
@@ -0,0 +1,199 @@
...
+import { Prisma } from '@prisma/client'
+import { getTransactionValue, TransactionWithAddressAndPrices, TransactionWithPrices } from 'services/transactionService'
+import { AddressWithTransactionsWithPrices, fetchAllUserAddresses, fetchAddressById, AddressWithPaybuttons } from 'services/addressService'
+import { fetchPaybuttonArrayByUserId } from 'services/paybuttonService'
+
+import { RESPONSE_MESSAGES, PAYMENT_WEEK_KEY_FORMAT, KeyValueT } from 'constants/index'
+import moment from 'moment'
+import { CacheSet } from 'redis/index'
+import { ButtonDisplayData, Payment } from './types'
+// ADDRESS:payments:YYYY:MM
```
Klakurka, line 11: ?

services/paybuttonService.ts
```diff
@@ -154,7 +154,7 @@ export async function createPaybutton (values: CreatePaybuttonInput): Promise<Pa
   // This has to be done before, or the connection
   // on relation tables to these addresses will fail
   const addressIdList: string[] = []
-  const updatedAddressIdList: string[] = []
+  const updatedAddressStringList: string[] = []
   const createdAddresses: addressService.AddressWithTransactions[] = []
 
   // Prevent IFP addresses from being created
```
Klakurka, line 160: Any reason to keep this (specifically referring to the IFP address, not blocking addresses in general)?

services/paybuttonService.ts
```diff
@@ -209,13 +209,11 @@ export async function createPaybutton (values: CreatePaybuttonInput): Promise<Pa
       include: includeAddresses
     })
     // Update cache for existent addreses
```
Klakurka, line 211: Better comment would be explaining _when_ this happens.

prisma/seed.ts
```diff
@@ -69,6 +68,7 @@ async function main (): Promise<void> {
   }
 
   // PRODUCTION
+  /*
```
Klakurka, line 71: Remove?

Klakurka (changes requested):

balanceCache.ts
chedieck: It's the format of the caching key. In this case it adds not too much because this is easily readable some lines below at:
```
return `${addressString}:balance`
```

but it's more useful in `paymentCache.ts`, where `// ADDRESS:payments:YYYY:MM
`  is way more readable than:

``` 
return `${addressString}:payments:${moment.unix(timestamp).format(PAYMENT_WEEK_KEY_FORMAT)}`
```

...so I decided to leave it in all cache-specific files, for consistency.

constants/index.ts
chedieck: Done.

redis/balanceCache.ts
chedieck: This is actually the real file, the other one I accidentally put in the root dir, no idea how. I am deleting that other file so the thread will disappear, pasting here the answer I had left there:

---
image

services/paybuttonService.ts
chedieck: I think this was implemented mixing the two concepts: IFP addresses are the ones we block. But this is a matter of variable naming (if I understood your question correctly), so would be only a matter of renaming the constant `IFP_ADRESSES` to `BLOCKED_ADDRESSES`.

services/paybuttonService.ts
chedieck: I think this comment was more useful before the creation of `CacheSet`; removed.

I don't understand what would be the _when_ you refer to, since this function will always update the cache

prisma/seed.ts
chedieck: Done.

redis/dashboardCache.ts
chedieck: https://github.com/PayButton/paybutton-server/pull/742#discussion_r1403654024

redis/paymentCache.ts
chedieck: https://github.com/PayButton/paybutton-server/pull/742#discussion_r1403654024

services/paybuttonService.ts
Klakurka: Ok let's do this then.

services/paybuttonService.ts
Klakurka: Nevermind -- but comments should be useful if they're going to be there so I'll leave it up to you what you want to do with it.

Klakurka (changes requested):
Probably just the one renaming and we're good.

chedieck:
> Probably just the one renaming and we're good.

Addressed.

Klakurka (approved):

Issue #744, opened 2023-11-28
Tx syncing doesn't recover after node comes back online


Klakurka:
It looks good in the Networks page but the actual txs don't come through.

Issue #732, opened 2023-11-03
Txs missing from cache


Klakurka:
Cain is running into this.

Klakurka:
Yes -- still missing some:

<img width="1546" alt="image" src="https://github.com/PayButton/paybutton-server/assets/4713204/f1949565-9076-4593-ac99-7375714f8dd7">

(Should go back to Nov 15)

PR #745, opened 2023-11-30
[#732, #744] fix: new tx pipeline

chedieck:
Related to #732, #744

Description
---

Solves some bugs related to chronik and tx sync:
- Avoid duplicate calls when txs that are related to more than one address arrive.
For example, if a tx from address A to B is broadcasted and both addresses A and B are in our DB, the new tx pipeline will trigger twice and therefore run for both addresses twice. This PR introduces a 3s delay where txs that are already being processed won't be processed again. This won't conflict with other incoming transactions or arriving confirmations.

- Upon deploy connect transactions with one price to the missing second price

- Don't take the current month into account to consider the monthly time windows. This was accidentally done, such that in months with less than 31 days, payments that happened in the 31st of any month  wouldn't be considered in the dashboard. 

Test plan
---

- Some addresses (those with payments in the 31st) are having a divergence in the dashboard "Lifetime payments" numbers, in the main top display and in the top of the rightside chart. That shouldn't happen anymore. Note that this can only be tested today, since tomorrow will be December 1st, a month with 31 days.

- When a TX from A to B happens and both A and B are in our DB, we shouldn't get prisma errors anymore.

### DIFF
--- a/constants/index.ts
+++ b/constants/index.ts
@@ -178,3 +178,5 @@ export const COOKIE_NAMES = {
 }
 
 export const DEFAULT_EMPTY_TABLE_MESSAGE = 'No rows to show.'
+
+export const CHRONIK_MESSAGE_CACHE_DELAY = 3000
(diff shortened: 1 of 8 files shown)


Klakurka (approved):
Looks good.

PR #746, opened 2023-12-01
Fix: supertokens mysql version & price connection 

chedieck:
Description
---
Fix docker production start file and error where broadcast & confirmation in different days would lead to issues.

### DIFF
--- a/docker-compose-prod.yml
+++ b/docker-compose-prod.yml
@@ -4,7 +4,6 @@ services:
     container_name: paybutton-dev
     restart: always
     depends_on:
-      - db
       - users-service
       - cache
     build: .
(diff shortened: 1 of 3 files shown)


docker-compose-prod.yml
```diff
@@ -19,9 +18,8 @@ services:
   users-service:
     container_name: paybutton-users-service
     depends_on:
-      - db
       - cache
-    image: registry.supertokens.io/supertokens/supertokens-mysql:15.04
+    image: registry.supertokens.io/supertokens/supertokens-mysql:6.0
```
Klakurka, line 22: Version 6 of what?

Klakurka (changes requested):
Checks failings.

docker-compose-prod.yml
Klakurka: Ok they use different version numbers than vanilla MySQL.

PR #754, opened 2023-12-11
[#732] fix: avoid tx holes

chedieck:
Related to #732

Description
---
Make so we sync the missed transactions one address by one, and log the errors in case there are any.

Test plan
---
This can be tested in local dev by:
1. Have the containers running with `make dev` and some of your addresses in the database.
2. Stop all containers but the database one with `make stop-prod`
3. Make sure the database is up and running with `yarn docker db` and check how many txs you have on that address of yours with `SELECT count(*) FROM Address where address='ecash:youraddresshere';`
4. Transfer some XEC to that address
5. Build the project with `make prod` and make sure that after it is up the newly arrived tx is synced, that same counting query should return one more entry.

### DIFF
--- a/jobs/workers.ts
+++ b/jobs/workers.ts
@@ -89,7 +89,7 @@ export const connectAllTransactionsToPricesWorker = async (queueName: string): P
       console.log(`job ${job.id as string}: connecting prices to transactions...`)
       const txs = [
         ...await transactionService.fetchAllTransactionsWithNoPrices(),
-        ...await transactionService.fetchAllTransactionsWithOnePrice()
+        ...await transactionService.fetchAllTransactionsWithIrregularPrices()
       ]
       void await transactionService.connectTransactionsListToPrices(txs)
     },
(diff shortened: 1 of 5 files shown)


Klakurka (changes requested):
The transactions appeared but one of them was not marked as confirmed:

image

How I did this was:

1. Start the server.
2. Make a payment to the address.
3. Shut down the non-db containers.
4. Make another payment to the address (I used the same amount for both to try to intentionally confuse it).
5. Wait for a block to come in.
6. Make another payment to the address.
7. Start the containers.
8. View the /button/ page.

chedieck:
> The transactions appeared but one of them was not marked as confirmed:

Nice catch, this should be fixed now.

Klakurka (changes requested):
Works but checks are failing.

chedieck:
Fixed

Klakurka (changes requested):
Unconfirmed txs don't show up as payments until confirmed if received while the server was down.

Also I have a missing payment.

image

image

Klakurka (approved):
Works for me now!

Klakurka:
If it does come up again, we can address it at that time.

Issue #751, opened 2023-12-07
'GET AN ADDRESS' link missing from Button Generator


Klakurka:
<img width="160" alt="image" src="https://github.com/PayButton/paybutton-server/assets/4713204/0dca96bd-573b-4a6f-a2e4-b35768ab739f">

PR #755, opened 2023-12-11
[#751] UI: Add get address link to button generator

johnkuney:
Related to #751

Description
---
Adding the get address link back in to the button generator

Test plan
---
preview the site and check the link on the 'to' field works and looks good


Klakurka (approved):
In the future, we may want to add some kind of help dialog w/ some more info / context to help guide them to set up their wallet for the kind of stuff they're planning on doing with their buttons but this is good for now as it's the quickest way to get people set up with their first address.

PR #778, opened 2024-01-10
feat: update README

chedieck:
Description
---
Update the README:
1. No need to install docker-compose since now it comes with docker;
2. Missing environment variable (used to generate keys for users)


Klakurka (changes requested):
I think we should better explain what each of the config options actually are / what they do.

Can we add a section for that?

chedieck:
> I think we should better explain what each of the config options actually are / what they do.
> 
> Can we add a section for that?

Done

Klakurka (approved):

Issue #774, opened 2024-01-04
Add basic support for OP_RETURN


Klakurka:
- Parse OP_RETURN of each transaction coming in and insert entries into a new "OP_RETURN" table with each key/value (separated by `=`) as well as the unprocessed OP_RETURN in a new column in the transaction table
- Add `<op_return>` variable to payment triggers
- Add a new button on each Transaction row, under 'TX' that opens a dialog to view the OP_RETURN data

We can get fancier with this later w/ filters, searching, etc but for now we can just focus on inserting and then allowing them some way to read them.

Klakurka:
https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/doc/standards/bip21-ecash-additions.md

PR #782, opened 2024-01-12
[#774] feat: basic OP_RETURN parsing

chedieck:
Related to #774

Description
---
Implements the parsing of  [NULL DATA](https://learnmeabitcoin.com/technical/nulldata) (OP_RETURN) scripts on transactions.

Test plan
---
1. Restart the containers. Necessary to reload chronik.

2. Send a TX with ElectrumABC OP_RETURN field as `paybutton$anypieceofdatayouwant` to any address you have on the running app,

3. Use `yarn docker db` and then `select createdAt, hash, amount, extraData from Transaction where not extraData='';` to see which txs have some data on it. Besides the data for the just sent tx, there should  be some data for txs I already sent when developing this feature.


Klakurka (approved):
The field is `opReturn`, not `extradata`, but works great!

PR #783, opened 2024-01-12
[#774] feat: op return variable on triggers

chedieck:
Related to #774

Depends on
---
- [x] #782 

Description
---
Adds <opReturn>  as a valid variable for setting up triggers.

Test plan
---
1. Create your trigger, use a website you control or a dummy one such as https://httpbin.org/post.
2. Make sure to include the variable in the trigger body, such as ` {"op": <opReturn> }` 
3. Send a tx to the address with OP_RETURN=`paybutton$anyDataYouWant` and make sure it appears in the receiving end of the POST request e.g. `{"op": 'anyDataYouWant'}`


Klakurka (approved):
Really cool!

PR #786, opened 2024-01-17
feat: scripts waits for DB connection

chedieck:
Description
---
Right now the `paybutton-dev` container fails when trying to connect to the `paybutton-db` container the first times simply because the DB container takes longer to start.

This change makes it wait until it has started.

Test plan
---
Check that the containers start normally

### DIFF
--- a/scripts/paybutton-server-start.sh
+++ b/scripts/paybutton-server-start.sh
@@ -1,6 +1,14 @@
 #!/bin/sh
 . .env
 . .env.local
+
+echo Waiting for db...                                              
+while true; do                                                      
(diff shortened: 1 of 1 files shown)


Klakurka:
I get this at startup now.... 
image

Klakurka (changes requested):
I actually can't get it to build -- just outputs `bad port` over and over again.

Klakurka (approved):

Klakurka:
Needed to fix my CRLF (windows formatted line breaks) on .env and .env.local.

PR #793, opened 2024-01-21
Fix/windows env crlf

Klakurka:
Description
---
On Windows, `.env` was getting its line breaks set to CRLF instead of LF which was causing problems. This forces it to LF.

Test plan
---
Build should work.


chedieck (dismissed):
Works for me

chedieck (approved):

Issue #801, opened 2024-01-27
Use explicit </div> tag on visible button generator code snippet


Klakurka:
This is causing an issue on some websites.

Klakurka:
<img width="562" alt="image" src="https://github.com/PayButton/paybutton-server/assets/4713204/7f1297c3-4513-4341-ac85-82c430ef158a">

PR #807, opened 2024-01-29
fix(#801): close div properly

lissavxo:
Related to #

Description
---
- changed the example to close the div to avoid errors 
- added indentation to the code

Test plan
---
Open the project, run the build `docker-compose up` and check the button generator if the html example is closing the div properly


Klakurka:
How about removing the extra space line here?

image

Klakurka (changes requested):
Alignment issues.

image

Issue #798, opened 2024-01-27
Send hmac of account secret, not the secret itself, in payment triggers


Klakurka:

chedieck:
we will have to change from sending a hmac to sending signature, in order for it to be verifiable with assymetric cryptography

PR #820, opened 2024-02-16
[#798] feat: change hmac to signature

chedieck:
Related to #798 

Depends on
---
- [x] #821

Description
---
Signs the POST request instead of using a hashmac.

The `<signature>` variable now is an object of the format:
```
{
  "payload": [variables present in the POST data concataned by + sign]
  "signature": [signature of the payload using the user's private key]
}
``` 
Therefore, users could form the payload themselves using their logic or just used the one we used out of the box.

Test plan
---
After receiving a POST request from a trigger, use the public key available in the **Account** view, the signature and the payload to verify integrity. An example I used on my running `staking` dev app:
```
import crypto from 'crypto'


interface Signature {
  payload: string
  signature: string
}

const verifySignature = (pubKey: string, signature: Signature) => {
  const pubKeyBuffer = Buffer.from(pubKey, 'hex')
  const publicKey = crypto.createPublicKey({ key: pubKeyBuffer, format: 'der', type: 'spki' });
  return crypto.verify(
    null,
    Buffer.from(signature.payload),
    publicKey,
    Buffer.from(signature.signature, 'hex')
  )
}

```

### DIFF
--- a/pages/account/index.tsx
+++ b/pages/account/index.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from 'react'
-import Image from 'next/image';
+import Image from 'next/image'
 import ThirdPartyEmailPasswordNode from 'supertokens-node/recipe/thirdpartyemailpassword'
 import supertokensNode from 'supertokens-node'
 import * as SuperTokensConfig from '../../config/backendConfig'
@@ -8,8 +8,8 @@ import { GetServerSideProps } from 'next'
 import Page from 'components/Page'
 import ChangePassword from 'components/Account/ChangePassword'
 import style from './account.module.css'
-import { getUserSecretKey } from 'services/userService'
-import CopyIcon from '/assets/copy-black.png';
+import { getUserPrivateKey } from 'services/userService'
+import CopyIcon from '../../assets/copy-black.png'
 
 export const getServerSideProps: GetServerSideProps = async (context) => {
   // this runs on the backend, so we must call init on supertokens-node SDK
@@ -33,7 +33,7 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
     props: {
       userId,
       user: await ThirdPartyEmailPasswordNode.getUserById(userId),
-      userSecret: await getUserSecretKey(userId)
+      userSecret: await getUserPrivateKey(userId)
     }
   }
 }
@@ -47,7 +47,7 @@ interface IProps {
 export default function Account ({ user, userId, userSecret }: IProps): React.ReactElement {
   const [changePassword, setChangePassword] = useState(false)
   const [secretKeyInfo, setSecretKeyInfo] = useState(false)
-  const [isCopied, setIsCopied] = useState(false);
+  const [isCopied, setIsCopied] = useState(false)
   const toggleChangePassword = (): void => {
     setChangePassword(!changePassword)
   }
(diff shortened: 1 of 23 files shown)


Klakurka (changes requested):
I think I'd prefer to test this using the "simple nextjs server" when it's rdy.

Why not simplify it by exclusing the 'payload' and then just make it clear in the documentation how to do the concatenation?

chedieck:
> I think I'd prefer to test this using the "simple nextjs server" when it's rdy.

Ok

> Why not simplify it by exclusing the 'payload' and then just make it clear in the documentation how to do the concatenation?

Because however clear we make it, it still wouldn't be clear as seeing the thing itself. And since our rules of concatenation (variables present in the request, ordered alphabetically, + sign as separator, and the particularity of the OP_RETURN data) are not obvious, I think a lot of people would just prefer to use the already formed payload.

There is no insecurity in that as far as I can tell; we form the payload using the same variables we also send. Unless they go into the Blockchain themselves to see this variables, seeing the variables WE passed or seeing the concatenated string WE formed (from those same variables) is effectively the same.

Klakurka (changes requested):
I'm also getting this error on the dashboard (it's blank but I found a link to it in dev tools):

image

pages/account/index.tsx
```diff
@@ -86,35 +85,34 @@ export default function Account ({ user, userId, userSecret }: IProps): React.Re
...
               This can be used to verify the authenticity of a message received from a paybutton trigger.
               <br/>
               <br/>
-              To verify, take the variables &lt;amount&gt;, &lt;currency&gt;, &lt;address&gt;, &lt;timestamp&gt;, and &lt;txId&gt; and concatenate them into a single string using the plus (+) symbol as a separator. The string should look like this: `&lt;amount&gt;+&lt;currency&gt;+&lt;address&gt;+&lt;timestamp&gt;+&lt;txId&gt;`.
-              <br/>
-              <br/>
-              Next, pass this concatenated string to the `sha256` hash function to generate a hash. Make sure to get the output in hexadecimal format. This will give you the data hash digest.
+              To verify, check variable &lt;signature&gt;, it should contain two keys:
               <br/>
+              - payload: The transaction data variables present in the POST request concatenated using the plus (+) symbol as a separator
```
Klakurka, line 104: Add a period to the end of this.

pages/account/index.tsx
```diff
@@ -86,35 +85,34 @@ export default function Account ({ user, userId, userSecret }: IProps): React.Re
...
+            <div className={style.public_key_info_ctn}>
+              !! WIP !!
               This can be used to verify the authenticity of a message received from a paybutton trigger.
               <br/>
               <br/>
-              To verify, take the variables &lt;amount&gt;, &lt;currency&gt;, &lt;address&gt;, &lt;timestamp&gt;, and &lt;txId&gt; and concatenate them into a single string using the plus (+) symbol as a separator. The string should look like this: `&lt;amount&gt;+&lt;currency&gt;+&lt;address&gt;+&lt;timestamp&gt;+&lt;txId&gt;`.
-              <br/>
-              <br/>
-              Next, pass this concatenated string to the `sha256` hash function to generate a hash. Make sure to get the output in hexadecimal format. This will give you the data hash digest.
+              To verify, check variable &lt;signature&gt;, it should contain two keys:
```
Klakurka, line 102: To verify, check the &lt;signature&gt; variable. It should contain two keys:

Klakurka (changes requested):

Klakurka (dismissed):

Issue #803, opened 2024-01-29
Add button to allow viewing / copying all payments for a particular button, or set of buttons


Klakurka:
Add a small button to the /payments page which will either display a simple page of every payment where the text can be copied, or to a .CSV with the same data.

Columns (in order from left to right):
 - Date (YYYY-MM-DD)
 - Amount (in XEC); 2 decimal places
 - Value (in their preferred currency); 2 decimal places
 - Rate (value / amount); 14 decimal places
 - Transaction ID
 - Button Name

Klakurka:
Maybe we can put the button on the end (right side) of the 'Transactions' title section on /button/xxx ?

![image](https://github.com/PayButton/paybutton-server/assets/4713204/9e517c98-d0d0-4bd2-95bd-01e6224d45c8)

chedieck:
I think MVP here would be an "Export" button that exports it to CSV, you agree?

lissavxo:
> Maybe we can put the button on the end (right side) of the 'Transactions' title section on /button/xxx ?
> 
[...]

PR #832, opened 2024-06-12
Feature/add csv util

lissavxo:
Related to #803 

Description
---
Added a new util that receives headers and data and writes it in a CSV file using a stream, in this case the stream will be the NextResponse. 
This will be used in the new endpoint to download the txs in a csv 

Test plan
---
Run yarn docker test

### DIFF
--- a/utils/files.ts
+++ b/utils/files.ts
@@ -3,14 +3,14 @@ import { NextApiResponse } from "next";
 import { Transform } from "stream";
 
 export function valuesToCsvLine(values: string[], delimiter: string): string {
-    return values.join(delimiter);
+    return values.join(delimiter) + '\n';
 }
 interface DataObject {
     [key: string]: any;
 }
 
-export function getDataFromValues(data: DataObject, headers: string[]): Partial<DataObject> {
-    const result: Partial<DataObject> = {};
+export function getDataFromValues(data: DataObject, headers: string[]): DataObject {
+    const result: DataObject = {};
     headers.forEach(header => {
         result[header] = data[header];
     });
@@ -21,7 +21,7 @@ export function getDataFromValues(data: DataObject, headers: string[]): Partial<
 export const getTransform = (headers: string[], delimiter: string) => new Transform({
     objectMode: true,
     transform(chunk: any, encoding: BufferEncoding, callback: () => void) {
-        const csvLine = valuesToCsvLine(headers.map(header => chunk[header]), delimiter) + '\n';
+        const csvLine = valuesToCsvLine(headers.map(header => chunk[header]), delimiter);
         this.push(csvLine);
         callback();
     },
@@ -33,7 +33,7 @@ export function streamToCSV(values: object[], headers: string[], delimiter: stri
     const transform = getTransform(headers, delimiter);
 
     try {
-        res.write(csvLineHeaders + '\n');
+        res.write(csvLineHeaders);
         values.slice(0, maxRecords).forEach((data: object) => {
             const formattedData = getDataFromValues(data, headers);
             transform.write(formattedData);
--- a/utils/files.ts
+++ b/utils/files.ts
@@ -5,12 +5,9 @@ import { Transform } from "stream";
 export function valuesToCsvLine(values: string[], delimiter: string): string {
     return values.join(delimiter) + '\n';
 }
-interface DataObject {
-    [key: string]: any;
-}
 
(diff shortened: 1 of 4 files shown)


utils/files.ts
```diff
@@ -0,0 +1,46 @@
...
+    transform(chunk: any, encoding: BufferEncoding, callback: () => void) {
+        const csvLine = headers.map(header => chunk[header]).join(',') + '\n';
+        this.push(csvLine);
+        callback();
+    },
+});
+
+export function streamtoCSV(values: object[], headers: string[], res: NextApiResponse | any) {
+    const maxRecords = 10; 
+    const formattedHeaders = resolveHeaders(headers);
```
chedieck, line 29: The variable is called `formattedHeaders`, the function `resolveHeaders` and what it does is join the headers with a comma.
So a few things here:
1. Consistent naming: the variable could be called `concatenatedHeaders` and the function `concatenateHeaders`, for example. Best even something like `concatenateColumns` since I guess we'd be using it to both headers and other rows.
2. Maybe use another delimiter than `,`?I suppose we'll have the Button Name in one of the columns, so best use a char that wouldn't appear there such as `\t`.

utils/files.ts
```diff
@@ -0,0 +1,46 @@
+import { RESPONSE_MESSAGES } from "../constants/index";
+import { NextApiResponse } from "next";
+import { Transform } from "stream";
+
+export function resolveHeaders(headers:string[]): string {
+    return headers.join(',');
+}
+
+export function getDataFromValues(data:any, headers: string[]): any {
```
chedieck, line 9: A better way to type this would be:
`export function getDataFromValues<T extends Record<string, any>>(data: T, headers: string[]): Partial<T> { ...`

then below you can define `result: Parital<T>` so we avoid using `any`.

utils/files.ts
```diff
@@ -0,0 +1,46 @@
...
+        result[header] = data[header];
+    });
+
+    return result;
+}
+
+export const getTransform = (headers:string[]) => new Transform({
+    objectMode: true,
+    transform(chunk: any, encoding: BufferEncoding, callback: () => void) {
+        const csvLine = headers.map(header => chunk[header]).join(',') + '\n';
```
chedieck, line 21: instead of repeating the join here you should use the function you defined above.

utils/files.ts
```diff
@@ -0,0 +1,46 @@
...
+        res.write(formattedHeaders + '\n');
+        values.slice(0, maxRecords).forEach((data: object) => {
+            const formattedData = getDataFromValues(data, headers);
+            transform.write(formattedData);
+        });
+
+        transform.end();
+        transform.pipe(res);
+    } catch (error: any) {
+        console.log(error.message)
```
chedieck, line 42: Is this on purpose? If so it should be `console.error`

chedieck (changes requested):

utils/files.ts
lissavxo: Thanks for the suggestion about naming, I improved that, and also added the delimiter option and some more tests.

utils/files.ts
```diff
@@ -0,0 +1,48 @@
...
+    },
+});
+
+export function streamToCSV(values: object[], headers: string[], delimiter: string, res: NextApiResponse) {
+    const maxRecords = 10;
+    const csvLineHeaders = valuesToCsvLine(headers, delimiter);
+    const transform = getTransform(headers, delimiter);
+
+    try {
+        res.write(csvLineHeaders + '\n');
```
chedieck, line 36: Both times you call the function `valuesToCsvLine` you end up appending a `'\n'` to it, so might as well make the function add that already.

utils/files.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export function valuesToCsvLine(values: string[], delimiter: string): string {
+    return values.join(delimiter);
+}
+interface DataObject {
+    [key: string]: any;
+}
+
+export function getDataFromValues(data: DataObject, headers: string[]): Partial<DataObject> {
+    const result: Partial<DataObject> = {};
```
chedieck, line 13: AFAIK here `Partial<DataObject>` is the same as `DataObject`, since there are no explicitly defined keys you are saying could not be present.

chedieck (changes requested):

chedieck (approved):

Klakurka (approved):

PR #833, opened 2024-06-17
feat: get addresses by paybuttonId

lissavxo:
Related to #803 

Description
---
Added a service method to fetch addresses by paybuttonId

Test plan
---
Run project with `docker-compose up` make sure the project runs correctly,  check `yarn docker test` and  `yarn docker testcoverage` to analyse coverage reports

### DIFF
--- a/services/addressService.ts
+++ b/services/addressService.ts
@@ -309,9 +309,16 @@ export async function fetchAddressWithTxsAndPrices (addressString: string): Prom
 }
 
 export async function fetchAddressesByPaybuttonId(paybuttonId: string): Promise<string[]> {
-  return await prisma.addressesOnButtons.findMany({
+  const addresses  = await prisma.addressesOnButtons.findMany({
     where: {
       paybuttonId: paybuttonId
     },
-  }).then(results => results.map(result => result.addressId))
+  })
+  const addressesIds = addresses.map(result => result.addressId)
+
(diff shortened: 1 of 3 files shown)


services/addressService.ts
```diff
@@ -307,3 +307,11 @@ export async function fetchAddressWithTxsAndPrices (addressString: string): Prom
   }
   return result
 }
+
+export async function fetchAddressesByPaybuttonId(paybuttonId: string): Promise<string[]> {
```
chedieck, line 311: Could you refactor it so:
1. It doesn't uses `then`, only the `await` in parenthesis (we stick to this pattern of only `async/await` across the code) 
2. It doesn't uses the `Address.id` but the `Address.address`. Truth is this was implemented badly by me: there is absolutely no need for the address to have an UUID, since it's address string (e.g. ecash:qw82f...) is already an unique identifier. I noticed that when the codebase was already infected with too much `address.id` but have been avoiding it since then. The `Address.address` always fills up the same purpose but it's way more useful since it's information which is not exclusive to our databases but actually useful externally on the networks.

chedieck (changes requested):

services/addressService.ts
lissavxo: thanks for the feedback, refactored

chedieck (approved):

PR #835, opened 2024-06-25
Feat/get tx by paybutton

lissavxo:
Related to #803

Description
---
Added a service method to fetch transactions by paybuttonId

Test plan
---
Run `yarn docker test` and `yarn docker testcoverage` to check test coverage reports

### DIFF
--- a/services/transactionService.ts
+++ b/services/transactionService.ts
@@ -478,8 +478,8 @@ export async function fetchAllTransactionsWithIrregularPrices (): Promise<Transa
 }
 
 export async function fetchTransactionsByPaybuttonId(paybuttonId: string): Promise<Transaction[]>{
-  const addressesList = await fetchAddressesByPaybuttonId(paybuttonId);
-  const transactions = await fetchAddressListTransactions(addressesList)
+  const addressIdList = await fetchAddressesByPaybuttonId(paybuttonId);
+  const transactions = await fetchAddressListTransactions(addressIdList)
   
   if(transactions.length === 0){
     throw new Error(RESPONSE_MESSAGES.NO_TRANSACTION_FOUND_404.message)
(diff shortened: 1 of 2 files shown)


services/transactionService.ts
```diff
@@ -476,3 +476,14 @@ export async function fetchAllTransactionsWithIrregularPrices (): Promise<Transa
   })
   return txs.filter(t => t.prices.length !== 2)
 }
+
+export async function fetchTransactionsByPaybuttonId(paybuttonId: string): Promise<Transaction[]>{
+  const addressesList = await fetchAddressesByPaybuttonId(paybuttonId);
```
chedieck, line 481: addressIdList

services/transactionService.ts
```diff
@@ -476,3 +476,14 @@ export async function fetchAllTransactionsWithIrregularPrices (): Promise<Transa
...
+  const transactions = await fetchAddressListTransactions(addressesList)
+  
+  if(transactions.length === 0){
+    throw new Error(RESPONSE_MESSAGES.NO_TRANSACTION_FOUND_404.message)
```
chedieck, line 485: Is this really desirable? Just asking if we want to throw an error if there are no transactions in a paybutton, which could happen.

ALso, if it's the case, we should throw a plural error: this 404 is for when we look for a specific transaction and don't find it.

chedieck (changes requested):

services/transactionService.ts
lissavxo: if you check the same service you will see a function called fetchManyTransactionsById that has the same behavior.
I believe 404 will be interpreted by UI that is consuming the api as a no data found issue.
I could fix the message for No transaction found for the paybuttonId but I don't think it is necessary

lissavxo: > Is this really desirable? Just asking if we want to throw an error if there are no transactions in a paybutton, which could happen.
> 
> ALso, if it's the case, we should throw a plural error: this 404 is for when we look for a specific transaction and don't find it.

if you check the same service you will see a function called fetchManyTransactionsById that has the same behavior.
I believe 404 will be interpreted by UI that is consuming the api as a no data found issue.
I could fix the message for No transaction found for the paybuttonId but I don't think it is necessary

lissavxo (commented):
fixed

chedieck (approved):

PR #837, opened 2024-06-28
feat: get tx by paybuttonId

lissavxo:
Related to #803 

Description
---
Added endpoint to fetch transactions by paybuttonId

Test plan
---
- Create a button with at least two adresses
- call route `http://localhost:3000/paybutton/transactions/{paybuttinId}
- Make sure the transactions returned are the same you see in the button page

### DIFF
--- a/pages/api/paybutton/transactions/[id].ts
+++ b/pages/api/paybutton/transactions/[id].ts
@@ -1,0 +1,23 @@
+import { NextApiRequest, NextApiResponse } from 'next'
+import { fetchTransactionsByPaybuttonId } from 'services/transactionService'
+
+export default async ( req: NextApiRequest, res: NextApiResponse): Promise<void> => {
+  if (req.method === 'GET') {
+    const paybuttonId = req.query.id as string;
+    try {
+      const transactions = await fetchTransactionsByPaybuttonId(paybuttonId);
+      
+      res
+        .status(200)
+        .json({ transactions });
+    } catch (err: any) {
+      res
+        .status(500)
+        .json({ statusCode: 500, message: err.message })
+    }
+  } else {
+    res
+      .status(405)
+      .json({ statusCode: 405, message: 'Method not allowed' })
+  }
+}
--- a/pages/api/paybutton/transactions/[id].ts
+++ b/pages/api/paybutton/transactions/[id].ts
@@ -25,6 +25,6 @@ export default async ( req: NextApiRequest, res: NextApiResponse): Promise<void>
   } else {
     res
       .status(405)
-      .json({ statusCode: 405, message: 'Method not allowed' })
+      .json(RESPONSE_MESSAGES.METHOD_NOT_ALLOWED)
   }
 }
--- a/pages/api/paybutton/transactions/[id].ts
+++ b/pages/api/paybutton/transactions/[id].ts
@@ -1,36 +1,35 @@
-import { RESPONSE_MESSAGES } from 'constants/index';
-import { NextApiRequest, NextApiResponse } from 'next'
+import { RESPONSE_MESSAGES } from 'constants/index'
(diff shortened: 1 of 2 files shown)


pages/api/paybutton/transactions/[id].ts
```diff
@@ -0,0 +1,23 @@
...
+        .status(200)
+        .json({ transactions });
+    } catch (err: any) {
+      res
+        .status(500)
+        .json({ statusCode: 500, message: err.message })
+    }
+  } else {
+    res
+      .status(405)
```
Comment (chedieck) [inline pages/api/paybutton/transactions/[id].ts:20]: Where is this gonna be thrown?

Also, missing case for NO_TRANSACTION_FOUND_404

chedieck (changes requested):

pages/api/paybutton/transactions/[id].ts
Comment (lissavxo) [reply pages/api/paybutton/transactions/[id].ts:27]: the " method not allowed " error is something present in almost every /api endpoint 
added case for 404

Comment (chedieck) [reply pages/api/paybutton/transactions/[id].ts:27]: Looks like a good idea indeed, but since we should be using this for endpoints from now on, you can create a variable in RESPONSE_MESSAGES like it's done everywhere else.

pages/api/paybutton/transactions/[id].ts
```diff
@@ -0,0 +1,30 @@
+import { RESPONSE_MESSAGES } from 'constants/index';
+import { NextApiRequest, NextApiResponse } from 'next'
+import { fetchTransactionsByPaybuttonId } from 'services/transactionService'
+
+export default async ( req: NextApiRequest, res: NextApiResponse): Promise<void> => {
+  if (req.method === 'GET') {
```
Comment (chedieck) [inline pages/api/paybutton/transactions/[id].ts:6]: We should check if the user requesting the txs is indeed the owner of the paybutton. Or if it's an admin.

Check `pages/api/paybutton/[id].ts` for an example of that being done.

chedieck (changes requested):

pages/api/paybutton/transactions/[id].ts
```diff
@@ -0,0 +1,43 @@
...
+import { NextApiRequest, NextApiResponse } from 'next'
+import { fetchTransactionsByPaybuttonId } from 'services/transactionService'
+import * as paybuttonService from 'services/paybuttonService'
+import { setSession } from 'utils/setSession'
+
+export default async ( req: any, res: any): Promise<void> => {
+  if (req.method === 'GET') {
+    await setSession(req, res)
+    const userId = req.session.userId
+    console.log("validou")
```
Comment (chedieck) [inline pages/api/paybutton/transactions/[id].ts:11]: remove this

pages/api/paybutton/transactions/[id].ts
```diff
@@ -0,0 +1,43 @@
...
+      if (err.message === RESPONSE_MESSAGES.NO_TRANSACTION_FOUND_404.message){
+        res
+        .status(RESPONSE_MESSAGES.NO_TRANSACTION_FOUND_404.statusCode)
+        .json(RESPONSE_MESSAGES.NO_TRANSACTION_FOUND_404)
+      }
+      if (err.message === RESPONSE_MESSAGES.RESOURCE_DOES_NOT_BELONG_TO_USER_400.message){
+        res
+        .status(RESPONSE_MESSAGES.RESOURCE_DOES_NOT_BELONG_TO_USER_400.statusCode)
+        .json(RESPONSE_MESSAGES.RESOURCE_DOES_NOT_BELONG_TO_USER_400)
+      }else{
```
Comment (chedieck) [inline pages/api/paybutton/transactions/[id].ts:32]: spacing is weird here. Linting didn't went through?

chedieck (changes requested):

chedieck (changes requested):
still many linting issues, let's solve the linting problem before moving on

chedieck (approved):

Klakurka (approved):

PR #839, opened 2024-07-12
Feat/return raw opreturn

lissavxo:
Description
---
Added raw opReturn data to SimplifiedTransaction response.
It will help us compare the transaction opReturn with the Paybutton opReturn set in the client

Test plan
---
Run the server with docker-compose up make a transaction and check the data if the rawMessage is being sent as expected. Run yarn docker test to make sure tests are covering the change.


chedieck (approved):
LGTM

PR #841, opened 2024-07-18
Feat/download transactions file

lissavxo:
Related to #803 

Description
---
Added a new endpoint to download transactions of a paybuttonId  in a CSV file.
here are some defaults : 

> csv delimiter =  \t 
> maxRecords = 2000
> currency = USD

Test plan
---
Run project with `docker-compose up`
call this URL with your `paybuttonId` ->` http://localhost:3000/api/paybutton/download/transactions/<paybuttonId>`
check if the file is downloaded and if data is compatible.


constants/index.ts
```diff
@@ -195,3 +196,24 @@ export const TRIGGER_POST_VARIABLES = [
   '<timestamp>',
   '<txId>'
 ]
+
+export const DEFAULT_PAYBUTTON_TRANSACTIONS_FILE_HEADERS = [
```
chedieck, line 200: These two could just be a mapping,  no?

constants/index.ts
```diff
@@ -195,3 +196,24 @@ export const TRIGGER_POST_VARIABLES = [
...
+  'date',
+  'amount',
+  'value',
+  'rate',
+  'transactionId',
+  'paybuttonName'
+];
+
+export const DEFAULT_PAYBUTTON_TRANSACTIONS_FILE_DELIMITER = '\t'
+export const DEFAULT_MAX_RECORDS_FILES = 2000
```
chedieck, line 219: Shouldn't it be DEFAULT_MAX_RECORDS_PER_FILE?

BTW, if this is not gonna be changeable, should remove the DEFAULT prefix.

chedieck (changes requested):
Left some comments but found linting issues, so I'll just leave those for now and finish the review when the precommit hook is properly set up

types/index.ts
chedieck, line 1: This is a legacy file which is not being used anywhere (you can even take the opportunity and delete it)

IMO you can just leave these type definitions in the utils/files.ts file, since they concern the functions there.

pages/api/paybutton/download/transactions/[paybuttonId].ts
```diff
@@ -0,0 +1,136 @@
...
+import { PaybuttonWithAddresses, fetchPaybuttonById } from 'services/paybuttonService'
+import { streamToCSV } from 'utils/files'
+import { setSession } from 'utils/setSession'
+import { NextApiResponse } from 'next'
+import { TransactionFileData, FormattedTransactionFileData } from 'types'
+type SupportedQuotesType = typeof SUPPORTED_QUOTES[number]
+
+const cors = Cors({ methods: ['GET', 'HEAD'] })
+
+const getPrice = (transaction: TransactionWithAddressAndPrices, currency?: SupportedQuotesType): number => {
```
Comment (chedieck) [inline pages/api/paybutton/download/transactions/[paybuttonId].ts:23]: - Better name: get transaction cost / value; since the price is already there in the transaction and what we want is to multiply it for the crypto amount.
- This could be in services/transactionService. Getting how much a tx in was valued in total looks generic enough to be there IMO

pages/api/paybutton/download/transactions/[paybuttonId].ts
```diff
@@ -0,0 +1,136 @@
...
+    if (p.price.quoteId === CAD_QUOTE_ID) {
+      result.cad = p.price.value.times(amount).toNumber()
+    }
+  }
+  return (!isCurrencyEmptyOrUndefined ? result[stringCurrency] : result.usd)
+}
+
+const getPaybuttonTransactionsFileData = (transaction: TransactionWithAddressAndPrices, paybutton: PaybuttonWithAddresses): TransactionFileData => {
+  const { amount, createdAt, id } = transaction
+  const value = getPrice(transaction)
```
Comment (chedieck) [inline pages/api/paybutton/download/transactions/[paybuttonId].ts:46]: `value = getPrice` reinforces the point I made in the comment above about the function `getPrice` renaming

pages/api/paybutton/download/transactions/[paybuttonId].ts
Comment (chedieck) [reply pages/api/paybutton/download/transactions/[paybuttonId].ts:23]: These four first lines in the function could just be 
`const stringCurrency = currency === '' || currency === undefined ? DEFAULT_QUOTE_SLUG : currency`

and then below, instead of 
`return (!isCurrencyEmptyOrUndefined ? result[stringCurrency] : result.usd)`

just
`return result[stringCurrency]`


...unless I am missing something?

types/index.ts
chedieck, line 1: This file should be deleted

chedieck (changes requested):

Klakurka:
Tests failing.

Klakurka (changes requested):
- Everything is going into the 1st column - there should be a separate column for each.

Klakurka (changes requested):
One other issue is we need to be using the actual network txid, not the one we're storing in the db.

Klakurka (dismissed):
Looks good!

I think we should probably also do the same thing but for payments as an alternative for people... thoughts?

For our own purposes though, this will work just fine.

lissavxo:
> Looks good!
> 
> I think we should probably also do the same thing but for payments as an alternative for people... thoughts?
> 
> For our own purposes though, this will work just fine.
nice, I think it is a good idea and It will be pretty similar to this logic so It should be not so hard to implement

constants/index.ts
```diff
@@ -205,3 +206,21 @@ export const TRIGGER_POST_VARIABLES = [
...
+  value: 'Value',
+  rate: 'Rate',
+  transactionId: 'Transaction Id',
+  paybuttonName: 'Paybutton Name'
+}
+
+export const DEFAULT_PAYBUTTON_CSV_FILE_DELIMITER = ','
+export const MAX_RECORDS_PER_FILE = 2000
+
+export const DECIMALS: { [key: string]: number } = {
```
chedieck, line 222: `Record<string, number>` for the type is more clear IMO

pages/api/paybutton/download/transactions/[paybuttonId].ts
```diff
@@ -0,0 +1,152 @@
...
+    headers,
+    DEFAULT_PAYBUTTON_CSV_FILE_DELIMITER,
+    res,
+    formmattedHeaders
+  )
+}
+
+export default async (req: any, res: any): Promise<void> => {
+  try {
+    await new Promise((resolve, reject) => {
```
Comment (chedieck) [inline pages/api/paybutton/download/transactions/[paybuttonId].ts]: Why is this needed?

- We only use CORS on endpoints that are supposed to be accessed by the PayButton client (or any external domain, for that matter).
- We have a function called `runMiddleware` in utils to do this set up.

services/transactionService.ts
```diff
@@ -3,14 +3,16 @@ import { Address, Prisma, Transaction } from '@prisma/client'
...
-import { RESPONSE_MESSAGES, USD_QUOTE_ID, CAD_QUOTE_ID, N_OF_QUOTES, KeyValueT, UPSERT_TRANSACTION_PRICES_ON_DB_TIMEOUT } from 'constants/index'
+import { RESPONSE_MESSAGES, USD_QUOTE_ID, CAD_QUOTE_ID, N_OF_QUOTES, KeyValueT, UPSERT_TRANSACTION_PRICES_ON_DB_TIMEOUT, SUPPORTED_QUOTES } from 'constants/index'
 import { productionAddresses } from 'prisma/seeds/addresses'
 import { appendTxsToFile } from 'prisma/seeds/transactions'
 import _ from 'lodash'
 import { CacheSet } from 'redis/index'
 import { SimplifiedTransaction } from 'ws-service/types'
 import { OpReturnData } from 'utils/validators'
 
+export type SupportedQuotesType = typeof SUPPORTED_QUOTES[number]
```
chedieck, line 14: I don't get this. What is happening here with this `[number]`? My LSP says this `SupportedQuotesType`  is the same as `string`.

services/transactionService.ts
```diff
@@ -478,13 +480,39 @@ export async function fetchAllTransactionsWithIrregularPrices (): Promise<Transa
...
+
+export const getTransactionValueInCurrency = (transaction: TransactionWithAddressAndPrices, currency: SupportedQuotesType): number => {
+  const {
+    prices,
+    amount,
+    id,
+    timestamp
+  } = transaction
+
+  const result: { [key in SupportedQuotesType]: number } = {}
```
chedieck, line 502: `Record<SupportedQuotesType, number>` for better readability.

chedieck (changes requested):

services/transactionService.ts
lissavxo: This [number] accesses the types of the elements within the SUPPORTED_QUOTES array. In TypeScript, using number as an index for an array type means "any element of the array."

Klakurka (approved):

pages/api/paybutton/download/transactions/[paybuttonId].ts
```diff
@@ -0,0 +1,142 @@
...
+  transactionId: string
+  address: string
+}
+
+function isCurrencyValid (currency: SupportedQuotesType): boolean {
+  return SUPPORTED_QUOTES.includes(currency)
+}
+
+const getPaybuttonTransactionsFileData = (transaction: TransactionWithAddressAndPrices, currency: SupportedQuotesType): TransactionFileData => {
+  const { amount, createdAt, hash, address } = transaction
```
Comment (chedieck) [inline pages/api/paybutton/download/transactions/[paybuttonId].ts:43]: `createdAt` won't necessarily reflect the true date the tx was sent/created.

When testing this PR, every tx in the CSV got today's date as their date since the txs have just been inserted to the DB when I reset the containers.

So to fix this, use the `timestamp` field, convert from unix timestamp to moment with `moment(timestamp * 1000)` and it should be good. 

For that, we have the field `timestamp` (which is a number for the

chedieck (changes requested):

pages/api/paybutton/download/transactions/[paybuttonId].ts
Comment (Klakurka) [reply pages/api/paybutton/download/transactions/[paybuttonId].ts:43]: Nice catch; missed this one - would have figured this out pretty quick after using it.

Comment (lissavxo) [reply pages/api/paybutton/download/transactions/[paybuttonId].ts]: > `createdAt` won't necessarily reflect the true date the tx was sent/created.
> 
> When testing this PR, every tx in the CSV got today's date as their date since the txs have just been inserted to the DB when I reset the containers.
> 
> So to fix this, use the `timestamp` field, convert from unix timestamp to moment with `moment(timestamp * 1000)` and it should be good.

thanks, addressed

Klakurka (changes requested):
Sort by date (moment), ASC.

Klakurka (approved):

chedieck (approved):
I would suggest bettering the header names, but this can easily be done in a future separate PR

LGTM

Issue #847, opened 2024-08-22
Have 'value' column use 2 decimal places on CSV export


Klakurka:
Right now it rounds with no decimal.

PR #849, opened 2024-08-23
fix: value fixed decimal

lissavxo:
Related to #847 

Description
---
Changed value decimal places to 2

Test plan
---
Run server with docker compose up, go to button page, download CSV and check value column

### DIFF
--- a/pages/api/paybutton/download/transactions/[paybuttonId].ts
+++ b/pages/api/paybutton/download/transactions/[paybuttonId].ts
@@ -69,7 +69,7 @@ const formatPaybuttonTransactionsFileData = (data: TransactionFileData): Formatt
     ...data,
     amount: amount.toFixed(DECIMALS[currency]),
     date: date.format(PRICE_API_DATE_FORMAT),
-    value: value.toFixed(DECIMALS[currency]),
+    value: value.toFixed(2),
     rate: rate.toFixed(14)
   }
 }


Klakurka (changes requested):
On this branch, the CSV export isn't working on many of the buttons.

It seems like it's not working if there's too many transactions to go through.

Able to reproduce?

I'm simply testing a few of the buttons in the local dev@ account.

lissavxo:
> On this branch, the CSV export isn't working on many of the buttons.
> 
> It seems like it's not working if there's too many transactions to go through.
> 
> Able to reproduce?
> 
> I'm simply testing a few of the buttons in the local dev@ account.

I was able to reproduce an error but It is related to missing price.
When there is 1 tx without price we stop generating the csv and throw error. 
Testing Paybutton:
  **account**: `dev@`
  **name**: `DEV` 
  **id**: `f03981ab-8ded-4aab-93e8-b1d38f9b219b `
  
I received this json : 
 `{ "message": "Missing price for transaction., txId 57f47d4b-6560-11ef-af58-0242ac120002, at 1705126827" }`
 
 If we check `getTransactionByPaybuttonId` endpoint `http://localhost:3000/paybutton/transactions/f03981ab-8ded-4aab-93e8-b1d38f9b219b` with the ID of the button mentioned above we can see that tx `57f47d4b-6560-11ef-af58-0242ac120002` is not returning prices

chedieck:
> I was able to reproduce an error but It is related to missing price.

So looks like this was related to the bug I introduced in master that would prevent the software from syncing all the prices.

It is fixed now, and the PR seems to work well.

I will approve it because it does what is supposed to do,  but first:
- Shouldn't we actually use 8 decimal cases for bitcoin cash? As of now 0.003 BCH =~1USD, so small txs would basically be all shown as 0 USD.

Klakurka:
This decimal change is only supposed to be on the 'value', not the 'amount'.

chedieck (approved):
> This decimal change is only supposed to be on the 'value', not the 'amount', so all good right?

Ahhh of course, approving in then.

PS: I still think we should at least put the units in the headings, like:
`amount (USD)` and `value (XEC)` (or `value (BCH)`) 

a lot more clarity in exchange for a few characters —the implicit difference that exists here between the words "amount" and "value" is something that we essentially bred inside our own codebase.

And even when used/interpreted by "codebase people", as shown by earlier my misunderstanding, it still can lead to confusion.

Klakurka:
> > This decimal change is only supposed to be on the 'value', not the 'amount', so all good right?
> 
> Ahhh of course, approving in then.
> 
> PS: I still think we should at least put the units in the headings, like: `amount (USD)` and `value (XEC)` (or `value (BCH)`)
> 
> a lot more clarity in exchange for a few characters —the implicit difference that exists here between the words "amount" and "value" is something that we essentially bred inside our own codebase.
> 
> And even when used/interpreted by "codebase people", as shown by earlier my misunderstanding, it still can lead to confusion.

Sure we can make a new task for that

Klakurka (approved):

PR #854, opened 2024-08-28
Fix/expected res for prices

chedieck:
Related to #

Description
---
When testing the response from coin.dance to check it's validity, we weren't considering that prices may be values in the response object.

Test plan
---
Price sync should be failing before and working now.


Klakurka (approved):
I'll approve but we do want to be careful about not using 'CAD' and 'USD' in the code since in the future we will probably support a dynamic set of currencies.

Issue #848, opened 2024-08-22
v2 CSV exporting


Klakurka:
- Add 'Export as CSV' button to 'Payments' page.
- Somehow separate out the transactions/payments by currency. Either have them appear on different "tabs" in the downloaded file (not sure if this is possible) or have them be on the same spreadsheet but separated by an empty row.

POTENTIALLY OBSOLETE:
- However you want to do it, allow exporting just from single addresses within the page. For example we could add the 'Export as CSV' button to the top right of each address tx list (probably the easiest). Another option could be changing the 'Export as CSV' button into a dropdown which includes 'All' at the top then a list of the addresses below.

Klakurka:
We decided we're going to go with this BUT as a single "button dropdown": ![ae5ace44-11c2-42f1-97bd-ca517048933d](https://github.com/user-attachments/assets/4ef77f2e-f2a7-433b-9b30-28fba9dd3e47)

PR #857, opened 2024-09-10
Feat/update csv download

lissavxo:
Related to #848 

Description
---
Added a selector in CSV export to split the transaction by currency

Test plan
---
- Test export csv all currencies and single currency

### DIFF
--- a/constants/index.ts
+++ b/constants/index.ts
@@ -166,6 +166,8 @@ export const NETWORK_TICKERS: KeyValueT<string> = {
   bitcoincash: 'BCH'
 }
 
+export type NetworkTickersType = 'XEC' | 'BCH'
+
 export const NETWORK_TICKERS_FROM_ID: KeyValueT<string> = {
   1: 'XEC',
   2: 'BCH'
--- a/constants/index.ts
+++ b/constants/index.ts
@@ -220,7 +220,6 @@ export const PAYBUTTON_TRANSACTIONS_FILE_HEADERS = {
   rate: 'Rate',
   transactionId: 'Transaction Id',
   address: 'Address'
-
 }
 
 export const DEFAULT_PAYBUTTON_CSV_FILE_DELIMITER = ','
--- a/pages/api/paybutton/download/transactions/[paybuttonId].ts
+++ b/pages/api/paybutton/download/transactions/[paybuttonId].ts
@@ -9,9 +9,10 @@ import {
   DEFAULT_QUOTE_SLUG,
   SupportedQuotesType,
   NetworkTickersType,
-  NETWORK_TICKERS
+  NETWORK_TICKERS,
+  NETWORK_IDS
 } from 'constants/index'
-import { TransactionWithAddressAndPrices, fetchTransactionsByPaybuttonId, getTransactionValueInCurrency } from 'services/transactionService'
+import { TransactionWithAddressAndPrices, fetchTransactionsByPaybuttonIdGroupedByNetwork, getTransactionValueInCurrency } from 'services/transactionService'
 import { PaybuttonWithAddresses, fetchPaybuttonById } from 'services/paybuttonService'
 import { streamToCSV } from 'utils/files'
 import { setSession } from 'utils/setSession'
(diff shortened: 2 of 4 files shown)


Klakurka (changes requested):
- USD/CAD is going to be an account setting. What I meant here was to have BCH and XEC split out.
- The dropdown label appears in the dropdown list but isn't clickable: 
image - you can just add a `hidden` attribute.
- Cursor should be the pointer when hovering over the dropdown buttons.

Klakurka (changes requested):
- Only show a currency in the dropdown if the button actually has an address for that currency.
- Only show the dropdown at all if the button has addresses for 2 or more currencies.

constants/index.ts
```diff
@@ -166,6 +166,8 @@ export const NETWORK_TICKERS: KeyValueT<string> = {
   bitcoincash: 'BCH'
 }
 
+export type NetworkTickersType = 'XEC' | 'BCH'
```
Klakurka, line 169: I'm pretty wary of stuff like this but we can leave it if Estevao thinks it's fine.

I made this phase 3 task though: https://github.com/PayButton/paybutton-server/issues/859

chedieck (changes requested):
Looks good, only some small changes in the code.

As for the pointer cursor in the dropdown, it would require to refactor the dropdown as something other than an actual `<select>` dropdown (such as a list of `<li>` elements) because using CSS to set the cursor as pointer in an actual dropdown doesn't work: the dropdown is rendered by the operating system, not the browser.

We could just ignore it but TBH I think would be way nicer UX if the cursor was really a pointer, since the experience of selecting something that is actually a button feels weird.

Klakurka (dismissed):

pages/button/[id].tsx
```diff
@@ -94,36 +96,52 @@ export default function Button (props: PaybuttonProps): React.ReactElement {
...
   useEffect(() => {
     void getDataAndSetUpSocket()
   }, [])
 
-  const downloadCSV = async (paybutton: { id: string }): Promise<void> => {
+  const downloadCSV = async (paybutton: { id: string, name: string }, currency: string): Promise<void> => {
     try {
-      const response = await fetch(`/api/paybutton/download/transactions/${paybutton.id}`)
+      let url = `/api/paybutton/download/transactions/${paybutton.id}`
+      const isCurrencyEmpty = (value: string): boolean => (value === '')
```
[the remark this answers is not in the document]
Comment (lissavxo) [reply pages/button/[id].tsx]: because I am using it in two places and I changed a bit to add undefined check too

Klakurka (dismissed):
Looks good to me.

We just need a way to get CAD working even if it's by manually setting it in the db for the account that generates the CSV.

pages/api/paybutton/download/transactions/[paybuttonId].ts
```diff
@@ -81,24 +88,32 @@ const formatNumberHeaders = (headers: string[], currency: string): string[] => {
...
-  }).map(tx => {
+  currency: SupportedQuotesType,
+  networkTicker?: NetworkTickersType): Promise<void> => {
+  let networkIdArray = Object.values(NETWORK_IDS)
+  if (networkTicker !== undefined) {
+    const slug = Object.keys(NETWORK_TICKERS).find(key => NETWORK_TICKERS[key] === networkTicker)
+    const networkId = await getNetworkIdFromSlug(slug ?? NETWORK_TICKERS.ecash)
+    networkIdArray = [networkId]
+  }
+  const transactionsGrouped = await fetchTransactionsByPaybuttonIdGroupedByNetwork(paybutton.id, networkIdArray)
```
Comment (chedieck) [inline pages/api/paybutton/download/transactions/[paybuttonId].ts:99]: Why do you need them grouped if you just ungroup it afterwards? This variable is not used anywhere further.

services/transactionService.ts
```diff
@@ -489,6 +500,24 @@ export async function fetchTransactionsByPaybuttonId (paybuttonId: string): Prom
   return transactions
 }
 
+export async function fetchTransactionsByPaybuttonIdGroupedByNetwork (
```
chedieck, line 503: This is created to be used in a place where the grouping is not relevant. I don't get why it exists, but in any case would be better if it took the txs list, not the button and networkids as a parameter.

chedieck (changes requested):

pages/api/paybutton/download/transactions/[paybuttonId].ts
Comment (lissavxo) [reply pages/api/paybutton/download/transactions/[paybuttonId].ts:99]: this is to separate/group by network

services/transactionService.ts
lissavxo: grouping is relevant, we order by date and group by network.. this way we separate the txs list.

Klakurka (approved):

Issue #860, opened 2024-09-21
Navigating around can cause logged-in site nav to appear before actually being logged in


Klakurka:
![image](https://github.com/user-attachments/assets/e985246d-5d96-4e17-b09c-179f495d613f)

To reproduce:

1. Navigate to https://paybutton.org/ (not logged in).
2. Click 'Sign In'.
3. Click 'Sign Up Now'.
4. Navigate back a page in the browser (so it goes back to 'Sign In').
5. You'll see the nav. Clicking any of the nav buttons will show you the 'Something went wrong.' error page.

johnkuney:
This is happening on the live site for you?
I tried at paybutton.org but couldnt reproduce, but it does happen if I'm running locally

johnkuney:
but not everytime lol

Klakurka:
Yes I can reproduce it easily on the live site.

Klakurka:
![image](https://github.com/user-attachments/assets/a615de00-579c-41e2-ada2-6d4136eaa723)

This is the live site (paybutton.org).

johnkuney:
okay, yeah its odd I cant get it to do it at all now, but will check it out

PR #865, opened 2024-09-24
[UI] Prevent sidebar when not logged in 

johnkuney:
Related to #860

Description
---
Preventing the sidebar from showing on sign in pages when navigating around 

Test plan
---
preview the site and check you cant cause the issue. It stopped happening for me all together so not sure if this solves it or not


Klakurka (approved):
This does seem to work. I will keep an eye on it in prod.

PR #866, opened 2024-09-25
Feat/preferred quote csv

lissavxo:
Related to #848 

Description
---
Added currency filter and implemented preferredCurrency filter in the export csv logic

Test plan
---
Run the project with `docker compose up`  go to button page and export CSV, check if currency is USD
Change the `preferredCurrencyId` in `UserProfile` table and check if the Value column in CSV is the one you set


Klakurka (approved):
It requires a refresh which is not a huge deal but maybe we can have it apply instantly to the page in a future update.

chedieck (approved):
LGTM

Issue #253, opened 2022-10-01
(Optionally) Send email upon receiving payment


Klakurka:
We don't want to just send emails for every transaction as it could be very easy for it to get out of hand / be abused.

We'll probably want to have a separate table for tracking button payment payloads but in general this kind of feature touches on the BIP 70 idea a bit so we'll want to have some discussions about how it'll all work before getting into it.

Klakurka:
Could also start with "email on every tx" to keep it simple for v1 since the userbase is going to be small for a while.

Klakurka:
I think a nicely fleshed-out v1 would be:

- Send emails on payments made to buttons that have the "Send email to: X@Y.Z" option set (under Payment Triggers).
- Log the emails we're sending out
- Limit it to 100 per day per account
- On the 100th email in 24 hours, include an extra line in bold + red font saying that they've reached their daily limit of email notifications
- Show how many emails they've used of their 100 email allotment next to where they set their email on any buttons where they have it active

PR #879, opened 2024-10-22
[#253]: send email on trigger

chedieck:
Related to #253

Description
---
Implements a v1 for #253.

### General changes:
- Users now can have two triggers at most in one paybutton: one for post requests, one for emails.
- This limit also applies to addresses. A user cannot create a second paybutton with a same address and add a trigger to it as if it was a purely new paybutton.
- Both triggers can be set, updated and deleted in the button detail view: one above the other. Their forms are independent.

### DB changes:
1. Added columns `PaybuttonTrigger.emails` for setting up which emails the email trigger should send to (right now we allow only for one email)
2. Removed `PaybuttonTrigger.sendEmail`
2. Added `PaybuttonTrigger.isEmailTrigger` to differentiate between the two types of triggers (defaults to false, to preserve our DB state)
4. Added `UserProfile.emailCredits` for tracking how many credits a user has for sending emails. Default is 100.

 

As usual, this PR implements only a rudimentary front-end, enough for testing purposes.

Test plan
---
Reset the containers, check if the conditions above are being met. This PR doesn't implement  the execution of the email sending, only the infrastructure to allow it. Poster triggers should work normally as they did before.


components/Paybutton/PaybuttonTrigger.tsx
```diff
@@ -133,23 +187,45 @@ export default ({ paybuttonId }: IProps): JSX.Element => {
...
+                      <button type='button' onClick={() => setClearModal('poster')} className={style.delete_btn}>Clear</button>
+                  }
+                </div>
+              </div>
+            </div>
+            </form>
+          <form onSubmit={(e) => { void handleSubmitEmailTrigger(getSubmitTriggerHandler('email'))(e) }} method='post'>
+            <h5>Receive Email</h5>
+            <div>
+              <label htmlFor="emails">Receive email</label>(<span>{emailCredits} credits</span>)
```
Klakurka, line 207: Change 'Receive Email' to 'Send email' to match post URL section.

Klakurka, line 207: Maybe we can add another "what is this?" for the email credits?

image

Klakurka (changes requested):

components/Paybutton/PaybuttonTrigger.tsx
chedieck: done

chedieck: done, but with a mocked explanation

Klakurka: Can you match the capitalization (Send request vs Send Email)?

Klakurka: Also change 'Receive email' below it to just 'Email:'.

Klakurka: Also while you're in there, remove the ':' from 'URL:'.

Klakurka: I've had a change of heart on this.

Let's scrap the word "credits" in the UI and remove the "What is this?" helper and replace it with something that says:

> You may send up to 100 emails per day.

or

> You have sent X of a maximum of 100 emails per day.

We can eventually add a "need more?" button link thing instead of the "What is this?" one to allow them to upgrade their account to "Pro" or w/e we end up calling that.

chedieck: Addressed, also modified the placeholder to be a little more clear.

chedieck: done, I changed "You have sent X of a maximum of 100 emails per day." to "You have sent X of a daily maximum of 100 emails.", let me know what you think.

Also created a dev3@paybutton.org user that has one button with a trigger and has not sent any email, to allow to test that too.

Klakurka (changes requested):
- Can't read the info bit in dark mode 
image
- Where does the values come from for this? 
image

chedieck:
> * Can't read the info bit in dark mode

Fixed

> Where does the values come from for this?

It's a constant. IDK why it didn't appear for you, it's not even connected to the DB, but probably restarting the containers or even just refreshing the page with Shift+f5 should fix it

Klakurka (approved):

PR #885, opened 2024-11-12
[#848] feat: use currency setting

chedieck:
Related to #848

Description
---
Implements the usage of the user prefered currency.

Test plan
---
1. Spin up the containers on **master**.

2. Switch the currency of the dev user to CAD. Checking out the dashboard then, should yield no difference.

3. Checkout to this branch. Still this should yield no difference

4. Now reset the cache. `yarn docker cr`.

5. Finally, the values in the dashboard and elsewhere should all be in CAD instead of USD (thus they will be significantly higher).

### DIFF
--- a/components/Chart.tsx
+++ b/components/Chart.tsx
@@ -1,7 +1,6 @@
 import type { NextPage } from 'next'
 import React from 'react'
 import { formatQuoteValue } from 'utils/index'
-import { USD_QUOTE_ID } from 'constants/index'
 import {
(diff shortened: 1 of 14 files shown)


Klakurka (changes requested):
- When I reset the cache, the lifetime revenue in the dashboard header stopped updated when I would toggle between USD and CAD: 
image
- Tests failing.

Klakurka (approved):

Issue #804, opened 2024-01-29
Prevent wallet add/edit dialog from growing too tall and running off the top/bottom of the screen


Klakurka:
This will happen if you have too many addresses added to the account.

Issue #794, opened 2024-01-22
Add new 'advanced' options to button generator


Klakurka:
- op_return (message / data)
- disable payment ID

PR #890, opened 2024-11-19
[UI] Add advanced button generator options

johnkuney:
Related to #794 & #804

Description
---
Adding op return and disable payment id to the button generator
Also sneaking in a CSS tweak to prevent the wallet dialog from running off the page

Test plan
---
Check the button generator and check the create or edit wallet dialog in an account with lots of buttons

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,4 +1,2 @@
 #!/bin/sh
-echo "PATH is: $PATH"
-which npx
 npx lint-staged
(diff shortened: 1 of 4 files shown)


Klakurka (changes requested):
- Nit but when you toggle the "disable payment id" option on/off, we're inconsistent with the other toggles where we normally just remove it from the div so they're not copying extra stuff they don't need.

johnkuney:
Ah yeah good catch, added

Klakurka (approved):

PR #892, opened 2024-11-30
Feat/paginate payments endpoint

lissavxo:
Related to #741 

Description
---
Added server side pagination to payments page, before the payments where fetched all at once in a single api call and the pagination was made in the client, now the client will fetch payments using page and pageSize parameters. 
Changed Pagination to fetch tx data from database instead of cache, because it was too complex to paginate and order getting data from cache.

Test plan
---
Run server and check payments page pagination, try to change pageSize and pass between the pages

### DIFF
--- a/redis/index.ts
+++ b/redis/index.ts
@@ -4,7 +4,7 @@ import { TransactionsWithPaybuttonsAndPrices } from 'services/transactionService
 import { fetchUsersForAddress } from 'services/userService'
 import { cacheBalanceForAddress, clearBalanceCache, getBalanceForAddress, updateBalanceCacheFromTx } from './balanceCache'
 import { clearDashboardCache, getUserDashboardData } from './dashboardCache'
-import { appendPaybuttonToAddressesCache, cacheGroupedPayments, cacheManyTxs, generateAndCacheGroupedPaymentsAndInfoForAddress, getCachedPaymentsCountForUser, getPaymentList, initPaymentCache, removePaybuttonToAddressesCache } from './paymentCache'
+import { appendPaybuttonToAddressesCache, cacheGroupedPayments, cacheManyTxs, generateAndCacheGroupedPaymentsAndInfoForAddress, getCachedPaymentsCountForUser, getCachedPaymentsForUserWithPagination, getPaymentList, initPaymentCache, removePaybuttonToAddressesCache } from './paymentCache'
 import { DashboardData, Payment } from './types'
 import { Address } from '@prisma/client'
 
@@ -130,5 +130,9 @@ export class CacheGet {
     return await this.executeCall(userId, 'paymentsCount', async () => {
       return await getCachedPaymentsCountForUser(userId)
     })
+  }
+  
+  static async paymentListPaginated (userId: string, page: number, pageSize: number): Promise<Payment[]> {
+    return await getCachedPaymentsForUserWithPagination(userId, page, pageSize)
   }
 }
--- a/redis/index.ts
+++ b/redis/index.ts
@@ -4,7 +4,7 @@ import { TransactionsWithPaybuttonsAndPrices } from 'services/transactionService
 import { fetchUsersForAddress } from 'services/userService'
 import { cacheBalanceForAddress, clearBalanceCache, getBalanceForAddress, updateBalanceCacheFromTx } from './balanceCache'
 import { clearDashboardCache, getUserDashboardData } from './dashboardCache'
-import { appendPaybuttonToAddressesCache, cacheGroupedPayments, cacheManyTxs, generateAndCacheGroupedPaymentsAndInfoForAddress, getCachedPaymentsCountForUser, getCachedPaymentsForUserWithPagination, getPaymentList, initPaymentCache, removePaybuttonToAddressesCache } from './paymentCache'
+import { appendPaybuttonToAddressesCache, cacheGroupedPayments, cacheManyTxs, generateAndCacheGroupedPaymentsAndInfoForAddress, getCachedPaymentsCountForUser, getPaymentList, initPaymentCache, removePaybuttonToAddressesCache } from './paymentCache'
 import { DashboardData, Payment } from './types'
 import { Address } from '@prisma/client'
 
@@ -130,9 +130,5 @@ export class CacheGet {
     return await this.executeCall(userId, 'paymentsCount', async () => {
       return await getCachedPaymentsCountForUser(userId)
     })
-  }
-
-  static async paymentListPaginated (userId: string, page: number, pageSize: number, orderDesc: boolean): Promise<Payment[]> {
-    return await getCachedPaymentsForUserWithPagination(userId, page, pageSize, orderDesc)
   }
 }
--- a/redis/index.ts
+++ b/redis/index.ts
@@ -131,8 +131,8 @@ export class CacheGet {
       return await getCachedPaymentsCountForUser(userId)
     })
   }
-  
-  static async paymentListPaginated (userId: string, page: number, pageSize: number): Promise<Payment[]> {
-    return await getCachedPaymentsForUserWithPagination(userId, page, pageSize)
+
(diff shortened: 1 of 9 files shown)


Klakurka:
Tests failing.

Klakurka (changes requested):
Viewing `/payments` initially works but if I do an F5 then it throws an error:

image

Klakurka (changes requested):
Can't get it to build:

image

chedieck:
> Can't get it to build:

This is happening on master, I put #897 up to fix it

Klakurka:
Needs a rebase.

Klakurka (approved):

Klakurka (dismissed):

Klakurka (dismissed):

redis/paymentCache.ts
```diff
@@ -32,6 +32,66 @@ export const getPaymentList = async (userId: string): Promise<Payment[]> => {
...
+      }
+    }
+    if (currentIndex >= endIndex) {
+      break
+    }
+  }
+  return allPayments
+}
+
+function sortByDate (keys: string[], order: 'ascending' | 'descending'): string[] {
```
chedieck, line 81: more clear name: `sortWeekKeysByDate`, since this function is specific for it.

chedieck (changes requested):
One thing about this PR is that it breaks the sorting using other columns. E.g, if we want to order by amount, there is not really a way to do that without looking at **all** the payments and then sorting them by amount.

Sorting by network would still be possible in this way, since the network of the address is in the cache key name. 

The idea of the issue was to avoid fetching all this data at once, which is possible when the user has it ordered by date or network, but not in the other cases (amount & buttons). I don't really see a way out of it, so maybe we can use pagination only in the case that the ordering is set to happen by date and fetch the whole lot of the data if the user wants it sorted in some other way.

lissavxo:
> One thing about this PR is that it breaks the sorting using other columns. E.g, if we want to order by amount, there is not really a way to do that without looking at **all** the payments and then sorting them by amount.
> 
> Sorting by network would still be possible in this way, since the network of the address is in the cache key name.
> 
> The idea of the issue was to avoid fetching all this data at once, which is possible when the user has it ordered by date or network, but not in the other cases (amount & buttons). I don't really see a way out of it, so maybe we can use pagination only in the case that the ordering is set to happen by date and fetch the whole lot of the data if the user wants it sorted in some other way.

don’t you think having two types of pagination might end up looking a bit messy?

Because for backend pagination, it’s one component, and for frontend, it’s another. I’d have to create something super complex. Is it really worth it?

chedieck:
> don’t you think having two types of pagination might end up looking a bit messy?
> 
> Because for backend pagination, it’s one component, and for frontend, it’s another. I’d have to create something super complex. Is it really worth it?

Yeah, would be messy indeed.

The main reason for this issue is that with many payments we get messages like "API response shouldn't be > 4MB", but the overall behavior is better the way already is, I think.

chedieck:
I have an idea: when e.g. requesting it sorted by amount, fetch the list of payments normally (from the DB, in the server), sort it, and then serving it to the client paginated. This fixes the issue of having large API responses and keeps the better answer time if sorted by date. If sorted by amount the answer time will be basically the same as it was already, but is not bad already for what I've tested, even for large addresses.

chedieck (changes requested):
Payments for a button with only ecash address are showing as being bitcoin cash payments

chedieck:
Bug I mentioned looks solved, but now tests are failing.

services/transactionService.ts
```diff
@@ -595,3 +612,202 @@ export async function getOldestPositiveTxForUser (userId: string): Promise<Trans
...
+    }
+    for (const p of JSON.parse(tx.prices)) {
+      if (p.quoteId === USD_QUOTE_ID) {
+        ret.usd = ret.usd.plus(p.priceValue * tx.amount)
+      }
+      if (p.quoteId === CAD_QUOTE_ID) {
+        ret.cad = ret.cad.plus(p.priceValue * tx.amount)
+      }
+    }
+    const buttonDisplayDataList: Array<{ name: string, id: string, providerUserId: string}> = []
```
chedieck, line 710: There exist already a type ButtonDisplayData, you should use it here

chedieck (changes requested):
image


Amount ordering is wrong.

I am thinking is because the ordering is on the cripto amount,  not on the converted value taking into account the price at the time of the tx.

chedieck (changes requested):
Lookgs good, but when ordering by button name, the network shows BCH icon where it should show the XEC icon.

services/transactionService.ts
```diff
@@ -595,3 +612,205 @@ export async function getOldestPositiveTxForUser (userId: string): Promise<Trans
...
+    const tx = transactions[index]
+    if (Number(tx.amount) > 0) {
+      const payment = await generatePaymentFromTx(tx)
+      transformedData.push(payment)
+    }
+  }
+  return transformedData
+}
+
+export async function fetchAllPaymentsByUserId (
```
chedieck, line 796: I feel this function shouldn't exist. The `if` could just be inside the `fetchAllPaymentsByUserIdWithPagination`, which is already a better name for what this function, given what it does, and then no else would be required.

services/transactionService.ts
```diff
@@ -123,6 +124,23 @@ const transactionsWithPaybuttonsAndPrices = Prisma.validator<Prisma.TransactionD
 
 export type TransactionsWithPaybuttonsAndPrices = Prisma.TransactionGetPayload<typeof transactionsWithPaybuttonsAndPrices>
 
+export const transactionWithAddressAndPricesAndButtons = Prisma.validator<Prisma.TransactionDefaultArgs>()({
```
chedieck, line 127: this won't be used anywhere externally, it's only needed to create the type below. No need to export it.

services/transactionService.ts
```diff
@@ -123,6 +124,23 @@ const transactionsWithPaybuttonsAndPrices = Prisma.validator<Prisma.TransactionD
...
+            paybutton: true // Include full Paybutton data
+          }
+        }
+      }
+    },
+    prices: true // Include related price data
+  }
+})
+
+export type TransactionWithAddressAndPricesAndButtons = Prisma.TransactionGetPayload<typeof transactionWithAddressAndPricesAndButtons>
```
chedieck, line 142: Why is this being created? I don't see it being used anywhere.

redis/paymentCache.ts
chedieck: No need to do that anymore, since this is obsolete, it can be deleted. However, if you think this could be useful for a future task, just put a comment indicating that and rename it for more clarity.

redis/index.ts
```diff
@@ -131,4 +131,8 @@ export class CacheGet {
       return await getCachedPaymentsCountForUser(userId)
     })
   }
+
+  static async paymentListPaginated (userId: string, page: number, pageSize: number, orderDesc: boolean): Promise<Payment[]> {
```
chedieck, line 135: This also is being created not to be used anywhere.

redis/paymentCache.ts
```diff
@@ -32,6 +32,65 @@ export const getPaymentList = async (userId: string): Promise<Payment[]> => {
   return await getCachedPaymentsForUser(userId)
 }
 
+export const getCachedPaymentsForUserWithPagination = async (
```
chedieck, line 35: This is also reminiscent of the earlier stages of this PR; being created not to be used anywhere.

services/transactionService.ts
chedieck: Actually, check the comment below, probably this shouldn't exist together with the type it is used for.

chedieck (changes requested):
Added requests for the code, many things reminiscent of earlier stages of this PR that need not be created anymore.

lissavxo:
> Looks good, but when ordering by button name, the network shows BCH icon where it should show the XEC icon.

there was a typo in the query, fixed

chedieck (approved):
Looks good!

Klakurka (approved):

Issue #896, opened 2024-12-08
Dashboard showing no data


Klakurka:

Klakurka:
Probably fixed but we'll keep an eye out.

PR #900, opened 2024-12-18
#896 fix/dashboard issues

chedieck:
Related to #896

Description
---
Dashboard on master was back to showing data, but it was all wrong. Payments would appear one month after they real date, and the view for **ALL** payments would include months with no payments in the beginning. This PR fixes these issues.

Test plan
---
On **master**:
1. Take any address with some payments, and check how on the dashboard they would be displaced: shifted one to the right, i.e, a payment from september would appear in october, for example.
2. Check how the **ALL** view in the dev user would include the months Dec2020 and Jan2021, with no payments. That is useless info that shouldn't be there, **ALL** should start from the first payment.

Then checkout this branch, run `yarn docker cr` to reset the cache and try it again — all the above should be fixed.

Remarks
---
I left a WIP comment to do the accounting for timezone — after we have that in the DB, labeling the payments in a tz-aware manner should be as simple as including the timezone info there.

### DIFF
--- a/redis/dashboardCache.ts
+++ b/redis/dashboardCache.ts
@@ -5,7 +5,7 @@ import { Prisma } from '@prisma/client'
 import moment, { DurationInputArg2 } from 'moment'
 import { XEC_NETWORK_ID, BCH_NETWORK_ID } from 'constants/index'
 import { QuoteValues } from 'services/priceService'
-import { getOldestTxForUser } from 'services/transactionService'
+import { getOldestPositiveTxForUser } from 'services/transactionService'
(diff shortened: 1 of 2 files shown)


Klakurka (changes requested):
I tried deleting all of the buttons in my local dev env, then adding all of the ones for BVC (prod). Dashboard is blank (only nav showing) and this is the error:

image

Klakurka (approved):
Master is also broken for me so I'm just going to accept this and we can address it if it's a problem in prod.

Issue #799, opened 2024-01-27
Some txs not showing as confirmed


Klakurka:
<img width="491" alt="image" src="https://github.com/PayButton/paybutton-server/assets/4713204/b8054c50-f210-4e55-958b-e5ef56bdb67d">

Klakurka:
Potentially fixed but we'll keep our eyes out for more instances of it.

Issue #768, opened 2023-12-22
Some txs not coming through


Klakurka:
Transaction is missing, payment is missing, and payment trigger didn't fire.

Klakurka:
Potentially fixed but we'll keep our eyes out for more instances of it.

PR #901, opened 2024-12-18
fix: unconfirmed middle txs

chedieck:
Related to #799, #768

Description
---
Two bugs are fixed here:
1. The `syncing` column of addresses were never updated to `false`, meaning they were always syncing and that the syncing of missing transactions was not working
2. The syncing of missed transactions would skip unconfirmed transactions that happened to be caught in the middle of confirmed transactions. That only could happen because of the bug above, but I fixed so that even if that happens, it will set these txs to confirmed

#768 and #799 should both be fixed here, except for the part that mentions triggers (I am assuming this is about triggers being send after downtime, which will be implemented in another PR)

Test plan
---
This is already running in prod, so there should be no more old unconfirmed txs nor missing txs.

To test in local, one can manually set a old tx in the db to `confirmed=0`, or delete an old tx and then reset **ONLY** the `paybutton-dev` container and see if it was fixed. This reset would probably need to be done twice, since the first time would only fix bug 1. (set the syncing flag to false) and the second time would fix bug 2. (sync the old txs)

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "pretest": "dotenv -e .env.test -- prisma migrate reset --force",
     "prisma": "dotenv -e .env -c -- prisma",
     "docker": "./scripts/docker-exec-shortcuts.sh",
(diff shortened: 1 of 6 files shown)


chedieck:
Changed to draft for debugging, tests are failing in github but not in my local env.

Klakurka (approved):

Issue #893, opened 2024-12-01
Transaction appearing a day later in dashboard (time zone issue)


Klakurka:
See that on Nov 30, 2024 at 17:29:18 UTC we received a donation to CD (BCH): https://blockchair.com/bitcoin-cash/address/bitcoincash:qzqh7rwaq9zm4zcv40lh9c9u50gy07gcesdmja8426

But in the PayButton dashboard, it's appearing under December 1: 
![image](https://github.com/user-attachments/assets/09121bf9-2721-4338-a29c-d8c508862ef1)

My local time zone is PST (GMT-8) so that doesn't account for the difference.

Klakurka:
Sort of related to #619.

PR #902, opened 2024-12-19
Feat/user timezone set up

lissavxo:
Related to #893 

Description
---
Added base set up for timezone 

Test plan
---
Run server with docker compose up make sure it works


chedieck (changes requested):
Looks good

chedieck (approved):

Issue #905, opened 2024-12-26
Record has changed since last read in table Price


chedieck:
```
2|next  | 2024-12-26T16:04:43: Problem inserting current price for 2 at timestamp 1733356800 using data {
2|next  | 2024-12-26T16:04:43:   Price_in_CAD: '827.10607987490000',
2|next  | 2024-12-26T16:04:43:   Price_in_USD: '587.54494568373000',
2|next  | 2024-12-26T16:04:43:   Price_in_BTC: '0.00569735466942',
2|next  | 2024-12-26T16:04:43:   '%_Price_Change': '2.57'
2|next  | 2024-12-26T16:04:43: } :
2|next  | 2024-12-26T16:04:43: PrismaClientUnknownRequestError: 
2|next  | 2024-12-26T16:04:43: Invalid `prisma.price.upsert()` invocation:
2|next  | 2024-12-26T16:04:43: 
2|next  | 2024-12-26T16:04:43: 
2|next  | 2024-12-26T16:04:43: Error occurred during query execution:
2|next  | 2024-12-26T16:04:43: ConnectorError(ConnectorError { user_facing_error: None, kind: QueryError(Server(MysqlError { code: 1020, message: "Record has changed since last read in table 'Price'", state: "HY000" })), transient: false })
```

Issue #903, opened 2024-12-20
Introduce 'submission' loading spinners


Klakurka:
For example when you go to create a new button with tons of addresses, it takes some time -- disable the submit button so it can't be clicked twice before the page updates and add some sort of visual spinner or other indicator that it's doing something.

Klakurka:
CSV generation button could use this too.

PR #907, opened 2024-12-27
fix: price upsert error

chedieck:
Related to #905

Description
---
When adding a big address (like ecash:prfhcnyqnl5cgrnmlfmms675w93ld7mvvqd0y8lz07), if the syncing process encountered some missing prices, this would lead to the error mentioned on #905.

This PR fixes this error, by uncoupling the process of filling up the missing prices and connecting it to the transaction.

Test plan
---
On master, after restarting the containers, try adding the aforementioned address to a button. After adding 200 txs around 10 times, it should result on the error happening multiple times.

Then restart the containers again on this branch, it should work fine.

### DIFF
--- a/services/priceService.ts
+++ b/services/priceService.ts
@@ -6,7 +6,7 @@ import { HUMAN_READABLE_DATE_FORMAT, PRICE_API_TIMEOUT, PRICE_API_MAX_RETRIES, P
 import { validatePriceAPIUrlAndToken, validateNetworkTicker } from 'utils/validators'
 import moment from 'moment'
 
-function flattenTimestamp (timestamp: number): number {
+export function flattenTimestamp (timestamp: number): number {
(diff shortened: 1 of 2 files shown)


Klakurka:
Potential options for speeding up the syncing process? We're currently doing 200 tx/s. Do we run into issues above this?

chedieck:
> Potential options for speeding up the syncing process? We're currently doing 200 tx/s. Do we run into issues above this?

It's the Chronik's limit of how many txs to fetch per call. Then it takes a while between each subsequent calls because we have to connect all these txs to addresses and prices we have in the DB, but I can have a look in another PR into making each one of these iterations take less time.

Klakurka (approved):
The speed isn't really that bad and IMO it's totally fine if we can improve the UI (eg. #903). Being able to submit a new address and have it sync in the bg would be even better.

PR #908, opened 2024-12-30
Feat/user timezone

lissavxo:
Related to #893 

Description
---
Implement timezone option in user Account settings. 
This PR will enable timezone option, the selector  is initialized with the timezone of the user by location, once user changes the selector value it will set the timezone and it will make the dashboard data, the txs listing of each button and payments listing to use the timezone selected. 

Test plan
---
Start the server with `docker compose up`
- Go to account page, check if selector is initialized with timezone of your location
- Change selector to another timezone
- Check payments list see if the date of each payment is in the preferred timezone  
- Check the transaction list in button page see if the date of each tx is in the preferred timezone 
- Check dashboard data, it should display the data according to the new timezone set


Klakurka (changes requested):
If I change my account's time zone, the dashboard doesn't update to reflect the fact that txs made on a certain date in one time zone are actually a day ahead/behind in another.

lissavxo:
> If I change my account's time zone, the dashboard doesn't update to reflect the fact that txs made on a certain date in one time zone are actually a day ahead/behind in another.

I was not clearing cache after setting the timezone, now It should work

Klakurka (changes requested):
I got this when trying to update my time zone:

image

Klakurka (changes requested):
I changed the time zone a few times and eventually got this on `/payments`:

image

Klakurka (approved):

pages/payments/index.tsx
```diff
@@ -40,17 +41,21 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
   return {
     props: {
       user,
-      userId
+      userId,
+      userProfile: user.userProfile
```
chedieck, line 45: This seems really redundant. You're sending the samething twice, first `user` then `user.userProfile` as userProfile

redis/paymentCache.ts
```diff
@@ -141,7 +141,7 @@ export const getCachedPaymentsForUser = async (userId: string): Promise<Payment[
 }
 
 export const getCachedPaymentsCountForUser = async (userId: string): Promise<number> => {
-  const dashboardData = await getUserDashboardData(userId)
+  const dashboardData = await getUserDashboardData(userId, moment.tz.guess())
```
chedieck, line 144: this doesn't seem right. It will be run in the server. I tested adding 
```
  console.log('WIP tz', moment.tz.guess(), typeof window, 'is window type')
```
right above this line and got
```
2|next  | 2025-01-07T00:45:35: WIP tz Africa/Abidjan undefined is window type
```

... as a result, which shows:
- guess is wrong
- is being run in the server, since this was in the **server** logs (not the browser console) and even shows that `typeof window === undefined`.

chedieck (changes requested):
Requested some changes in the code, also:

The other settings (Preferred currency & organization) give us feedback when they're updated. This could do the same as preferred currency, showing **"Updated timezone successfully"** after the update, or the error in case there is some.

components/Timezone Selector/index.tsx
```diff
@@ -24,9 +27,18 @@ const TimezoneSelector: React.FC<TimezoneSelectorProps> = ({ value }) => {
         })
         if (res.status === 200) {
           setSelectedTimezone(selectedOption.value)
+          setError('')
+          setSuccess('Timezone updated successfully.')
         }
       } catch (err: any) {
```
chedieck, line 33: this should actually be at the else of the `if ` above. If the api runs into an error, it will return a status code 500 response message, not throw here.

chedieck (approved):

Issue #911, opened 2025-01-08
Some transactions appearing a day late


Klakurka:
![c6fffcad-7b3d-4cb7-9b1b-a37d7d458f2c](https://github.com/user-attachments/assets/b254cc32-1412-4154-b949-cafb07b8a3a6)
![19a65046-1aa3-48ca-83d3-7215b97136f0](https://github.com/user-attachments/assets/77dfdcb5-8388-4f1b-9e4f-8925fcc1e092)

Klakurka:
https://explorer.e.cash/tx/98880f6b2a3314751acfc0fa0aebb26cee4e497a0b83cff35a977b488da494aa

This tx is appearing as having come in on Jan 12 instead of Jan 11 (UTC-8).

PR #912, opened 2025-01-09
fix: add timezone to csv

lissavxo:
Related to #911 

Description
---
Added timezone logic to CSV creation

Test plan
---
Run the server with `docker compose up`
download CSV it should display the date according to the timezone set.


Klakurka (changes requested):
I can't get Export to CSV to work... no errors either when clicking the button.

Klakurka (approved):
Seems to match all of the time stamps I'm expecting - I just couldn't get the CSV to generate due to a missing price error on a new button... unrelated to this branch though.

Issue #916, opened 2025-01-11
Add 'tx inputs' to payment trigger options


Klakurka:
Some apps will find it useful to know which address(es) was used for the input of a given tx (payment). Cashtab will (as of this writing) only shows 1 but electrum would be >=1.

PR #917, opened 2025-01-12
feat: broadcast input addresses

lissavxo:
Related to [#916]([https://github.com/PayButton/paybutton-server/issues/916])

Description
---
Send input addresses in broadcast incoming tx data, this will be used in paybutton client, this will make input addresses available on client callbacks 

Test plan
---
- Run server with docker compose up
- Run client in this PR https://github.com/PayButton/paybutton/pull/445
- Test if inputs are being returned as a param (inputAddresses) of tx object in onSuccess/OnTransaction callbacks

### DIFF
--- a/services/chronikService.ts
+++ b/services/chronikService.ts
@@ -381,10 +381,12 @@ export class ChronikBlockchainClient implements BlockchainClient {
         console.log(`${this.CHRONIK_MSG_PREFIX}: [${msg.msgType}] ${msg.txid}`)
         const transaction = await this.chronik.tx(msg.txid)
         const addressesWithTransactions = await this.getAddressesForTransaction(transaction)
+        const inputAddresses = transaction.inputs.map(inp => outputScriptToAddress(this.networkSlug, inp.outputScript))
+
         for (const addressWithTransaction of addressesWithTransactions) {
           const { created, tx } = await createTransaction(addressWithTransaction.transaction)
           if (tx !== undefined) {
-            const broadcastTxData = this.broadcastIncomingTx(addressWithTransaction.address.address, tx)
+            const broadcastTxData = this.broadcastIncomingTx(addressWithTransaction.address.address, tx, inputAddresses)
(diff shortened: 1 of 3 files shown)


Klakurka (approved):

ws-service/types.ts
```diff
@@ -17,6 +17,7 @@ export interface SimplifiedTransaction {
   timestamp: number
   address: string
   rawMessage: string
+  inputAddresses?: Array<string | undefined>
```
chedieck, line 20: I don't get this typing (here and in the other parts of this PR where this same type appears): `Array<string | undefined>` . The function you created, `getInputAddresses`, returns `string[]`, so why this undefined?

services/chronikService.ts
```diff
@@ -356,6 +356,24 @@ export class ChronikBlockchainClient implements BlockchainClient {
     }
   }
 
+  private getInputAddresses (transaction: Tx_InNode): string[] {
```
chedieck, line 359: maybe call this `getSortedInputAddresses` since is does that

services/chronikService.ts
```diff
@@ -434,10 +453,12 @@ export class ChronikBlockchainClient implements BlockchainClient {
     }
     for (const transaction of blockTxsToSync) {
       const addressesWithTransactions = await this.getAddressesForTransaction(transaction)
+      const inputAddresses = transaction.inputs.map(inp => outputScriptToAddress(this.networkSlug, inp.outputScript))
```
chedieck, line 456: why not use the function you created here? Why sorted input addresses elsewhere but not here?

Klakurka:
Tested and it works as expected (checked a few edge cases just to be sure) but I'll wait to approve until @chedieck has done so.

chedieck (dismissed):

Klakurka:
Conflicts.

lissavxo:
resolved

Klakurka (approved):

PR #922, opened 2025-01-14
Feat/inputs payment trigger option

lissavxo:
Related to #916

Depends on
---
- #917 

Description
---
Added input addresses variable to payment triggers options

Test plan
---
- Test payment triggers <inputAddresses> variable

### DIFF
--- a/services/triggerService.ts
+++ b/services/triggerService.ts
@@ -395,7 +395,7 @@ export interface PostDataParameters {
   buttonName: string
   address: string
   opReturn: OpReturnData
-  inputAddresses?: Array<string | undefined>
+  inputAddresses?: string[]
 }
 
 async function postDataForTrigger (trigger: TriggerWithPaybutton, postDataParameters: PostDataParameters): Promise<void> {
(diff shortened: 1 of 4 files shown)


utils/validators.ts
```diff
@@ -314,7 +315,8 @@ export const parsePaybuttonTriggerPOSTRequest = function (params: PaybuttonTrigg
       buttonName: '',
       address: '',
       timestamp: 0,
-      opReturn: EMPTY_OP_RETURN
+      opReturn: EMPTY_OP_RETURN,
+      inputAddresses: ['']
```
chedieck, line 319: This looks wrong. Should be `[]`, that would be an empty input addresses list. `['']` is a list with one input address, namely `''`, which is unnecessary.

services/triggerService.ts
```diff
@@ -393,6 +395,7 @@ export interface PostDataParameters {
   buttonName: string
   address: string
   opReturn: OpReturnData
+  inputAddresses?: Array<string | undefined>
```
chedieck, line 398: Why `Array<string | undefined>`? Should be `string[]`, as in the other PRs.

chedieck (changes requested):
Some issues with the code

chedieck (approved):
LGTM. Is not present on the emails, but I suppose the aim of the issue is to implement in the POST trigger only.

Issue #918, opened 2025-01-13
Chronik tests


chedieck:
Add tests for chronik functionality, including trigger execution.

PR #923, opened 2025-01-15
feat: chronik rework

chedieck:
Related to #918 

Description
---
Reworks the chronik logic:

- Removes obsolete abstractions, such as the `blockchainService.ts` file.
- Regroups initialization logic into a unique place, instead of having jobs + the chronik initialization competing against each other. The only remaining job is the one that syncs the current price every minute.
- Moves around some functions for better organization.
- Eliminates circular imports in `chronikService.ts`, allowing thus chronik to be tested with jest.
- Creates a new class `MultiBlockchainClient` to be used as the interface to access both eCash and Bitcoin Cash blockchains, inferred by the method arguments.

Test plan
---
Make sure everything chronik related run seamlessly. This PR should not alter the application behavior, except for:
- Better logs
- Probably solves some bugs, such as the `Missing prices for tx....` which I suppose was caused by the race condition between the jobs and the chronik initialization.


ScottMcDermid (approved):
Did a quick PCR, and the actual chronikService -> blockchainService swap looks good.
I skimmed the rest of the refactor, and didn't find any issues, although I'm less familiar with the codebase.

Klakurka:
We can merge once the tests are passing.

Klakurka (approved):

Issue #924, opened 2025-01-16
Account > Currency input hard to read in light mode


Klakurka:
![image](https://github.com/user-attachments/assets/2e3a3dc4-3013-44c1-bb6a-db7523374385)

PR #925, opened 2025-01-17
ui: Improve account page UI

johnkuney:
Related to #924

Description
---
I started to fix the currency select display, but the whole page needed some attention really. Here are some proposed ui edits to make the page design more cohesive 

Test plan
---
preview the site and check /account


Klakurka (changes requested):
Nice!

Just one thing - in dark mode, the time zone hover situation is still hard to read: image

johnkuney:
ah yeah, added

Klakurka (approved):

Issue #740, opened 2023-11-21
Allow exporting payments to .csv


Klakurka:
For v1, we can just allow them to export all payments from the 'Payments' page. Later we can allow them to filter it.

Klakurka:
Let's copy + paste the dropdown from 'transactions' and have it behave the same (as in filtering by currency).

Later we can add more filters.

PR #926, opened 2025-01-18
Feat/export csv payments

lissavxo:
Related to #740 

Description
---

Added export csv functionality to payments page 

Test plan
---
Run the server with `docker compose up`
Go to payments page click `Export as CSV` button it should download the payments in a csv

### DIFF
--- a/pages/api/paybutton/download/transactions/[paybuttonId].ts
+++ b/pages/api/paybutton/download/transactions/[paybuttonId].ts
@@ -1,11 +1,8 @@
 import moment from 'moment-timezone'
 import {
-  PRICE_API_DATE_FORMAT,
   RESPONSE_MESSAGES,
   DEFAULT_PAYBUTTON_CSV_FILE_DELIMITER,
   PAYBUTTON_TRANSACTIONS_FILE_HEADERS,
-  DECIMALS,
-  SUPPORTED_QUOTES,
   SupportedQuotesType,
   NetworkTickersType,
   NETWORK_TICKERS,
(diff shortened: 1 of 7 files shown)


Klakurka (changes requested):
It needs to change to the dropdown when there's more than 1 currency involved - same behaviour as on the existing transactions one.

Klakurka (approved):
The exported CSV filename should be the same format as the transaction page ones but without the button name on the front.

Klakurka (changes requested):
The exported CSV filename should be the same format as the transaction page ones but without the button name on the front.

Klakurka (dismissed):

chedieck (changes requested):
Left a minor request, but my biggest concern is:

Is it necessary to mantain both API endpoints `pages/api/payments/download/index.ts` and `pages/api/paybutton/download/transactions/[paybuttonId].ts`?

They have some differences but most code is repeated. My concern is because this separation may make future maintenance harder. In any case, I think they share some utility functions which could be imported from `utils/index.ts`, like `isNetworkValid`

A less obvious example is: `sortPaymentsByNetworkId`, which in the other view is called `sortTransactionsByNetworkId` and it does exactly the same thing, except that it takes Transactions, not Payments.

Couldn't both just take payments? In the end payments are what are used to generate the CSV anyway. 

It's an honest question, I haven't dug that deep into it, maybe there is a good reason to keep it separate like this but seeing this amount of really similar but not identical code being repeated does strike me as something odd.

chedieck (approved):

Klakurka (approved):

PR #927, opened 2025-01-21
[ui] Add loading state for button creation and export csv

johnkuney:
Related to #903

Description
---
Adding some loading state when creating a button or exporting a csv. Just went for the basic "Loading..." Could potentially do a spinner/animation if we want something fancier

Test plan
---
preview the site, and try making a button with several addresses and exporting a csv. Should see the button change to a loading state

### DIFF
--- a/components/Paybutton/LoadingSpinner.tsx
+++ b/components/Paybutton/LoadingSpinner.tsx
@@ -1,0 +1,9 @@
+import style from './paybutton.module.css'
+
+export default (): JSX.Element => {
+  return (
+    <div className={style.loading_spinner_ctn}>
+    <div className={style.loading_spinner} />
+    </div>
+  )
+}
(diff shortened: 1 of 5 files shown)


Klakurka (changes requested):
IMO may as well go for the the actual spinner for this one.

- Change the text to a spinner
- Make sure the button width doesn't change (it currently gets wider to accomodate the "Downloading..." text)

johnkuney:
Okay for sure, added. I'm realizing the button code/styles in general throughout the app is a bit unkempt...I'd like to consolidate everything and clean it up, but can do that in another PR

Klakurka (approved):

Klakurka:
Good point @chedieck - let's do that in a follow up. This is at least a bit better than what's in master.

Issue #781, opened 2024-01-12
Add copy button next to address in button detail


chedieck:
Could be here:
![image](https://github.com/PayButton/paybutton-server/assets/21281174/794a7ee4-5951-44a7-9527-892c21758412)

chedieck:
Worth mentioning that a button can have many addresses, so in this case we could have one copy button next to each address, aligned in a column (or not?). Or at the left of each address. The latter would make it automatically aligned, the former I think would be a more "traditional" UX, so not sure what is the best way to proceed here.

E.g: 

![image](https://github.com/PayButton/paybutton-server/assets/21281174/d8508ceb-4821-4187-8a42-b850d4555f5c)

PR #928, opened 2025-01-28
feat: copy address buttons

lissavxo:
Related to #781 

Description
---
Added a copy icon close to the address in the buttons page

Test plan
---
Run server with docker compose up
go to buttons page and test the copy icon close to the address


Klakurka (changes requested):
- Match the same font colours as the 'view' button: 
image
- There's no indication that the copy was successful. I'm thinking it's better if @johnkuney takes over this one and we take some time to refresh the entire div (as well as the nearly-matching one at the top of `/button/xxx`.

Klakurka:
For @johnkuney https://github.com/PayButton/paybutton-server/issues/929

Klakurka:
Superseded by #933.

Issue #931, opened 2025-01-29
Button Generator revisions for BCH


Klakurka:
- Don't include 'currency="BCH"' in the button generator's copyable HTML when a BCH address is entered as it's not necessary.
- Change 'Send XEC' hover text placeholder text to 'Send Payment'.
- Change the 'To' field placeholder text to 'Your Address (XEC or BCH)'.

PR #932, opened 2025-01-30
Feat/button generator revision

lissavxo:
Related to #931 

Description
---
Removed XEC and BCH options from currency selector
Changet 'tot and 'hover-text' placeholders

Test plan
---
Run server and check generator


Klakurka (changes requested):
We do want to keep the currency dropdown still showing XEC / BCH. It's always blank in this PR even after entering a valid address.

Klakurka (approved):

Issue #934, opened 2025-02-01
Support collapsing a day's micro payments to 1 row in CSV


Klakurka:
Say for example we have 1000 1c payments over the course of a day. 

Instead of showing 1000 rows, we'd show 1 row for $10 and pop a note somewhere in the row that it was actually several txs combined.

If done smartly, shouldn't be too much additional processing during CSV generation.

I'm thinking the threshold can be adjustable in the settings but we can do that as a v2 and just go with something like $1 USD.

Klakurka:
eCash Land addresses to test with:

- `ecash:qzgyg3llme32kpru03ff7dq7lz8087ma6glfxngvp2`
- `ecash:qr6mxvfp2hlr0qg5phhqapqz8ajv7uaxk55z9332rl`
- `ecash:qqhe60kagdq6fftrd2m2j62ygeyd7w26fvwlv0ntjk`
- `ecash:qp0g6tty4re6cqtttgus34s7ht80xvqndqtynvfhd9`

PR #942, opened 2025-02-10
Feat/collapse payments csv

lissavxo:
Related to #934 

Description
---
Change the CSV logic to collapse payments with value smaller than $1 USD in the same day. 
Added two new columns to the file **Collapsed** and **Notes**. 
If the line contains payment  **Collapsed** the column will be `true` and in **Notes** we will display `"<button-name> - <payments-collapsed-count> transactions"`

Test plan
---
- Test with one address that has a bunch of payments smaller than $1 USD se if the behavior of collapsing is working
- Also test with a list of payments smaller than $1 USD and  a larger payment between those payments, the behavior should be 3 lines 2 collapsed and  1 not collapsed following the order of the date.


Klakurka:
Tests failing.

constants/index.ts
```diff
@@ -246,8 +246,17 @@ export const PAYBUTTON_TRANSACTIONS_FILE_HEADERS = {
...
+
+export const PAYBUTTON_PAYMENT_FILE_HEADERS = {
+  ...BASE_FILE_HEADERS,
+  collapsed: 'Collapsed',
+  notes: 'Notes'
+}
+
 export const DEFAULT_PAYBUTTON_CSV_FILE_DELIMITER = ','
 export const MAX_RECORDS_PER_FILE = 10000
+export const DEFAULT_COLLAPSE_THRESHOLD_FILE_USD = 1
```
Klakurka, line 259: How about changing this from USD to `DEFAULT_CSV_COLLAPSE_THRESHOLD` and in a follow-up branch, we can make this configurable in the account settings? The threshold will use whatever currency they have set, not just USD.

Klakurka (changes requested):
- Let's remove the 'Collapsed' column from the CSV.
- Let's put all of the transaction Ids into the tx id column, comma separated (assuming that's possible).
- How about adding some tests around making sure the amounts and values add up correctly (as in the same amount and value totals as when not collapsed).

lissavxo:
> * Let's remove the 'Collapsed' column from the CSV.
> * Let's put all of the transaction Ids into the tx id column, comma separated (assuming that's possible).
> * How about adding some tests around making sure the amounts and values add up correctly (as in the same amount and value totals as when not collapsed).

The CSV uses comma to separate the columns so separating the tx Ids with commas will mess up the file, can we do semicolon (;) ?

Klakurka (changes requested):
- I know some may not want this but for us, we definitely need this to be done in both the Payments and Transactions CSVs. We'll eventually make this configurable.
- Let's do the same thingh we're doing with txids with addresses (put them all in but semicolon-separated).

Klakurka (changes requested):
- Something seems to be breaking the formatting: image
- The 'Amount' column is being rounded. We want to keep the decimal places (2 for XEC, 8 for BCH).
I'm testing w/ EL addresses.

lissavxo:
> * Something seems to be breaking the formatting: image
> * The 'Amount' column is being rounded. We want to keep the decimal places (2 for XEC, 8 for BCH).
>   I'm testing w/ EL addresses.

this is due to opening CSV in excel, changed the delimiter to semicolon (;) it  will make it readable in excel

Klakurka (changes requested):
Same issue again

image

Klakurka (changes requested):
Still an issue:

image

Looks like the problem is that there's a max length on each row (or cell).

I think I'm ok with just omitting the tx ids if there's multiple. Businesses are unlikely going to need to reference the tx id on small payments. The important ones will be the big ones. If there's more than one tx (collapsed row), just say 'Multiple'.

constants/index.ts
```diff
@@ -243,11 +243,15 @@ export const PAYBUTTON_TRANSACTIONS_FILE_HEADERS = {
   value: 'Value',
   rate: 'Rate',
   transactionId: 'Transaction Id',
-  address: 'Address'
+  address: 'Address',
+  notes: 'Notes'
 }
 
 export const DEFAULT_PAYBUTTON_CSV_FILE_DELIMITER = ','
+export const DEFAULT_MULTI_VALUES_FILE_DELIMITER = ' & '
```
Klakurka, line 251: No need for this anymore right?

Klakurka (changes requested):

constants/index.ts
lissavxo: if we wont need to concatenate addresses yes, no need. 
will we stop concatenating addresses too ?

Klakurka (changes requested):
I went through and did a more thorough check of the transactions and it appears that we're missing a bunch.

Here's some dates where we should have transactions and we somehow don't in the CSV (for the 5 `eCash Land` addresses; 6733 txs total):

- 2024-02-13
- 2024-03-06
- 2024-05-23
- 2024-05-30
- 2024-06-11
- 2024-08-23
- 2024-10-17
- 2024-10-21
- 2024-11-19
- 2025-01-09

Klakurka (changes requested):
The tx count lines up now, but the amounts and values don't.

For example, 2024-10-19 should be 2068.67 XEC, not 2069 XEC (seemingly rounded).

Klakurka (changes requested):
Exchange rates are sometimes rounding to 8 decimal places when they should be 14 (which is what the CD API returns).

Klakurka (dismissed):
Looks good - in production, I'm going to spend some extra energy making sure everything is as expected before closing the task.

utils/files.ts
```diff
@@ -98,3 +118,165 @@ export function streamToCSV (
...
+
+  return Object.values(groupedByNetworkIdPayments).reduce(
+    (acc, curr) => acc.concat(curr),
+    []
+  )
+}
+
+const getPaybuttonTransactionsFileData = (transactions: TransactionsWithPaybuttonsAndPrices[], currency: SupportedQuotesType, timezone: string): TransactionFileData[] => {
+  const paymentsFileData: TransactionFileData[] = []
+  transactions.forEach(element => {
```
chedieck, line 236: `element` is a bad name for what you know is a transaction. Any array has elements, this is the variable-naming equivalent of typing something as `any`. Just call it `tx` or `transaction`.

utils/files.ts
```diff
@@ -98,3 +118,165 @@ export function streamToCSV (
...
+    const { amount, hash, address, timestamp } = element
+    const value = getTransactionValueInCurrency(element, currency)
+    const date = moment.tz(timestamp * 1000, timezone)
+    const rate = value / amount.toNumber()
```
chedieck, line 240: I couldn't find any float precision error myself, but I don't feel confortable knowing it can happen.

We do have this price rate in the DB, no need to recalculate it from the `value / amount`, since we literally calculated the `value` from this price. 

You can use something like
```
const price = tx.prices.find(p => p.price.quoteId = QUOTE_IDS[currency.toUpperCase()])!.price.value
```

utils/files.ts
```diff
@@ -98,3 +118,165 @@ export function streamToCSV (
...
+    } as TransactionFileData)
+
+    tempGroups[groupKey] = []
+  }
+
+  const pushTx = (tx: TransactionsWithPaybuttonsAndPrices): void => {
+    const { timestamp, hash, address, amount } = tx
+    const values = getTransactionValue(tx)
+    const value = Number(values[currency])
+    const rate = value / Number(amount)
```
chedieck, line 167: Same thing I commented below about float precision errors

utils/files.ts
```diff
@@ -98,3 +118,165 @@ export function streamToCSV (
...
 }
+
+export const collapseSmallPayments = (
+  payments: TransactionsWithPaybuttonsAndPrices[],
+  currency: SupportedQuotesType,
+  timezone: string,
+  collapseThreshold: number
+): TransactionFileData[] => {
+  const treatedPayments: TransactionFileData[] = []
+  const tempGroups: Record<string, TransactionsWithPaybuttonsAndPrices[]> = {}
```
chedieck, line 129: `tempTxGroups` is more clear

utils/files.ts
```diff
@@ -98,3 +118,165 @@ export function streamToCSV (
...
+  let totalPaymentsTreated = 0
+
+  const pushTempGroup = (groupKey: string): void => {
+    const tempGroup = tempGroups[groupKey]
```
chedieck, line 133: `tempTxGroup`  here

utils/files.ts
```diff
@@ -98,3 +118,165 @@ export function streamToCSV (
...
+    if (tempGroup === undefined || tempGroup.length === 0) return
+    if (tempGroup.length === 1) {
+      pushTx(tempGroup[0])
+      tempGroups[groupKey] = []
+      return
+    }
+    const totalAmount = tempGroup.reduce((sum, p) => sum + Number(p.amount), 0)
+    const totalValue = tempGroup.reduce((sum, p) => sum + Number(getTransactionValue(p)[currency]), 0)
+    const rate = totalValue / totalAmount
```
chedieck, line 142: Same thing  I commented below about float precision error, with the difference that here we have many txs. These txs, however, all belong to the same UTC day, so they should have the same price. What you could do is

```
const uniquePrices = new Set(
  tempTxGroup
    .map(tx => tx.prices.find(p => p.price.quoteId === QUOTE_IDS[currency.toUpperCase()])!.price.value)
);
```

and then 
```
if (uniquePrices.length !== 1) {
// Should never come here
throw new Error(RESPONSE_MESSAGES.500_INVALID_PRICES_FOR_TX_ON_CSV_CREATION(uniquePrices.length))
}

const rate = uniquePrices[0]
```

utils/files.ts
```diff
@@ -40,14 +59,15 @@ export const formatPaybuttonTransactionsFileData = (data: TransactionFileData):
...
   } = data
+  const networkTicker = NETWORK_TICKERS_FROM_ID[newtworkId]
 
   return {
     ...data,
-    amount: amount.toFixed(DECIMALS[currency]),
+    amount: amount.toFixed(DECIMALS[networkTicker]),
     date: date.format(PRICE_API_DATE_FORMAT),
-    value: value.toFixed(2),
+    value: value.toFixed(DECIMALS[networkTicker]),
```
chedieck, line 70: This is wrong, will lead to USD/CAD values having 8 decimal digits if the networkId is 2, for BCH.

Should be `DECIMALS.FIAT`.

pages/api/paybutton/download/transactions/[paybuttonId].ts
```diff
@@ -112,9 +43,15 @@ export default async (req: any, res: any): Promise<void> => {
     const userReqTimezone = req.headers.timezone as string
     const userPreferredTimezone = user?.preferredTimezone
     const timezone = userPreferredTimezone !== '' ? userPreferredTimezone : userReqTimezone
-
+    let networkIdArray = Object.values(NETWORK_IDS)
+    if (networkTicker !== undefined) {
+      const slug = Object.keys(NETWORK_TICKERS).find(key => NETWORK_TICKERS[key] === networkTicker)
+      const networkId = getNetworkIdFromSlug(slug ?? NETWORK_TICKERS.ecash)
```
Comment (chedieck) [inline pages/api/paybutton/download/transactions/[paybuttonId].ts:49]: Why are you defaulting to  `ecash`? Maybe if slug is invalid, we should just throw an error, we already have `INVALID_NETWORK_SLUG_400`.

Also could have `const slug =  NETWORK_SLUGS_FROM_IDS[NETWORK_IDS[ticker]]`

Comment (chedieck) [reply pages/api/paybutton/download/transactions/[paybuttonId].ts:49]: Also, this same codeblock appears right above `downloadTxsFile`, on both the only two places this function is ever called. Why not put it inside it, then?

Comment (lissavxo) [reply pages/api/paybutton/download/transactions/[paybuttonId].ts:49]: > Also, this same codeblock appears right above `downloadTxsFile`, on both the only two places this function is ever called. Why not put it inside it, then?

Because the networkIdArray is used to fetch the txs something we are doing outside downloadTxsFile function

pages/api/paybutton/download/transactions/[paybuttonId].ts
```diff
@@ -45,10 +45,13 @@ export default async (req: any, res: any): Promise<void> => {
     const timezone = userPreferredTimezone !== '' ? userPreferredTimezone : userReqTimezone
     let networkIdArray = Object.values(NETWORK_IDS)
     if (networkTicker !== undefined) {
-      const slug = Object.keys(NETWORK_TICKERS).find(key => NETWORK_TICKERS[key] === networkTicker)
-      const networkId = getNetworkIdFromSlug(slug ?? NETWORK_TICKERS.ecash)
+      const slug =  NETWORK_SLUGS_FROM_IDS[NETWORK_IDS[networkTicker]]
```
Comment (chedieck) [inline pages/api/paybutton/download/transactions/[paybuttonId].ts:48]: Just noticed: why get the slug from the id if you're gonna get the id from the slug? if the goal is to get the id, just forget the slug, check if the ticker is valid (we have a variable with the valid network tickers) and then get the id with `NETWORK_IDS[networkTicker]`

pages/api/payments/download/index.ts
```diff
@@ -38,8 +38,11 @@ export default async (req: any, res: any): Promise<void> => {
     res.setHeader('Content-Type', 'text/csv')
     let networkIdArray = Object.values(NETWORK_IDS)
     if (networkTicker !== undefined) {
-      const slug = Object.keys(NETWORK_TICKERS).find(key => NETWORK_TICKERS[key] === networkTicker)
-      const networkId = getNetworkIdFromSlug(slug ?? NETWORK_TICKERS.ecash)
+      const slug =  NETWORK_SLUGS_FROM_IDS[NETWORK_IDS[networkTicker]]
```
chedieck, line 41: Same thing here

constants/index.ts
```diff
@@ -46,6 +46,7 @@ export const RESPONSE_MESSAGES = {
   FAILED_TO_FETCH_PRICE_FROM_API_500: (day: string, ticker: string) => { return { statusCode: 500, message: `Failed to fetch ${ticker} price for day ${day}` } },
   MISSING_WS_AUTH_KEY_400: { statusCode: 400, message: 'Missing WS_AUTH_KEY environment variable' },
   MISSING_PRICE_FOR_TRANSACTION_400: { statusCode: 400, message: 'Missing price for transaction.' },
+  INVALID_PRICES_FOR_TX_ON_CSV_CREATION_500: (pricesLenght: number) => {  return  {statusCode: 500, message: `Missing price for transaction in CSV creation. ${pricesLenght}` }},
```
chedieck, line 49: The variable name and message are not right:

To be more precise, is the number of prices that is wrong, not the prices themselves. But more than that, this happens if there is a number != 1 of prices, so there could be missing prices or there could be more prices than expected. This error says "Missing price", which could not always be the case, and just throw a random number at the end without explain what it is. A better approach here would be:
```
INVALID_PRICES_AMOUNT_FOR_TX_ON_CSV_CREATION_500: (pricesLenght: number) => { return  {
statusCode: 500, 
message: `Wrong number of prices for transactions group in CSV creation. Expected 1, got ${pricesLenght}`. 
}},
```

chedieck (approved):
Approving since this has been requested already for some time, but I think tests should improve in the future to also include the verification of the correctness of dividing same-utc-day lines into different lines across different timezones.

Klakurka (approved):
I'll approve but I agree on the tests needing to be better on follow-ups.

Issue #940, opened 2025-02-10
Transaction processed really late


Klakurka:
This transaction was sent but didn't appear when logged into paybutton for ~10 minutes: https://explorer.e.cash/tx/80ee4b0ed162c17178f8a7096f6b41720e11e0e37237cac6d58cf39ba888063d.

Additionally, it didn't sent out the payment trigger for it, at least not right away.

Expected behaviour is obviously that the transaction is sent to all listening client buttons right away and also appear in the /button and /payments pages right away as well.

Klakurka:
Seems to happen fairly frequently.

Klakurka:
![Image](https://github.com/user-attachments/assets/776e3c49-88ca-409f-abdc-d437072999c7)

Klakurka:
Txs that should be marked as confirmed also take way too long to be marked as such, so it's not just the 1 tx that's impacted each time. It looks as though it's not updating things properly via Chronik or the caching is causing it.

PR #948, opened 2025-02-12
[#940] big queries fix

chedieck:
Related to #940 

Description
---
Mainly, this PR fixes some big prisma queries that were throwing errors in production.
Additionally:

- Adds a new make command `make reset-dev-keep-db` to reset only the dev container and keep the db intact
- Adds a new make commands `make dev-from-dump` to create a dev environment but instead of seeding, using a `dump.sql` file in the repo's root directory.
- Fixes failing test on master that was failing because github updated their docker and no longer supports `docker-compose`, only `docker compose`
- Adds `logs-dev` to `make deploy`
- Freezes new messages that come during the initial sync. Then process them all `MAX_MESSAGES_TO_PROCESS_AT_A_TIME` at a time after the initial sync is done.

Test plan
---
Already running in prod.

### DIFF
--- a/.env.from-dump
+++ b/.env.from-dump
@@ -1,0 +1,1 @@
+IGNORE_SEEDING=true
--- a/.github/workflows/on-push-master.yml
+++ b/.github/workflows/on-push-master.yml
@@ -17,6 +17,6 @@ jobs:
       - name: Run unit tests
         run: make github-test-unit
       - name: Spin up required database containers
-        run: docker-compose up -d db users-service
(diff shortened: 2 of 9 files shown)


Klakurka (changes requested):
- Can we add the ne commands to the readme?
- Tests?
-

Klakurka (changes requested):
Somehow now mobile clients are not receiving payment notification messages from the server... desktop is fine.

Klakurka (approved):

PR #958, opened 2025-02-14
Fix:  blocking new incoming chronik messages

chedieck:
Description
---
Changes logic as to how new incoming chronik messages are blocked.

- Reduce blocking delay from 1s to 0.1
- Only block mempool txs coming
- Fix jobs current price syncing logs

Test plan
---
Already running on prod


constants/index.ts
```diff
@@ -262,4 +262,6 @@ export const MAX_DAILY_EMAILS = 100 // If changed, update the DB default accordi
 export const XEC_TX_EXPLORER_URL = 'https://explorer.e.cash/tx/'
 export const BCH_TX_EXPLORER_URL = 'https://blockchair.com/bitcoin-cash/transaction/'
 
-export const MAX_MESSAGES_TO_PROCESS_AT_A_TIME = 2
+export const MAX_MEMPOOL_TXS_TO_PROCESS_AT_A_TIME = 3
```
Klakurka, line 265: Why not more?

constants/index.ts
```diff
@@ -262,4 +262,6 @@ export const MAX_DAILY_EMAILS = 100 // If changed, update the DB default accordi
...
+export const CHRONIK_INITIALIZATION_DELAY = 500
+export const MEMPOOL_PROCESS_DELAY = 100
```
Klakurka, line 267: Why not 0?

chedieck: If there is more than `MAX_MEMPOOL_TXS_TO_PROCESS_AT_A_TIME`, the thread waits in a loop and checks it every `MEMPOOL_PROCESS_DELAY` miliseconds to try again. Using at 0 would add unecessary overhead, since it would check it many times without any changes. I changed the constant name to `MEMPOOL_BLOCKAGE_LOOP_DELAY` to improve the description of what it does.

constants/index.ts
chedieck: More can lead to the prisma error of  "too many db connections". This will only be used if two mempool txs arrive ALMOST at the same time, since they're processed really quickly. So this actually doesn't make much of a difference in payment processing time, only prevents the mentioned error.

Klakurka (approved):
Seems to be working well (in prod). We'll monitor.

Issue #968, opened 2025-02-26
Add paging to 'Registered Users' table in admin


Klakurka:
Same as we're doing on the other tables above it.

PR #969, opened 2025-02-28
feat: registered users refactor

lissavxo:
Related to #968 

Description
---
Changed registered users table to use TableContainer component 

Test plan
---
run server with `docker compose up` check admin page Registered Users section

### DIFF
--- a/components/Admin/RegisteredUsers.tsx
+++ b/components/Admin/RegisteredUsers.tsx
@@ -33,16 +33,16 @@ export default function RegisteredUsers({ users }: IProps): JSX.Element {
       Header: 'Admin',
       accessor: 'isAdmin',
       Cell: ({ cell }: any) => (
-        cell.value === undefined ?  'No' : <span className={style.admin}>Yes</span>
+        cell.value === true ? <span className={style.admin}>Yes</span> : 'No'
       )
     }
   ]
-
+  console.log({users})
   const data = users.map(user => ({
     id: (user.stUser?.id === undefined || user.stUser?.id === '') ? user.userProfile?.id : user.stUser?.id,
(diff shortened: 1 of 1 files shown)


components/Admin/RegisteredUsers.tsx
```diff
@@ -1,50 +1,49 @@
...
+        >
+          <span className={style.userEmail}>{cell.value}</span>
+        </a>
+      )
+    },
+    {
+      Header: 'Admin',
+      accessor: 'isAdmin',
+      Cell: ({ cell }: any) => (
+        cell.value === undefined ?  'No' : <span className={style.admin}>Yes</span>
```
chedieck, line 36: This is wrong, outputting 'Yes' always

chedieck (changes requested):
"Is admin?" is always true

components/Admin/RegisteredUsers.tsx
```diff
@@ -1,50 +1,49 @@
...
+        >
+          <span className={style.userEmail}>{cell.value}</span>
+        </a>
+      )
+    },
+    {
+      Header: 'Admin',
+      accessor: 'isAdmin',
+      Cell: ({ cell }: any) => (
+        cell.value === null ?  'No' : <span className={style.admin}>Yes</span>
```
chedieck, line 36: You should actually check id `isAdmin === true` to return "Yes".

Non admin users can have this attribute both as NULL or 0 in the DB (null or false after prisma)

chedieck (changes requested):

Klakurka (approved):

chedieck (approved):

Issue #974, opened 2025-03-01
Wrong labels on CSV notes column


Klakurka:
Example:

[PayButton-xec-transactions.csv](https://github.com/user-attachments/files/19037381/PayButton-xec-transactions.csv)

PR #978, opened 2025-03-02
fix: button name csv notes

lissavxo:
Related to #974 

Description
---
Fixed button name in csv buildind
added name of button to notes

Test plan
---
Check both payments csv and button page csv building

### DIFF
--- a/utils/files.ts
+++ b/utils/files.ts
@@ -123,7 +123,8 @@ export const collapseSmallPayments = (
   payments: TransactionsWithPaybuttonsAndPrices[],
   currency: SupportedQuotesType,
   timezone: string,
-  collapseThreshold: number
+  collapseThreshold: number,
+  userId: string
 ): TransactionFileData[] => {
   const treatedPayments: TransactionFileData[] = []
   const tempTxGroups: Record<string, TransactionsWithPaybuttonsAndPrices[]> = {}
@@ -133,7 +134,7 @@ export const collapseSmallPayments = (
     const tempTxGroup = tempTxGroups[groupKey]
     if (tempTxGroup === undefined || tempTxGroup.length === 0) return
     if (tempTxGroup.length === 1) {
-      pushTx(tempTxGroup[0])
+      pushTx(tempTxGroup[0], groupKey)
       tempTxGroups[groupKey] = []
       return
     }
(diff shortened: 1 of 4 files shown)


utils/files.ts
```diff
@@ -210,19 +212,35 @@ export const collapseSmallPayments = (
     const value = Number(values[currency])
     const dateKey = moment.tz(timestamp * 1000, timezone).format('YYYY-MM-DD')
     const dateKeyUTC = moment.utc(timestamp * 1000).format('YYYY-MM-DD')
-    const groupKey = `${dateKey}_${dateKeyUTC}`
+    const uniqueButtonName = new Set(
+      tx.address.paybuttons
+        .filter(pb => pb.paybutton.providerUserId === userId)
+        .map(pb => pb.paybutton.name)
+    )
+    if (uniqueButtonName.size > 1) {
```
chedieck, line 220: this does not look right. If a tx belongs to more than one button, which is something that we allow, we silently show an error in the logs and just pick one of the button names at random?

I think the expected behavior here is that when generating the CSV from the button page we get THAT button name here, and when generating it from the payments page we get all the names here comma separated, or something of the kind.

chedieck (changes requested):

utils/files.ts
lissavxo: fixed, separated by semicolon so it does not mess the csv that is comma separated

Klakurka (approved):

utils/files.ts
```diff
@@ -184,11 +186,12 @@ export const collapseSmallPayments = (
     tempTxGroups[groupKey] = []
   }
 
-  const pushTx = (tx: TransactionsWithPaybuttonsAndPrices): void => {
+  const pushTx = (tx: TransactionsWithPaybuttonsAndPrices, groupKey: string): void => {
     const { timestamp, hash, address, amount } = tx
     const values = getTransactionValue(tx)
     const value = Number(values[currency])
     const rate = tx.prices.find(p => p.price.quoteId === QUOTE_IDS[currency.toUpperCase()])!.price.value
+    const buttonNames = groupKey.split('_').slice(2).join(' ; ')
```
chedieck, line 194: any reason for this spaces around the `;`? It's somewhat counter  intuitive and uncommon, so I'd take it out if there is no particular reason for it.

chedieck (changes requested):
Works well, just a minor request.

Nonetheless, this `collapseSmallPayments` function has 150+ lines of code, 2 different functions are defined inside it and still some routines that could be separated into a different function (such as the paybutton name extraction). I think in its current state it is hard to read it and hard to mantain it, but refactoring could be done in a separate PR, so I'll open a separate issue for that.

Issue #542, opened 2023-06-16
Auto-focus first input when a dialog is opened


Klakurka:
For example, when creating a new button.

NorielSylvire:
@Klakurka @johnkuney do you accept help from contributors outside your organization?

PR #979, opened 2025-03-05
feat: hook autofocus first input

lissavxo:
Related to #542 

Description
---
Added a generic hook to autoFocus first input of forms in the dashboard

Test plan
---
- Go to buttons page, click in the button + to add a new button, check if the first field of the form is automatically focused  
- Also check other forms like wallets page forms to add a new wallet and to edit

### DIFF
--- a/components/Account/ChangePassword.tsx
+++ b/components/Account/ChangePassword.tsx
@@ -83,7 +83,7 @@ export default function ChangePassword ({ toggleChangePassword }: IProps): React
       {success !== '' && <div className={style.success_message}>{success}</div> }
       <form onSubmit={(e) => { void handleSubmit(onSubmit)(e) }} method='post'>
         <label htmlFor='oldPassword'>Old password</label>
-        <input {...register('oldPassword')} type='password' id='oldPassword' name='oldPassword' required />
+        <input {...register('oldPassword')} type='password' id='oldPassword' name='oldPassword' required autoFocus/>
 
         <label htmlFor='newPassword'>New password</label>
         <input {...register('newPassword')} type='password' id='newPassword' name='newPassword' required />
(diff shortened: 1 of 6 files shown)


hooks/useAutoFocusFirstInput.tsx
```diff
@@ -0,0 +1,25 @@
...
+    }
+    setTimeout(() => focusFirstInput(), 0)
+
+    const handleClick = (): void => {
+      setTimeout(() => {
+        focusFirstInput(document)
+      }, 100)
+    }
+
+    document.addEventListener('click', handleClick)
```
chedieck, line 17: Not working well, clicking in another field refocuses the first field.

hooks/useAutoFocusFirstInput.tsx
```diff
@@ -0,0 +1,25 @@
+import { useEffect } from 'react'
+
+function useAutoFocusFirstInput (): void {
+  useEffect(() => {
+    const focusFirstInput = (element = document): void => {
```
chedieck, line 5: Why have this default parameter and then call with it anyway, below?

chedieck (changes requested):

Klakurka (approved):

chedieck (changes requested):
This is not a good solution. 

The idea of using an event listener to click events to focus the first input is a bad idea, clicks can have thousands of applications, and this is evident from all the bugs this PR introduces.

Some I found:
1. Selecting a wallet in the creation button form will refocus the button name input
2. Can't select a timezone on the account view anymore
3. Misclicking a radio selection in the wallet edit will annoyingly refocus the first input, scrolling up the modal
4. Ordering a column in the txs table of the button view page will focus POST URL trigger input, also annoyingly scrolling downwards if there are enough entries/tables.


In summary, the overall experience of having the first input focused after clicking anywhere is just bad UX.

 I much rather that this is done cleanly having to edit all files that have some kind of form than this general solution that creates more problems than it solves.

chedieck (approved):
works well, clean code.

Klakurka (approved):

Issue #973, opened 2025-02-28
Have button > Edit > Submit button use the same spinner as other buttons throughout the site


Klakurka:
Right now it just shows `...` and also the width of the button changes which it shouldn't do.

PR #980, opened 2025-03-06
chore: Add button component 

johnkuney:
Related to #973

Description
---
The buttons throughout the app were pretty disjointed. They relied on a series of classnames, applied at each instance which was just pretty messy and causing some inconsistencies. 
Adding a reusable button component here to help with this.
It also handles the display of the loading state so that is more consistent, per #973

Note: I switched the order so the 'submit' buttons are on the right side of the create and edit forms. It seems more natural that way to me, but open leaving the order alone if you disagree  

Test plan
---
Preview the site and check the buttons in the app look and perform well

### DIFF
--- a/components/Auth/ForgotPassword.tsx
+++ b/components/Auth/ForgotPassword.tsx
@@ -2,6 +2,7 @@ import { useForm } from 'react-hook-form'
 import React, { ReactElement, useState } from 'react'
 import style from './auth.module.css'
 import { sendPasswordResetEmail } from 'supertokens-web-js/recipe/emailpassword'
+import Button from 'components/Button'
 
 export default function SignUp (): ReactElement {
   const { register, handleSubmit, reset } = useForm<any>()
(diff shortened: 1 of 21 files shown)


chedieck (changes requested):
1. PayButton form margins  have gone away with the changes.

Before:
image

After:
image



2. In the wallet, "Submit" is white on blue instead of the previous black on blue:

Before:
image

After:
image

3. "Create organization" button won't work when clicked, only if *Enter* is pressed.

4. Export as CSV lost white border:

Before:
image

After:
image

johnkuney:
Cool good catches. I missed a few of those places, but got them in there now. And fixed the border and text color

Klakurka (changes requested):
The Organization Create button doesn't seem to have the loading spinner when clicked.

johnkuney:
ah yes, added

chedieck (approved):

Klakurka (approved):

Issue #977, opened 2025-03-02
Add <value> to payment trigger post data


Klakurka:
Bonus points to support sending multiple fiat values (eg. USD & CAD).

Klakurka:
This is a blocker for the woocommerce feature in the WordPress plugin.

PR #984, opened 2025-03-22
Feat/value prop trigger

lissavxo:
Related to #977 

Description
---
Added "value" variable to trigger variables. 
Value will be an object with usd and cad attributes indicating the price in each fiat currency. 

Test plan
---
Create a trigger and try to usa value variable.

### DIFF
--- a/utils/validators.ts
+++ b/utils/validators.ts
@@ -273,7 +273,7 @@ export function parseTriggerPostData ({ userId, postData, postDataParameters }: 
       .replace('<opReturn>', opReturn)
       .replace('<signature>', `${JSON.stringify(signature, undefined, 2)}`)
       .replace('<inputAddresses>', `${JSON.stringify(postDataParameters.inputAddresses, undefined, 2)}`)
-      .replace('<value>', `${JSON.stringify(postDataParameters.value)}`)
+      .replace('<value>', `"${postDataParameters.value}"`)
 
     const parsedResultingData = JSON.parse(resultingData)
     return parsedResultingData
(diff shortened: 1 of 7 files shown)


Klakurka (changes requested):
We're likely going to have a lot of fiat currencies supported down the line so IMO we should just show the one they have set in their account config.

We can potentially make a task to allow them to grab a different currency if they want (or even multiple) but for now let's just use their set currency as that's what they'd be expecting.

ws-service/types.ts
```diff
@@ -18,6 +18,17 @@ export interface SimplifiedTransaction {
   address: string
   rawMessage: string
   inputAddresses: string[]
+  prices: Array<{
+    price: {
```
chedieck, line 22: No need for all that. This is supposed to be simplified, so only take what is necessary, and for what I've gathered, it would be: `value` and `quoteId`.

utils/validators.ts
```diff
@@ -273,6 +273,8 @@ export function parseTriggerPostData ({ userId, postData, postDataParameters }:
       .replace('<opReturn>', opReturn)
       .replace('<signature>', `${JSON.stringify(signature, undefined, 2)}`)
       .replace('<inputAddresses>', `${JSON.stringify(postDataParameters.inputAddresses, undefined, 2)}`)
+      .replace('<value>', `${JSON.stringify(postDataParameters.value)}`)
```
chedieck, line 276: Why are you JSON.stringifing a string? Just do as it's done with the other strings, e.g.
\`"${postDataParameters.address}"\`

ws-service/types.ts
chedieck: Then alter `getSimplifiedTransaction` to take out these fields.

chedieck (changes requested):

chedieck (approved):
I think this works well for a MVP.

Some future ideas could be including the currency name in the response & limiting the decimal cases to 2 digits.

Issue #802, opened 2024-01-28
Add Button filter on both /dashboard and /payments


Klakurka:
MVP for this would be a simple dropdown.

++ for this would be a multi-select dropdown.

Klakurka:
This would be very useful for us - mainly the one on /payments.

Issue #972, opened 2025-02-28
Add wallet & button filters on /payments


Klakurka:
Make them multi-select so more than one wallet/button can be selected at the same time. Default is to show txs from all wallets of course.

Klakurka:
Bonus points if both multi-selects can be connected so that if you're messing with the filters on one, it will only show you what makes sense in the other (to reduce the payments further).

PR #989, opened 2025-03-26
Feat/button filter payments

lissavxo:
Related to #802 

Description
---
Added a selector in payments page to filter payments by button.

Test plan
---
Test payments page button filter


Klakurka (changes requested):
I know I considered it a bonus feature at the time but we could really benefit from this as a multi-select. Able to do that in this PR?

johnkuney:
Afraid I was working on #972 as well which I has some overlap here. I was initially just working on the UI, but took a swing at making it work and got something going with multi select with some ai help, so not sure if its the cleanest solution. But can check it out on this branch https://github.com/PayButton/paybutton-server/tree/payments-filter

chedieck:
Closed for #991

PR #991, opened 2025-03-27
Feat: Payments buttons filter

johnkuney:
Related to #972

Description
---
Add a multi select button filter on the payments page that will only show payments from those buttons

Test plan
---
Run the app and check out the payments page and mess with the filters


pages/api/payments/count/index.ts
```diff
@@ -10,7 +11,40 @@ export default async (req: any, res: any): Promise<void> => {
...
+          paybuttons: {
+            some: {
+              paybutton: { id: { in: buttonIds } }
+            }
+          }
+        },
+        amount: { gt: 0 }
+      }
+
+      const totalCount = await prisma.transaction.count({
```
lissavxo, line 38: this logic should be in the service layer

services/transactionService.ts
```diff
@@ -681,14 +681,17 @@ export async function fetchAllPaymentsByUserIdWithPagination (
...
   pageSize: number,
   orderBy?: string,
-  orderDesc = true
+  orderDesc = true,
+  buttonIds?: string[]
 ): Promise<Payment[]> {
   const orderDescString: Prisma.SortOrder = orderDesc ? 'desc' : 'asc'
+
   if (orderBy === 'buttonDisplayDataList') {
     return await getPaymentsByUserIdOrderedByButtonName(
```
lissavxo, line 690: filter by buttons in getPaymentsByUserIdOrderedByButtonName is missing, this way filters will not work in the order by button

lissavxo (changes requested):

johnkuney:
Okay think I got those sorted

lissavxo (dismissed):

services/transactionService.ts
```diff
@@ -745,31 +730,40 @@ export async function fetchAllPaymentsByUserIdWithPagination (
...
-      address: {
-        userProfiles: {
-          some: {
-            userId
-          }
-        },
-        networkId: {
-          in: networkIds ?? Object.values(NETWORK_IDS)
-        }
+  const whereClause: any = {
```
chedieck, line 736: `Prisma.TransactionWhereInput`

services/transactionService.ts
```diff
@@ -713,19 +685,32 @@ export async function fetchAllPaymentsByUserIdWithPagination (
...
-        userProfiles: {
-          some: {
-            userId
-          }
-        }
-      },
-      amount: {
-        gt: 0
+  // Build the base where clause
+  const whereClause: any = {
```
chedieck, line 689: Better to use `Prisma.TransactionWhereInput` instead of `any`.

services/transactionService.ts
```diff
@@ -745,31 +730,40 @@ export async function fetchAllPaymentsByUserIdWithPagination (
...
+      networkId: {
+        in: networkIds ?? Object.values(NETWORK_IDS)
       }
     },
+    amount: {
+      gt: 0
+    }
+  }
+
+  if (Array.isArray(buttonIds) && buttonIds.length > 0) {
```
chedieck, line 750: In the function above we had similar code but with `if ((buttonIds != null)`, which was basically checking if `buttonIds !== undefined`. This `Array.isArray(buttonIds)` achieves the same result but with in a different way, which makes what it is doing a little less obvious and harder to read.

services/transactionService.ts
```diff
@@ -783,3 +777,24 @@ export async function fetchTxCountByPaybuttonId (paybuttonId: string): Promise<n
     }
   })
 }
+
+export const getFilteredTransactionCount = async (
+  userId: string,
+  buttonIds: string[]
+): Promise<number> => {
+  const whereClause: any = {
```
chedieck, line 785: No need to create this constant

chedieck (changes requested):

chedieck (changes requested):
Had some minor code complaints, but I decided to fix them myself.

I noticed that changing the filtered buttons makes the order reset to "by Date": would be nice to have it stay in place (e.g. "by Amount" ascendent, if that was selected)

johnkuney:
Cool thanks for fixing those. Sorting should be persistent now when filtering

chedieck (changes requested):
Now indeed the order is maintained after changing the paybuttons to filter with, but the page is not reset.

Meaning if I select two buttons: one with a lot of txs and one with only a few, and go to page 25 of payments, and then unselect the one with a lot of txs leaving only the one with a few, the page stays in 25 (which doesn't exist)

image


Would be nice if it would go back to the first one when some change is made to the paybuttons to filter with.

johnkuney:
ah yes good catch, should be good now

Klakurka (changes requested):
- Let's change "Filter by PayButton" to "Filter by Button" to keep the wording consistent in the UI.
- Maybe we can make a new task for improving the CSV export filename when exporting a filtered set of buttons.

Looks good to me otherwise.

johnkuney:
Okay updated that label
I did also see that error you sent in telegram about "unexpected token '<'...." Yeah, same thing if I create a new button and then try to filter by it right away. May require some investigation...

johnkuney:
Actually think I got it

Klakurka (commented):
Suggestion on what to do when we run into that error?

@chedieck

johnkuney:
@chedieck any updates here?

chedieck:
Since I can't seem to reproduce this error, I left [a commit](https://github.com/PayButton/paybutton-server/pull/991/commits/a625196bd513924b80c133d2bfe5f3f0dbbd561e) with some logs to understand better what is erroring. So far I can't tell if this would happen in prod or not since I am not being able to reproduce.

Klakurka:
I got this error when filtering right after adding a few new buttons:

image

Klakurka (approved):
Let's just go ahead with this and if there's issues in prod then we can tackle them separately.

PR #995, opened 2025-04-08
fix: client not receiving broadcast tx

chedieck:
Related to #

Description
---
Fixes issue where the websocket broadcast TX wouldn't appear for the client.

Test plan
---
Check if either a paybutton client hooked up to localhost OR localhost:3000/button/<uuid> are both  receiving living updates when a tx arrive.

### DIFF
--- a/ws-service/route.ts
+++ b/ws-service/route.ts
@@ -46,6 +46,11 @@ function handleConnection ({ namespace, socket, addressList }: HandleConnectionA
     totalConnectedInNamespace
   }
   if (addressList !== undefined) info.addressList = addressList
+  if (info.addressList?.length !== 0) {
+    info.addressList?.forEach(address => {
+      void socket.join(address)
+    })
+  }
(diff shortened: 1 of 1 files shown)


ws-service/route.ts
```diff
@@ -118,7 +123,8 @@ const altpaymentRouteConnection = async (socket: Socket): Promise<void> => {
   const headersForwardedAddresses = socket.handshake.headers['x-forwarded-for'] as string
   const userIp = headersForwardedAddresses === undefined ? '' : headersForwardedAddresses.split(',')[0]
   if (userIp === '') {
-    throw new Error('Local IP not defined.')
+    console.error('Local IP not defined.')
```
Klakurka, line 126: IMO this is a "so what" error message. What needs to happen for this to resolve itself? What would cause it in the first place?

Klakurka (approved):
One comment nit; non-blocking.

PR #996, opened 2025-04-08
fix: master workflow

chedieck:
Description
---
Fixes failing tests on master "on push" workflow.


chedieck:
This really can only be testing by merging this PR

Issue #943, opened 2025-02-10
Paybutton with multiple addresses shows one txs for one of them upon new tx arrival


chedieck:

Klakurka:
Need some clarification on what the issue is here.

Issue #800, opened 2024-01-27
Merge transaction lists for all addresses on /button/


Klakurka:
Right now we have separate lists for each address but really it doesn't make sense to do this and makes the pagination controls awkward / buggy.

We can add a new column to display the address, but only visible on large screen sizes as it's not super important info.

PR #998, opened 2025-04-14
Feat/merge addresses button detail

lissavxo:
Related to #800 #943 

Description
---
Merged transactions in a single table and added a column with the address in the button details screen.

Test plan
---
- Go to button details page and check if transactions are showing up as expected
- Test if the screen is automatically updated when a new transaction arrives


services/transactionService.ts
```diff
@@ -144,6 +144,56 @@ export async function fetchTransactionsByAddressList (
...
+  pageSize: number,
+  orderBy?: string,
+  orderDesc = true,
+  networkIdsListFilter?: number[],
+): Promise<TransactionsWithPaybuttonsAndPrices[]> {
+
+  const orderDescString: Prisma.SortOrder = orderDesc ? 'desc' : 'asc'
+
+  let orderByQuery
+  if (orderBy !== undefined && orderBy !== '') {
```
chedieck, line 159: Can you add some comments for what this code block is doing?

chedieck (changes requested):
Alignment of columns Amount, Network, Tx and Address is wrong

image

components/Transaction/AddressTransactions.tsx
chedieck, line 1: The name of this component should change now, since the behavior changed.

`PaybuttonTransactions` is more appropriate.

chedieck (changes requested):
Ordering works an it's aligned, but the little gray arrow now always shows besides the "Address" column header, despite it being ordered by another column, e.g. "Date".

chedieck (dismissed):

Issue #936, opened 2025-02-05
Have dashboard summary header show data numbers for the selected time range below it


Klakurka:
Right now, it shows lifetime revenue and payments. We'll change this to payments over the past 1W, 1M, 1Y, or All, depending on which is selected.

PR #999, opened 2025-04-19
feat: show revenue selected time

lissavxo:
Related to #936 

Description
---
Change dashboard Revenue and Payments indicators to show only the results from selected range of time.

Test plan
---
Check dashboard revenue and payments indicator


chedieck:
Works well

Shouldn't we remove the text above the plots though? Now that it would become redundant.
image

PR #1000, opened 2025-04-22
feat: new props generator

lissavxo:
Related to #

Added contributtionOffset and disableAltpayment props to paybutton generator.

Description
---

Test plan
---

### DIFF
--- a/components/ButtonGenerator/index.tsx
+++ b/components/ButtonGenerator/index.tsx
@@ -78,7 +78,12 @@ export const initialButtonState: ButtonState = {
   disablePaymentId: false,
   opReturn: '',
   contributionOffset: 0,
-  disableAltpayment: false
+  disableAltpayment: false,
+  autoClose: true,
+  onOpen: '',
+  onClose: '',
+  wsBaseURL: '',
+  apiBaseURL: '',
 }
 
 export default function ButtonGenerator (): JSX.Element {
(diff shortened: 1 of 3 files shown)


johnkuney:
Looks good! Could probably take the opportunity to also add the other missing new props?

Looks like we don't have

ws-base-url
api-base-url
on-close
on-open
auto-close 

Also unrelated, but contribution-offset doesn't look be working in the preview for me is it for you? Its meant to show in the goal progress bar right?

lissavxo:
> Looks good! Could probably take the opportunity to also add the other missing new props?
> 
> Looks like we don't have
> 
> ws-base-url api-base-url on-close on-open auto-close
> 
> Also unrelated, but contribution-offset doesn't look be working in the preview for me is it for you? Its meant to show in the goal progress bar right?

Good call, will add the missing fields. 
About contribution-offset, I am not sure but I think It isn't working because we don't have the latest master deployed.

components/ButtonGenerator/index.tsx
```diff
@@ -74,7 +76,9 @@ export const initialButtonState: ButtonState = {
   editable: false,
   widget: false,
   disablePaymentId: false,
-  opReturn: ''
+  opReturn: '',
+  contributionOffset: 0,
+  disableAltpayment: false
```
johnkuney, line 81: Make sure to include all the new props here as well with default values, otherwise they wont reset with the reset button. All the strings should just be '' and auto-close should be true

components/ButtonGenerator/index.tsx
```diff
@@ -379,6 +383,8 @@ export default function ButtonGenerator (): JSX.Element {
                   randomSatoshis={button.randomSatoshis}
                   hideToasts={button.hideToasts}
                   disableEnforceFocus={button.disableEnforceFocus}
+                  contributionOffset={button.contributionOffset}
+                  disableAltpayment={button.disableAltpayment}
```
johnkuney, line 387: same here. lets make sure we're adding all the new props here so the preview button has the intended functionality. Its unlikely anyone would actually test a different Api base url here, or the auto close, but should add it anyway

johnkuney (changes requested):
Looks good just a few comments

johnkuney (approved):
looks good

Issue #528, opened 2023-06-01
Allow viewing / editing of "paid invoices"


Klakurka:
- Display invoice link on Transaction table
- Clicking the invoice button for that payment would take them to the invoice screen for that payment
- The amount would be filled in, but they could edit things like what it's for (eg. 1 pepperoni pizza), who they (the recipient) is, their address, same info for the customer who paid, invoice # (which we can auto-fill with the format YYYY-123, auto-inc)
- Option to save as PDF
- v2 could probably allow them to "combine" multiple payments into 1 invoice for the case when the customer pays too little initially for some reason (it would be rare but could happen at scale).

Related to #136.

PR #1007, opened 2025-05-27
Feat/invoice

lissavxo:
Related to #528 

Description
---
added actions column to transactions table in button details page. 
+ add invoice, edit invoice, view invoice.

Test plan
---
Test creating a new invoice, visualize it, try to edit the invoice.


services/invoiceService.ts
```diff
@@ -0,0 +1,113 @@
...
+interface UpdateInvoiceParams {
+  description: string
+  recipientName: string
+  recipientAddress: string
+  customerName: string
+  customerAddress: string
+}
+
+export async function createInvoice (params: CreateInvoiceParams): Promise<Invoice> {
+  // const invoiceNumber = await getNewInvoiceNumber(params.userId);
```
chedieck, line 26: remove this

services/invoiceService.ts
```diff
@@ -0,0 +1,113 @@
...
+  })
+  if (invoice === null) {
+    throw new Error(RESPONSE_MESSAGES.NO_INVOICE_FOUND_404.message)
+  }
+
+  return invoice
+}
+
+export async function updateInvoice (userId: string, invoiceId: string, params: UpdateInvoiceParams): Promise<Invoice> {
+  const invoice = await getInvoiceById(invoiceId, userId)
```
chedieck, line 72: Also should check for user ownership here

services/invoiceService.ts
```diff
@@ -0,0 +1,113 @@
...
+  // const invoiceNumber = await getNewInvoiceNumber(params.userId);
+  console.log('params', params)
+  return await prisma.invoice.create({
+    data: {
+      ...params
+    }
+  })
+}
+
+export async function getInvoices (userId: string): Promise<Invoice[]> {
```
chedieck, line 35: too generic name. getUserInvoices, for example, would be better

chedieck (changes requested):
Left some comments. Besides them:

1. I think this "Action" column should be either the first or the last (probably the last). Having it in the middle is a little weird for me.
2. You didn't actually implement the edit API.

chedieck (changes requested):

services/invoiceService.ts
```diff
@@ -0,0 +1,108 @@
...
+      userId
+    },
+    orderBy: {
+      invoiceNumber: 'desc'
+    }
+  })
+  const invoicesWithOurPattern = userInvoices.filter(invoice => defaultPattern.test(invoice.invoiceNumber))
+
+  if (invoicesWithOurPattern === null || invoicesWithOurPattern.length < userInvoices.length) {
+    if (userInvoices.length > 0) {
```
chedieck, line 98: This is redundant. This is inside the condition of `userInvoices.length` being strictly greater than `invoicesWithOurPattern.length`,  which is a non-negative number, so `userInvoices.length > 0` is already given.

chedieck (changes requested):

components/Transaction/InvoiceModal.tsx
```diff
@@ -0,0 +1,239 @@
+import React, { useState, useEffect, ReactElement } from 'react'
+import style from './transaction.module.css'
+import Button from 'components/Button'
+import { CreateInvoicePOSTParameters } from 'utils/validators'
+import axios from 'axios'
+
+export interface InvoiceData {
+  id?: string
+  invoiceNumber: string
+  amount: number
```
chedieck, line 10: Not a good idea to use `number` here, can get messy with float imprecision. Best to stick to `Prisma.Decimal`

Remember to change this also in the other interfaces you're creating.

chedieck (changes requested):

chedieck (approved):

PR #1014, opened 2025-06-18
Feat/invoice pdf

lissavxo:
Related to #528 

Description
---
Added donwload pdf function in view invoice modal.

Test plan
---
Go to a transaction in button details, create an invoice, click view invoice>Download as PDF


chedieck (changes requested):
This magic number indexing of invoices from a tx (`transaction.invoices[0]`) is very unsafe. Right now if two users add the same address to one of their buttons, and one of them creates an invoice for one of the address' txs, the other will be able to see that invoice.

johnkuney:
Im getting this error if I try to view a button details page on this branch. Master seems to work fine for me...not sure what the issue is 
Screen Shot 2025-06-22 at 9 46 09 PM

lissavxo:
> This magic number indexing of invoices from a tx (`transaction.invoices[0]`) is very unsafe. Right now if two users add the same address to one of their buttons, and one of them creates an invoice for one of the address' txs, the other will be able to see that invoice.

we are fetching the transactions filtering by the userId, so that wont happen

Klakurka (changes requested):
- Being able to sort the 'Actions' column doesn't make any sense and also throws an error. 
image
- We need to figure out a more clear way for people to know what the buttons do as it is not at all obvious without hovering and the tooltip takes a second to appear. We could possibly change 'Actions' to 'Invoice' in the table header.

Klakurka (changes requested):
Minor nit but let's move the dialog buttons around a bit:

image

- Change 'Close' to 'Cancel' to match the other buttons.
- Move the 'Cancel' button to the left.
- Have the Download PDF button be on the right.

For reference: 
image
image

Klakurka (changes requested):
Forgot to mention - the submit button should use the same loading spinner as throughout the rest of the site.

Klakurka (changes requested):
- Let's jus keep it to a name for now on the initial creation 
image
- What's going on w/ this cancel button?... 
image

Klakurka (approved):

Issue #1015, opened 2025-06-19
Allow setting an org address (organization feature)


Klakurka:
When set, automatically fill in the 'Customer Address' in the `Create Invoice` dialog.

PR #1023, opened 2025-07-14
feat: org address

lissavxo:
Related to #1015 

Description
---
Added address to organization

Test plan
---
Open the dashboard, go to account, check address config

### DIFF
--- a/services/organizationService.ts
+++ b/services/organizationService.ts
@@ -13,7 +13,8 @@ export async function createOrganization ({ creatorId, name }: CreateOrganizatio
         connect: {
           id: creatorId
         }
-      }
+      },
+      address: ''
     }
   })
 }
@@ -80,12 +81,14 @@ export async function updateOrganization ({ userId, name, address }: UpdateOrgan
     throw new Error(RESPONSE_MESSAGES.USER_HAS_NO_ORGANIZATION_400.message)
   }
 
+  const updateData: Partial<Organization> = {}
+
+  if (name !== undefined && name !== '') updateData.name = name
+  if (address !== undefined && address !== '') updateData.address = address
+
   return await prisma.organization.update({
     where: { id: organization.id },
-    data: {
-      ...(name !== undefined || name === '' ? { name } : {}),
-      ...(address !== undefined || address === '' ? { address } : {})
-    }
+    data: updateData
   })
 }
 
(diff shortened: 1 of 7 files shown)


lissavxo:
> * Let's jus keep it to a name for now on the initial creation
> 
> image - What's going on w/ this cancel button?... image

Fixed

Klakurka (changes requested):
Did you forget to push something?

- Still have the address input showing on the initial creation form.
- The Cancel button is still weird. The Update and Cancel buttons should be sized and positioned like we do on the dialogs.

services/organizationService.ts
```diff
@@ -82,7 +82,10 @@ export async function updateOrganization ({ userId, name }: UpdateOrganizationIn
 
   return await prisma.organization.update({
     where: { id: organization.id },
-    data: { name }
+    data: {
+      ...(name !== undefined || name === '' ? { name } : {}),
```
chedieck, line 86: Very hard to read, and I think that's why it's wrong — should be `name !== ''`, not `name === ''`. Same thing for the address in the line below. In any case, there are more readable ways to write this logic, such as:

```
const updateData: Partial<Organization> = {}

if (name !== undefined && name !== '') updateData.name = name
if (address !== undefined && address !== '') updateData.address = address

return await prisma.organization.update({
  where: { id: organization.id },
  data: updateData,
})

```

utils/validators.ts
```diff
@@ -513,10 +515,11 @@ export const parseJoinOrganizationPOSTRequest = function (params: JoinOrganizati
 
 export const parseUpdateOrganizationPUTRequest = function (params: UpdateOrganizationPUTParameters): UpdateOrganizationInput {
   if (params.userId === '' || params.userId === undefined) throw new Error(RESPONSE_MESSAGES.USER_ID_NOT_PROVIDED_400.message)
-  if (params.name === '' || params.name === undefined) throw new Error(RESPONSE_MESSAGES.ORGANIZATION_NAME_NOT_PROVIDED_400.message)
+  if ((params.name === '' || params.name === undefined) && (params.address === '' || params.address === undefined)) throw new Error(RESPONSE_MESSAGES.ORGANIZATION_MISSING_PARAMS_400.message)
```
chedieck, line 518: Just a nitpick, but this error name "ORGANIZATION_MISSING_PARAMS_400" leads one to believe  that the organization itself is missing these, better if it was something like `MISSING_PARAMS_TO_UPDATE_ORGANIZATION_400`

chedieck (changes requested):

Klakurka (approved):

chedieck (approved):

PR #1033, opened 2025-08-07
New ChronikService tests

Klakurka:
Fixed potential Chronik service hashing bug with unexpected decodeCashAddress return type, added tests.

Related to #918.

Test plan
---
Run `yarn docker testcoverage` and make sure nothing related to addresses / hashing is broken.

### DIFF
--- a/services/chronikService.ts
+++ b/services/chronikService.ts
@@ -656,7 +656,16 @@ export function fromHash160 (networkSlug: string, type: AddressType, hash160: st
 export function toHash160 (address: string): {type: AddressType, hash160: string} {
   try {
     const { type, hash } = decodeCashAddress(address)
-    return { type, hash160: hash }
+    // decodeCashAddress can return hash as either string or Uint8Array
+    let hash160: string
+    if (typeof hash === 'string') {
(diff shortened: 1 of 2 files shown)


services/chronikService.ts
```diff
@@ -656,7 +656,16 @@ export function fromHash160 (networkSlug: string, type: AddressType, hash160: st
 export function toHash160 (address: string): {type: AddressType, hash160: string} {
   try {
     const { type, hash } = decodeCashAddress(address)
-    return { type, hash160: hash }
+    // decodeCashAddress can return hash as either string or Uint8Array
+    let hash160: string
+    if (typeof hash === 'string') {
```
chedieck, line 661: This is always the case

chedieck (changes requested):
The tests are a good addition, but I don't get this so-called 'potential' bug, since the function decodeCashAddress returns an object with a property `hash` of type `string`. The AI is saying that it can be a `Uint8Array`, but that is not what the library says:
image
image

Klakurka (commented):
Looks like unnecessary / dubious precaution.

Issue #1021, opened 2025-07-10
Support binding arbitrary data to a future payment


Klakurka:
Let's do this our way first, then eventually support BIP 70 when we're able to.

Client task: https://github.com/PayButton/paybutton/issues/66

PR #1034, opened 2025-08-08
feat: add client payment table

lissavxo:
Related to https://github.com/PayButton/paybutton-server/issues/1021

Description
---
Added generate paymentId endpoint 

Test plan
---
Test api/payments/paymentId/ endpoint

## Summary by CodeRabbit

* **New Features**
  * New POST endpoint to create payment IDs for a recipient address (optional amount). Returns a paymentId for tracking.
  * Automatic payment status updates (Pending → Added to mempool → Confirmed) based on on-chain activity.
  * Improved request validation for address and amount inputs.

* **Chores**
  * Added backend support and data structures to track client payments and their statuses.


pages/api/payments/paymentId/index.ts
```diff
@@ -0,0 +1,12 @@
+import { generatePaymentId } from 'services/transactionService'
+
+export default async (req: any, res: any): Promise<void> => {
+  if (req.method === 'GET') {
+    const address = req.query.address as string
```
chedieck, line 5: Use parseAddress before commiting it to the DB.

chedieck (changes requested):

redis/types.ts
```diff
@@ -71,3 +71,5 @@ export interface ButtonData {
 export interface PaymentDataByButton {
   [id: string]: ButtonData
 }
+
+export type ClientPaymentStatus = 'PENDING' | 'COMFIRMED' | 'ADDED_TO_MEMPOOL'
```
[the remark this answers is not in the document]
chedieck: Also, after you address the `schema.prisma` change, this won't even have to be defined — just imported from prisma.

prisma-local/schema.prisma
```diff
@@ -267,4 +267,13 @@ model Invoice {
 
   @@unique([invoiceNumber, userId], map: "Invoice_invoiceNumber_userId_unique_constraint")
   @@unique([transactionId, userId], map: "Invoice_transactionId_userId_unique_constraint")
+}
+
+model ClientPayment {
+  id                   String                 @id @default(dbgenerated("(uuid())"))
+  paymentId            String                 @unique
+  status               String               
+  address              String
```
chedieck, line 276: Maybe is a good call to also add a foreign relation here:
call this one `addressString` and below add

```
  address      Address   @relation(fields: [addressString], references: [address])
```

...this way we can access easier the relation using prisma ORM in the TS code.

(will also have to add
```
clientPaymentIds                     ClientPaymentId[]
```
in the address table)

chedieck (changes requested):

chedieck (changes requested):
Actually something I missed and found when reviewing the other PR:
Problem here is that when we ask for a paymentId for an address that is NOT in the db yet, although it creates it, it does not sync the txs and, worse, it does not subscribe it to chronik. So we will never get the notification that the TX was made.


Another problem: this table should also hold the amount. Otherwise any TX with any amount would set the ClientPaymentId status to "Confirmed", even if it was way less than the intended.

pages/api/payments/paymentId/index.ts
```diff
@@ -0,0 +1,28 @@
+import { Decimal } from '@prisma/client/runtime/library'
+import { generatePaymentId } from 'services/transactionService'
+import { parseAddress } from 'utils/validators'
+import { RESPONSE_MESSAGES } from 'constants/index'
+
+export default async (req: any, res: any): Promise<void> => {
+  if (req.method === 'GET') {
```
chedieck, line 7: This should work the same way as it does now, but should actually be a POST request. It will create an entry in the DB (more than one, if the address doesn't exist), it's not idempotent and it changes state.

chedieck (changes requested):
Just rewrite it to be a POST request and it looks good to go.

utils/validators.ts
```diff
@@ -563,3 +563,24 @@ export interface CreateInvoicePOSTParameters {
...
+  amount?: string
+}
+export interface CreatePaymentIdInput {
+  address: string
+  amount?: string
+}
+
+export const parseCreatePaymentIdPOSTRequest = function (params: CreatePaymentIdPOSTParameters): CreatePaymentIdInput {
+  if (
+    params.address === undefined
```
chedieck, line 577: Check for `''` too

utils/validators.ts
```diff
@@ -563,3 +563,24 @@ export interface CreateInvoicePOSTParameters {
...
+  ) {
+    throw new Error(RESPONSE_MESSAGES.ADDRESS_NOT_PROVIDED_400.message)
+  }
+
+  return {
+    address: params.address,
+    amount: params.amount
```
chedieck, line 584: Also probably make so that `amount` is returned as `undefined` if it arrives as `''`.

prisma-local/schema.prisma
```diff
@@ -267,4 +268,21 @@ model Invoice {
...
+}
+
+enum ClientPaymentStatus {
+ PENDING
+ ADDED_TO_MEMPOOL
+ CONFIRMED
+}
+
+model ClientPayment {
+  id                   String                 @id @default(dbgenerated("(uuid())"))
```
chedieck, line 280: We don't really need this field, do we? `paymentId` will serve as a PK AFAIK.

chedieck (changes requested):

prisma-local/schema.prisma
```diff
@@ -277,7 +277,6 @@ enum ClientPaymentStatus {
 }
 
 model ClientPayment {
-  id                   String                 @id @default(dbgenerated("(uuid())"))
   paymentId            String                 @unique
```
chedieck, line 280: This does not work. You have to specify one @id, so you can change this @unique here in paymentId to @id.

Then don't edit the migration manually, delete it and run prisma migrate dev to generate it. Right now this scheme is out of sync with the migration file so it does not build.

utils/validators.ts
```diff
@@ -574,13 +574,14 @@ export interface CreatePaymentIdInput {
...
+    params.address === undefined ||
+    params.address === ''
   ) {
     throw new Error(RESPONSE_MESSAGES.ADDRESS_NOT_PROVIDED_400.message)
   }
 
   return {
     address: params.address,
-    amount: params.amount
+    amount: params.amount === undefined || params.amount === '' ? undefined : params.amount
```
chedieck, line 585: Unnecessarily complex.
Same as:

``` 
amount: params.amount === '' ? undefined : params.amount
```

chedieck (changes requested):

Issue #1037, opened 2025-08-10
Have new tx processing (eg. TX_ADDED_TO_MEMPOOL) not be blocked by unrelated syncing


Klakurka:
Transactions always need to be processed immediately.

PR #1038, opened 2025-08-11
[#1037] feat: dont block all addresses during initial syncing 

chedieck:
Related to #1037 

Description
---
Optimize syncing by avoiding unnecessary address refetching, preventing duplicate price connections, refining logs, and having the initial sync block addresses message processing one-by-one, not all at the same time.

Also have Supertokens core fixed in version 10.0.0.

Test plan
---
Run in this branch  `make dev-from-dump` using prod's DB as `dump.sql`  in the repo root folder.

Check the logs. Make a TX to a address and see it appear and say it is waiting for syncing to unblock it. See when that address is synced, and the new TX message being processed.

Maybe run something like `select count(*) from Address where syncing = 0;` in the DB, while the sync is happening, to see how many addresses have been cleared so far (and will now process new txs normally)

### DIFF
--- a/services/addressService.ts
+++ b/services/addressService.ts
@@ -299,7 +299,7 @@ export async function fetchAddressById (addressId: string, includePaybuttons = f
 }
 
 export async function setSyncingBatch (addressStringArray: string[], syncing: boolean): Promise<void> {
-  const result = await prisma.address.updateMany({
+  await prisma.address.updateMany({
     where: {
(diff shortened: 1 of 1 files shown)


Klakurka (dismissed):
Works great! It's really fast locally so I want to see how it is in prod.

Klakurka (approved):

PR #1046, opened 2025-08-18
Added <outputAddresses> post data variable option

Klakurka:
Description
---
With this, it's possible to get the output addresses for a transaction. We currently only list the addresses themselves and not the amount sent to each address.

Test plan
---

1. Create a payment trigger with the new <outputAddresses>
2. Use [httpbin.org/post](https://httpbin.org/post) as the URL.
3. Send a payment that would execute the payment trigger and check the TriggerLogs table for the output.

## Summary by CodeRabbit

- New Features
  - Trigger post data now supports outputAddresses with per-address amounts and preserves primary address ordering.
  - Transaction broadcasts and realtime payloads include input/output address amounts for richer context.
  - Signature payload generation skips empty values for cleaner signing.

- Documentation
  - Clarified integration test guidelines.
  - Updated Paybutton help text to list the new outputAddresses variable.

- Tests
  - Added comprehensive unit tests for trigger parsing, JSON validation, and transaction simplification.

- Chores
  - Switched pre-commit to run lint-staged via Yarn; refined lint-staged config and script.


services/chronikService.ts
```diff
@@ -387,22 +387,58 @@ export class ChronikBlockchainClient {
...
         const currentValue = addressSatsMap.get(address) ?? 0n
         addressSatsMap.set(address, currentValue + inp.sats)
       }
     })
-
     const sortedInputAddresses = Array.from(addressSatsMap.entries())
       .sort(([, valueA], [, valueB]) => Number(valueB - valueA))
-      .map(([address]) => address)
+    return sortedInputAddresses.map(([address, sats]) => {
+      const format = xecaddr.detectAddressFormat(address)
```
chedieck, line 402: This is unnecessary. We already know the network, accessible from `this.networkSlug`. 
Best to define something like `decimalCases` outside the `.map` to be either `1e2` or `1e8` based on that, and then here, inside the `.map`, only apply it, to avoid this redundant checks that could slow the code overall if they are ran many times in parallel (like when a block arrives)

services/chronikService.ts
```diff
@@ -387,22 +387,58 @@ export class ChronikBlockchainClient {
...
+      const address = outputScriptToAddress(this.networkSlug, out.outputScript)
+      if (address !== undefined && address !== '') {
+        const currentValue = addressSatsMap.get(address) ?? 0n
+        addressSatsMap.set(address, currentValue + out.sats)
+      }
+    })
+    const sortedOutputAddresses = Array.from(addressSatsMap.entries())
+      .sort(([, valueA], [, valueB]) => Number(valueB - valueA))
+      .map(([address, sats]) => {
+        const format = xecaddr.detectAddressFormat(address)
```
chedieck, line 429: Same thing here. No need to detect the network per tx address, it will always be the same one (`this.networkSlug`)

chedieck (changes requested):

PR #1050, opened 2025-08-25
[#1049] feat: async and improved initialization

chedieck:
Related to #1049 

Description
---
Parallelizes initial sync of transactions, making it way faster. This includes the parallelization of price connections, trigger execution, chronik txs fetch per address and almost everything else in the pipeline.

Furthermore:
- Separates `emailCredits` from `postCredits`
- Fixes general error on Tables when data is `undefined`
- Creates actual relation between `PayButton` and `UserProfile` (yes, this didn't exist until now)
-  limits for standard and Pro users, adds tracking of post credits, improves trigger execution and logging, and refines batching/concurrency for sync and pricing connections.
- uses 'https://httpbin.org/post' for local `dev` user seeded trigger
- Adds "es2019" to tsconfig in order to use `Array.flatMap`

Test plan
---
- Start app, either with `make dev` or with `make dev-from-dump` using a production DB to test how it performs with a big DB.
- Check the logs for, e.g.
```
2|next  | 2025-08-25T19:35:08: [CHRONIK — ecash] syncAddresses: 1:27.566 (m:ss.mmm)
```
... to see how long it took to sync all `ecash` / `bitcoincash` addresses (it took 1m27s in the example above)

- Check `TriggerLog` table in the db to see that logs are behaving as expected. We don't fire triggers when using prod DB locally, so this can be tested when running the app locally with the seeded DB.
- Confirm out-of-credits logs are generated.

## Summary by CodeRabbit

- **New Features**
  - Display daily POST limits in Pro account settings.
  - New bulk trigger execution API for batched posts/emails.

- **Improvements**
  - Parallel, batched transaction syncing for faster, more reliable syncs.
  - Batched price-linking for faster transaction-price associations.
  - Table loader now defaults to empty arrays when no results returned.

- **Behavior Changes**
  - New user defaults: email credits and post credits set to 5.

- **Documentation**
  - README and example config updated with daily POST limit settings.


lissavxo:
Shoud we add the post credits in the UI also like we do with email ?
Screenshot 2025-08-29 at 18 04 00

Klakurka (commented):
Looks like a few minor things copilot found.

Klakurka (changes requested):
I'm good merging after going through the above ^.

Klakurka (commented):
When deployed in prod, it does seem to block the UI while doing [PARALLEL FETCHING].

Klakurka:
image

Issue #1098, opened 2025-12-28
Dashboard sometimes just doesn't load without a refresh


Klakurka:
We just get the spinner with no page data.

Klakurka:
<img width="2062" height="323" alt="Image" src="https://github.com/user-attachments/assets/dcc88e85-ce2f-4912-a085-5705a30b5161" />

PR #1109, opened 2026-02-20
fix: Attempted fix for dashboard not loading sometimes

Klakurka:
Related to #1098

Description
---
Sometimes (especially after a re-deploy) the dashboard doesn't load without doing an extra refresh.

Test plan
---
I'm not 100% how to reproduce it without trying this for real in prod but perhaps there is a good way? We can also give it a quick deploy in prod together.

## Summary by CodeRabbit

* **Bug Fixes**
  * Concurrent requests now return the existing in-flight result instead of raising an error, improving reliability for identical simultaneous requests.
  * Per-user request tracking and cleanup corrected to avoid stale state after calls complete.

* **Refactor**
  * Internal request-tracking updated to store and manage in-flight results per method for more predictable behavior.

### DIFF
--- a/redis/index.ts
+++ b/redis/index.ts
@@ -76,8 +76,10 @@ export const CacheSet = {
 
 type MethodName = 'dashboardData' | 'paymentList' | 'addressBalance' | 'paymentsCount'
 
+type InFlightCalls = Partial<Record<MethodName, Promise<unknown>>>
+
 interface PendingCalls {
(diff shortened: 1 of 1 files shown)


Fabcien:
Not sure about the fix. The root cause seems that either the dashboard cache request is emitted concurrently, which is dubious, or that a call is missing an await.

PR #1114, opened 2026-02-25
fix:  memory, cpu usage & prisma connection limit

chedieck:
WIP
---

changes removed and marked as draft due to errors in prod

Related to #

Description
---
We were getting some errors in production related to prisma having too many connections open, this PR puts a limit (20) on them so that further connections are queued.

Also, in my local dev initJobs was using a lot of CPU and failing due to memory overload, so I decreased the size of constants related to the initial sync parallel batch sizes in order to reduce the overload. 

Finally, some refactor into the function that fetches the non-synced txs to avoid copying/storing data unnecessarily.

## Summary by CodeRabbit

* **Chores**
  * Optimized batch processing and concurrency parameters for improved system efficiency
  * Added new configuration for payment expiration timeouts (7 days), price connection handling, and trigger logging


Fabcien:
This is impossible to review as-is. There are 3 issues:
 - A prisma error: can you actually paste the error message ?
 - A memory error: what amount of memory is used? Is it only allocated during sync or is it high all the time ?
 - A high cpu usage during sync: without a flame graph you can't really do any analysis on the root cause

PR #1129, opened 2026-05-22
Improve memory consumption during syncing

Fabcien:
This implements a few strategies but is mostly a 2 part fix:
 - Avoid a producer/consumer race that can be lost. It happens that fetching from Chronik for eCash is faster than committing to DB, causing to queue to grow unbounded. This diff makes sure we stop adding txs to the queue and drain it fully each time before resuming.
 - Simplify the sync process by only filling up the DB info, deferring price update and cache rebuilds.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved blockchain sync to invalidate stale caches after sync so users see up-to-date balances and payments.
  * More reliable transaction draining/coordination to reduce missed or duplicated events.

* **Refactor**
  * Faster, safer bulk transaction persistence with smarter dedupe and targeted updates.
  * Optimized cache key discovery and deletion to avoid large Redis key scans.

* **Tests**
  * Expanded mocks to cover new sync helpers and ensure stable test runs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/PayButton/paybutton-server/pull/1129?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/constants/index.ts
+++ b/constants/index.ts
@@ -288,7 +288,7 @@ export const TRIGGER_EMAIL_CONCURRENCY = 100
 export const TRIGGER_LOG_BATCH_SIZE = 200
 
 export const PRICES_CONNECTION_BATCH_SIZE = 1_000
-// interactive $transaction timeout in ms (for the single delete + several createMany of prices)
+// interactive $transaction timeout in ms (for batched delete + createMany of prices)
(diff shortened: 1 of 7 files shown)


Fabcien:
I removed my fix commit for the dump I used in tests, it's irrelevant for prod code

chedieck (approved):

PR #1131, opened 2026-06-02
chore: dev from dump improvement

chedieck:
Description
---
- Fix `make dev-from-dump` workflow to spin up a local dev env with a prod database dump
- Named Docker volume for DB so the long import only happens once
- `pv` progress bar during dump import
- Skips prisma migrations (uses `migrate deploy` since dump already has schema + migration history)
- Skips cache rebuild in dev-from-dump (prevents MariaDB I/O saturation)
- Pages load in seconds instead of hanging for minutes with prod-sized data

### New make targets
- `make stop-dev-from-dump` — stop without losing DB
- `make reset-dev-from-dump` — full restart
- `make reset-dev-from-dump-keep-db` — restart dev container only, DB stays
- `make nuke-dev-from-dump` — destroy everything including DB volume

### Remarks
- Requires a `dump.sql` in the repo root (from `mariadb-dump` on prod)
- First run takes 15-30 min (importing 2.6GB dump). Subsequent runs are instant

Test Plan
---
1. Place a prod dump.sql in root
2. `make dev-from-dump` — watch db logs for pv progress
3. Once done, navigate to localhost:3000 — dashboard, payments, buttons, wallets should all load within seconds
4. `make stop-dev-from-dump && make dev-from-dump` — should skip import (instant start)
5. `make nuke-dev-from-dump && make dev-from-dump` — should reimport from scratch

## Summary by CodeRabbit

* **Chores**
  * Added development commands and Docker Compose support for dump-based workflows; new env flags to enable dump mode and skip cache rebuild.
  * Persisted dump-based database data to a named volume.
  * Updated ignore rules to exclude a runtime file.

* **Performance**
  * Optionally skip cache rebuild to speed startup when using database dumps.

* **Usability**
  * Improved database import with progress reporting and more robust startup behavior for migrations/seeding.


scripts/db/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM mariadb:latest
 RUN mkhomedir_helper mysql
 
 RUN apt-get update || echo && \
```
Fabcien: Not sure I understand the use of || echo here, if the goal is to avoid failures then || true seems more appropriated

scripts/db-entrypoint.sh
```diff
@@ -11,14 +11,24 @@ cd /home/mysql/raw_entrypoint || exit
...
         sed_vars "$file"
     fi
 done
 
 cd ../entrypoint/ || exit
 
 for file in *.sql; do
-    mariadb -u root -p"$MYSQL_ROOT_PASSWORD" < "$file"
+    if [[ "$file" == *dump* ]]; then
+        filesize=$(stat -c%s "$file" 2>/dev/null || stat -f%z "$file" 2>/dev/null)
```
Fabcien: I don't get the second part of the subcommand `stat -f%z`, what does this do ?

scripts/paybutton-server-start.sh
```diff
@@ -17,12 +17,18 @@ logtime=$(date +%Y-%m-%d@%H:%M)
...
+elif [ "$ENVIRONMENT" = "production" ]; then
     yarn prisma migrate deploy || exit 1
     yarn prisma generate || exit 1
     pm2 start yarn --time --interpreter ash --name jobs --output logs/jobs.log --error logs/jobs.log -- initJobs || exit 1
     pm2 start yarn --time --interpreter ash --name WSServer --output logs/ws-server.log --error logs/ws-server.log -- initWSServer || exit 1
-    pm2 start yarn --time --interpreter ash --name next --output logs/next.log --error logs/next.log -- prod || exit 1
+    if [ "$ENVIRONMENT" = "production" ]; then
+        pm2 start yarn --time --interpreter ash --name next --output logs/next.log --error logs/next.log -- prod || exit 1
+    else
+        pm2 start yarn --time --interpreter ash --name next --output logs/next.log --error logs/next.log -- dev || exit 1
```
Fabcien: This is unreachable, you're already in a if [ "$ENVIRONMENT" = "production" ]; branch
Edit: I see this is fixed in the follow up commit. Better squash in this case to make the review easier

.gitignore
```diff
@@ -39,3 +39,4 @@ prisma-local/seeds/productionTxs.csv
 paybutton-config.json
 
 dump.sql*
+.forever
```
Fabcien, line 42: out of curiosity, what is this?

scripts/db-entrypoint.sh
chedieck: Same thing, but for BSD/MacOS; overall unecessary bc this will run inside the container anyways, but was just something the AI did by default

.gitignore
chedieck: I took this pattern out of another blockchainventures project, it's just a hidden directory to put repo-related stuff that should not be committed, temp stuff, without adding untracked files and "dirtying" the worktree

scripts/db-entrypoint.sh
```diff
@@ -10,11 +10,25 @@ sed_vars () {
...
+    if [[ "$file" == *dump* ]]; then
+        filesize=$(stat -Lc%s "$file" 2>/dev/null || stat -Lf%z "$file" 2>/dev/null)
+        echo "Importing $file ($(numfmt --to=iec $filesize)) ..."
+        pv "$file" | mariadb -u root -p"$MYSQL_ROOT_PASSWORD"
+    else
+        echo "Importing $file ..."
+        mariadb -u root -p"$MYSQL_ROOT_PASSWORD" < "$file"
+    fi
+    echo "Done importing $file"
 done
```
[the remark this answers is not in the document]
Fabcien: `set -euo pipefail` at the top of the script is a good idea

Fabcien (approved):

PR #1132, opened 2026-06-02
fix: tx prices linking, dashboard load, initial sync errors

chedieck:
Depends On
---
- [ ] #1131

Description
---
- Transactions with missing prices no longer crash the entire sync batch or cache rebuild — they're skipped with a warning, and should be fixed by the jobs thread.
- Chronik WebSocket disconnections no longer kill the initial sync — reconnects with exponential backoff
- TX_CONFIRMED handler throttled to max 6 concurrent — prevents DB connection exhaustion on blocks with many tracked txs
- Cache rebuild no longer blocks page loads — uses cursor pagination, excludes unused relations, and doesn't starve the connection pool
- Payments/wallets endpoints use direct DB queries instead of slow subqueries

### Performance
- `/api/payments` uses `addressId IN (...)` instead of EXISTS subquery (120s → 4s for large users)
- `/api/payments/count` uses direct DB count instead of dashboard cache recomputation
- Wallet balance uses SQL aggregate instead of loading all transactions
- Cache rebuild uses cursor-based pagination and skips `inputs` relation

Test Plan
---
This is running on prod rn.
To test locally, preferrably use prod's DB locally according to the workflow detailed in #1131
1. Spin up the local mirror of prod
2. Verify addresses with missing prices log warnings but don't halt the batch
3. Check that, while the jobs are syncing the address and the dashboard is being created, we still don't get endless loading pages and get the partial data served
4. Wait for a block with multiple tracked txs — confirm no connection exhaustion errors
5. Navigate to payments page as a large user — should load in <5s

## Summary by CodeRabbit

* **New Features**
  * Added a “Data still being calculated” banner on the dashboard, with automatic polling to refresh results while rebuilds are in progress.
  * Improved dashboard and payment updates by moving long-running cache rebuilds to background processing.

* **Bug Fixes**
  * Reduced duplicate cache work under concurrent requests.
  * Updated transaction/payment counting and address/payment calculations to align with current filters.
  * Improved Chronik connection resiliency and capped confirmed-transaction processing to prevent overload.

* **Tests**
  * Updated dashboard and service tests to reflect the new rebuild status behavior.


constants/index.ts
```diff
@@ -277,10 +277,10 @@ export const CONFIRMED_TX_PROCESS_DELAY = 100
 export const CHRONIK_WS_MAX_RETRIES = 10
 export const CHRONIK_WS_BASE_DELAY_MS = 5000
 
-// Number of tries before failing a chronik call (min 1)
-export const CHRONIK_TRIES = 3
+// Number of tries before failing a chronik HTTP call (min 1)
+export const CHRONIK_HTTP_MAX_RETRIES = 3
```
Fabcien: This is not the number of retries, but the number of tries. It's now off by one

services/transactionService.ts
```diff
@@ -544,19 +545,32 @@ export async function connectTransactionsListToPrices (
...
+      continue
     }
     rows.push(...buildPriceTxConnectionInput(t, allPrices))
   }
+
+  if (skippedTxIds.length > 0) {
+    console.warn(`[PRICES] Skipped ${skippedTxIds.length} txs due to missing price data.`)
+  }
+
+  const txIdsToConnect = txList.filter(t => !skippedTxIds.includes(t.id)).map(t => t.id)
```
Fabcien: This is very inefficient, you can build txIdsToConnect on the fly (instead of the skipped ones) and check if txList.length > txIdsToConnect.length to determine if there is any skipped tx. This avoids the O(n²) search and map building + saves some memory from the temporary filter vector.

constants/index.ts
```diff
@@ -277,7 +277,7 @@ export const MEMPOOL_PROCESS_DELAY = 100
 // On chronik, the max allowed is 200
 export const CHRONIK_FETCH_N_TXS_PER_PAGE = 200
 
-export const INITIAL_ADDRESS_SYNC_FETCH_CONCURRENTLY = 16
+export const INITIAL_ADDRESS_SYNC_FETCH_CONCURRENTLY = 64
```
Fabcien: Not sure about that... this might not improve the bandwidth usage but can increase the memory consumption significantly

services/chronikService.ts
```diff
@@ -181,6 +182,42 @@ export class ChronikBlockchainClient {
     }
   }
 
+  private async connectWsWithRetry (maxRetries = 10, baseDelay = 5000): Promise<void> {
+    for (let attempt = 1; attempt <= maxRetries; attempt++) {
```
Fabcien: This is off by one

services/chronikService.ts
```diff
@@ -664,10 +701,16 @@ export class ChronikBlockchainClient {
...
+      onReconnect: (_: ws.Event) => {
+        console.log(`${this.CHRONIK_MSG_PREFIX}: Chronik webSocket unexpectedly closed. Attempting reconnection...`)
+        void this.reconnectWs()
+      },
       onConnect: (_: ws.Event) => { console.log(`${this.CHRONIK_MSG_PREFIX}: Chronik webSocket connection (re)established.`) },
-      onEnd: (e: ws.Event) => { console.log(`${this.CHRONIK_MSG_PREFIX}: Chronik WebSocket ended, type: ${e.type}.`) },
-      autoReconnect: true
+      onEnd: (e: ws.Event) => {
+        console.log(`${this.CHRONIK_MSG_PREFIX}: Chronik WebSocket ended, type: ${e.type}. Attempting reconnection...`)
+        void this.reconnectWs()
```
Fabcien: I'm not sure that's what you want, this could prevent you from stopping the app as it will keep trying to reconnect ?

constants/index.ts
Fabcien: NVM just saw the follow-up. You'd better squash this one

constants/index.ts
```diff
@@ -269,8 +269,10 @@ export const XEC_TX_EXPLORER_URL = 'https://explorer.e.cash/tx/'
 export const BCH_TX_EXPLORER_URL = 'https://blockchair.com/bitcoin-cash/transaction/'
 
 export const MAX_MEMPOOL_TXS_TO_PROCESS_AT_A_TIME = 2
+export const MAX_CONFIRMED_TXS_TO_PROCESS_AT_A_TIME = 6
```
Fabcien: OTOH this default seems very low to me. A block full will take ~10 minutes of total delay, not accounting for the tx processing itself

services/chronikService.ts
```diff
@@ -831,6 +832,12 @@ export class ChronikBlockchainClient {
         }
       } else if (msg.msgType === 'TX_CONFIRMED') {
         if (this.isAlreadyBeingProcessed(msg.txid, true)) return
+
+        while (this.confirmedTxsBeingProcessed >= MAX_CONFIRMED_TXS_TO_PROCESS_AT_A_TIME) {
```
Fabcien, line 868: This is very hacky. What about a proper message queue ?

redis/paymentCache.ts
```diff
@@ -319,14 +332,40 @@ export const initPaymentCache = async (address: Address): Promise<boolean> => {
...
+    const keys = await getCachedWeekKeysForAddress(address.address)
+    if (keys.length === 0) {
+      uncachedAddresses.push(address)
     }
   }
+
+  if (uncachedAddresses.length > 0 && process.env.SKIP_CACHE_REBUILD === undefined) {
+    console.log(`[CACHE] ${uncachedAddresses.length}/${allAddresses.length} uncached addresses for user ${userId}, starting background rebuild`)
+    void cacheAddressesInBackground(uncachedAddresses, userId)
+  }
```
Fabcien: This is duplicated from getPaymentList, you should call that instead

prisma-local/clientInstance.ts
```diff
@@ -1,22 +1,39 @@
...
 
 interface CustomNodeJsGlobal extends NodeJS.Global {
   prisma: PrismaClient
+  backgroundPrisma: PrismaClient
 }
 declare const global: CustomNodeJsGlobal
 
+const backgroundDbUrl = (process.env.DATABASE_URL ?? '').replace(
+  /connection_limit=\d+/,
+  'connection_limit=3'
```
Fabcien: Maybe make it a constant?

redis/paymentCache.ts
```diff
@@ -177,6 +177,7 @@ export const generateAndCacheGroupedPaymentsAndInfoForAddress = async (address:
         paymentCount++
       }
     }
+    await new Promise(resolve => setTimeout(resolve, 200))
```
Fabcien, line 181: You should add a comment on this one. I'm not sure it's really helping anything but at least if somebody else looks at the code there should be an explanation about why there is a seemingly random delay in the processing workflow

services/addressService.ts
```diff
@@ -228,17 +227,20 @@ export interface AddressPaymentInfo {
...
-  const balance = transactionsAmounts.reduce((a, b) => {
-    return a.plus(b)
-  }, new Prisma.Decimal(0))
-  const zero = new Prisma.Decimal(0)
-  const paymentCount = transactionsAmounts.filter(t => t > zero).length
-  const info = {
-    balance,
+  const address = await fetchAddressBySubstring(addressString)
+  const [balanceResult, paymentCount] = await Promise.all([
+    prisma.transaction.aggregate({
```
Fabcien, line 232: nice

services/transactionService.ts
```diff
@@ -1217,10 +1217,14 @@ export async function fetchAllPaymentsByUserIdWithPagination (
     }
   }
 
+  const userAddresses = await prisma.address.findMany({
+    where: { userProfiles: { some: { userId } } },
+    select: { id: true }
+  })
+  const userAddressIds = userAddresses.map(a => a.id)
```
Fabcien, line 1223: too bad prisma can't flatten the result directly! It's rebuilding the object from the SQL, unflattening the result with that single attribute so you can flatten it again, what a lost

Fabcien (changes requested):

docker-compose-from-dump.yml
```diff
@@ -70,6 +69,3 @@ services:
       - ./redis:/data/redis:z
       - ./scripts:/data/scripts
     command: sh -c "sed -i 's/\\r//g' ./scripts/redis-start.sh && sh ./scripts/redis-start.sh"
-
-volumes:
-  paybutton-dump-db-data:
```
Fabcien, line 75: Can you explain why this commit exists ? These changes where beneficial I don't see the point reverting them

redis/dashboardCache.ts
```diff
@@ -400,14 +399,6 @@ export const cacheDashboardData = async (userId: string, dashboardData: Dashboar
   await redis.set(key, JSON.stringify(cachable))
 }
 
-const cacheDashboardDataWithTTL = async (userId: string, dashboardData: DashboardData, ttlSeconds: number): Promise<void> => {
```
Fabcien: Same why the revert ? What's the problem ? These reverts make the PR impossible to review, you really need to squash the fixes so the actual changes can be reviewed

services/transactionService.ts
chedieck: True, fixed.

services/chronikService.ts
chedieck: I renamed `maxRetries` to `maxAttempts` and `CHRONIK_WS_MAX_TRIES` to `CHRONIK_WS_MAX_CONNECTION_ATTEMPTS`.  Now it should be good

services/chronikService.ts
chedieck: Indeed, I think the reconnection logic only at `onReconnect` should fix the issue.

constants/index.ts
chedieck: I bumped it to 20, but in the end we have to try: Since we have to connect new txs to address & prices, this would sometimes try to do too much at once and end up timing out the DB connection

redis/paymentCache.ts
chedieck: Refactored both to use same auxiliary function.

prisma-local/clientInstance.ts
chedieck: I took it out, we can change that in .env already

redis/paymentCache.ts
chedieck: Fixed

services/transactionService.ts
chedieck: Indeed.

docker-compose-from-dump.yml
chedieck: I don't understand what commit in particular you are referring to, nor how you got that diff. I am guessing is because this branch was behind master after we merged #1131 (where these lines were added to the file) to it, but I am not sure. I haven't changed this remote branch since this comment, but it says Outdated already

image

redis/dashboardCache.ts
chedieck: I removed the changes because I thought maybe not ideal to be serving out-of-date data for 30s, and thought we don't gain much with the partial caches.

About the reverts, usually I review the latest version. If I would squash this revert to the commit I added the changes, that would've been an empty commit, no? I don't get why it would make it better, but sure I can improve my workflow in order to make it easier for reviewers once I understand it.

